PR #60, opened 2026-02-19
adds ai2 deployed models

warmbowski:
## Description

* adds ai2 models that are deployed on the Ai2 Model Hub application (LiteLLM Gateway) to list of model presets
* adds vendor prefixes to litellm models to make sure litellm uses the correct api key
* revamps litellm config to pass api_base

## Type of Change

- [ ] Documentation update
- [x] Refactoring (no functional changes)

## Testing

- [ ] Unit tests pass locally (`pytest tests/ --ignore=tests/integration/`)

## Checklist

- [ ] My code follows the project's style guidelines
- [ ] I have performed a self-review of my code
- [ ] I have added/updated documentation as needed
- [ ] My changes generate no new warnings

### DIFF
--- a/README.md
+++ b/README.md
@@ -703,7 +703,8 @@ cluster: h100
 | `--name` | `-n` | required | Experiment name |
 | `--model` | `-m` | required | Model name or HuggingFace path (can specify multiple) |
 | `--task` | `-t` | required | Task name with optional `@priority` suffix (can specify multiple) |
-| `--override` | `-o` | none | Override for preceding `-m` or `-t` (can specify multiple) |
+| `--harness` | `-H` | `default` | Harness preset name |
+| `--override` | `-o` | none | Override for preceding `-m` or `-t` or `-H` (can specify multiple) |
 | `--cluster` | `-c` | required | Cluster alias (`h100`, `a100`, `aus`) or full name |
 | `--gpus` | `-G` | `1` | Number of GPUs per model instance |
 | `--parallelism` | `-P` | `1` | Number of model instances to run in parallel |
--- a/src/olmo_eval/common/constants/models.py
+++ b/src/olmo_eval/common/constants/models.py
@@ -71,32 +71,32 @@ def get_model_presets() -> dict[str, ProviderConfig]:
         ),
         "gpt-4o": ProviderConfig(
             kind=ProviderKind.LITELLM,
-            model="openai/gpt-4o",
-            base_url="https://api.openai.com/v1",
+            model="gpt-4o",
+            api_base="https://api.openai.com/v1",
             required_secrets=("OPENAI_API_KEY",),
         ),
         "gpt-4o-mini": ProviderConfig(
             kind=ProviderKind.LITELLM,
             model="openai/gpt-4o-mini",
-            base_url="https://api.openai.com/v1",
+            api_base="https://api.openai.com/v1",
             required_secrets=("OPENAI_API_KEY",),
         ),
         "gpt-4-turbo": ProviderConfig(
             kind=ProviderKind.LITELLM,
             model="openai/gpt-4-turbo",
-            base_url="https://api.openai.com/v1",
+            api_base="https://api.openai.com/v1",
             required_secrets=("OPENAI_API_KEY",),
         ),
         "claude-3-opus": ProviderConfig(
             kind=ProviderKind.LITELLM,
             model="anthropic/claude-3-opus-20240229",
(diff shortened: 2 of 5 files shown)


src/olmo_eval/inference/providers/litellm.py
```diff
@@ -49,6 +51,7 @@ def __init__(
         Args:
             model_name: Model identifier (e.g., "gpt-4", "claude-3-opus").
             base_url: Optional base URL for OpenAI-compatible endpoints.
+            api_base: Optional API base for litellm.completion (if different from base_url).
```
warmbowski: I can fix this description, that parenthetical is not needed.

src/olmo_eval/inference/providers/litellm.py
warmbowski, line 91: Is base_url and get_openai_client needed here? I see calls in the code to functions of the same name, but I think they are not this function. (unsure).

undfined: This was a generic method so all Provider class instances can get an openai client the same way. If you see an opportunity to simplify please propose :)

src/olmo_eval/harness/presets.py
```diff
@@ -75,6 +75,21 @@ def default(name: str) -> HarnessConfig:
...
+
+        return HarnessConfig(
+            name=name,
+            provider=ProviderConfig(kind=ProviderKind.VLLM_SERVER),
+            metrics=MetricsConfig(
+                enabled=True,
+                collect_gpu=False,
+            ),
+            batching=BatchConfig.batched(),
+        )
```
undfined, line 91: So two things here:

1) You actually have provider type set to vllm_server which I don't think is what you want
2) Default harness with command line overrides should be sufficient instead of a specific harness config here since you only need to set metrics config items.

undfined (changes requested):
Let's drop the harness preset in favor of cli overrides, but otherwise looks good.

src/olmo_eval/harness/presets.py
warmbowski: Oh I can get rid of this. I didn't understand that the order of args mattered for the -o flag, and that it applies to the previous arg passed. I just figured it out that I need to call out `-H default` then `-o metrics.enabled=true` to override the harness setting. 

I created this because of my misunderstanding.

README.md
```diff
@@ -703,7 +703,8 @@ cluster: h100
 | `--name` | `-n` | required | Experiment name |
 | `--model` | `-m` | required | Model name or HuggingFace path (can specify multiple) |
 | `--task` | `-t` | required | Task name with optional `@priority` suffix (can specify multiple) |
-| `--override` | `-o` | none | Override for preceding `-m` or `-t` (can specify multiple) |
+| `--harness` | `-H` | `default` | Harness preset name |
+| `--override` | `-o` | none | Override for preceding `-m` or `-t` or `-H` (can specify multiple) |
```
warmbowski, line 707: added harness to the list of options, as well as calling it out for overrides

src/olmo_eval/common/constants/models.py
```diff
@@ -71,36 +71,49 @@ def get_model_presets() -> dict[str, ProviderConfig]:
...
+            model="litellm_proxy/openai/Olmo-3-7B-Instruct",
+            api_base="https://ai2-model-hub.allen.ai",
+            required_secrets=("LITELLM_PROXY_API_KEY",),
+        ),
+        "modal-olmo-3-7b-instruct": ProviderConfig(
+            kind=ProviderKind.LITELLM,
+            model="litellm_proxy/openai/ai2-release-partners/Olmo-3-7B-Instruct",
+            api_base="https://ai2-model-hub.allen.ai",
+            required_secrets=("LITELLM_PROXY_API_KEY",),
+        ),
```
undfined, line 118: Nit: I might put the hosting service as a suffix so prefix filtering for models is more predictable.

undfined (approved):
One nit, otherwise lgtm!

src/olmo_eval/common/constants/models.py
warmbowski: The provider prefix is there for litellm.completions to use the right config and env var for the provider. It doesn't look like there is any other way to steer litellm to the proper provider, other than with this prefix. I can only think of splitting that prefix out into a field like `litellm_provider` on the LITELLM provider kind, and assembling the prefixed path at completions time, if that makes sense.

PR #95, opened 2026-03-13
tau2 verified

pdasigi:
## Description

Switch from the original Tau2 Bench to [Tau2 Bench Verified](https://github.com/amazon-agi/tau2-bench-verified). The verified version contains manual fixes of some instances with incorrect ground-truth values.

One related issue to note: Qwen often runs out of context window while evaluating on Tau2 Bench. In the original version, we would pass a truncation argument to the agent like `"truncate": "left"` to truncate the input to fit the context window. However, the verified version uses a newer version of litellm which does not support this argument. We need to call [trim_messages](https://docs.litellm.ai/docs/completion/message_trimming) on the input messages, which needs to happen in the benchmark code. But that is independent of this PR.

## Type of Change

- [ ] Documentation update

## Testing

## Checklist


undfined (approved):
Nice, easy change!

PR #122, opened 2026-04-09
Fix: ignore vllm-specific kwargs for HF provider

AkshitaB:
## Summary
- Only inject vllm-specific kwargs (`tensor_parallel_size`, `load_format`, `model_loader_extra_config`, `enable_auto_tool_choice`, `log_dir`) when the provider is `vllm` or `vllm_server`
- Fixes errors when using the HF provider, which doesn't accept these kwargs

## Test plan
- [x] Run eval with HF provider and verify no unexpected kwargs are passed
- [x] Run eval with vllm provider and verify vllm-specific kwargs still work


davidheineman:
Seems reasonable to me.

davidheineman (approved):

undfined (approved):

PR #153, opened 2026-04-23
Add SciCode

finbarrtimbers:
Adds [SciCode](https://scicode-bench.github.io/) as an `ExternalEval`.

Verification:
1. Reproduces Artificial Analysis's 41% target on `google/gemma-4-31b-it` with `enable_thinking=true`: `sub_step_accuracy` = 42.01% (121/288).
2. `microsoft/Phi-4-mini-instruct` gets 9.375% vs AA's 10.8% target.
3. Post-review re-run on `microsoft/Phi-4-mini-instruct` (commit 28cda45, 2026-04-29): `sub_step_accuracy` = 10.07% (29/288), matching AA's 10.8% target. [Beaker](https://beaker.org/ex/01KQCTH7HZ1ZTKF8EANKQ6K4D6)

## `h5py_host_path` is required

The eval expects the SciCode reference numerics file (`test_data.h5`) to be available on the host and bind-mounted into the sandbox. There is no default — pass `-A h5py_host_path=...` at launch. Running without it raises `ValueError`. Ai2 internal users (detected by `BEAKER_TOKEN` in env) get a hint in the error message pointing at `/weka/oe-adapt-default/finbarrt/scicode/test_data.h5`.

## Sandbox image changes

This PR also reworks `src/olmo_eval/harness/sandbox/image.py` to support SciCode's verifier:

- SciCode's verifier runs `uv sync --active --frozen --no-dev --project ...` inside the sandbox to install per-problem dependencies. `uv sync --active` requires a real venv pointed to by `VIRTUAL_ENV`, which the previous `/root/python` raw-interpreter layout did not provide. The Dockerfile now bakes a venv at `/root/venv` and exports `VIRTUAL_ENV`, so SciCode's runtime `uv sync` lands deps in the right place.
- Switches from a pinned `python-build-standalone` tarball (x86_64-only, ~2-year-old build) to `COPY --from=ghcr.io/astral-sh/uv:0.11.7 /uv /uvx ...` as a build stage. This is multi-arch, maintained, and lets `uv` pick the Python interpreter (`uv venv --python 3.12`).
- Drops the redundant second `pip install uv` step now that `uv` is the bootstrap.
- Bumps `SWEREX_IMAGE_VERSION` so the cache key reflects the new on-disk layout (`/root/python` → `/root/venv`).

## JSON round-trip for `-a` / `-K` args

The Beaker launcher passes parsed eval args to the inner `run-external` CLI as repeated `-a key=value` strings, and the inner CLI re-parses them via `_coerce_value`. Before, the launcher used Python's default `str()` formatting, so:

- `["13.1", "13.2"]` became `"['13.1', '13.2']"` — Python repr with single quotes is not valid JSON, so `_coerce_value` returned a literal string instead of a list.
- `True` / `None` similarly became `"True"` / `"None"` rather than JSON's `true` / `null`.

This bit the SciCode smoke test concretely: `--problem-ids` is list-typed, and the GPU-side eval received a string. The fix is symmetric: `json.dumps` on the way out (`job_assembler.py`), `json.loads` on the way in (`cli/utils.py`). `tests/cli/test_arg_round_trip.py` pins the round-trip for str / bool / int / float / list / dict / None.

## Reproducing

Phi-4 Mini (non-reasoning, AA-compliant, ~14m on 1×H100):
```
uv run olmo-eval beaker launch -m microsoft/Phi-4-mini-instruct -E scicode \
  -A with_background=true -A max_tokens=16384 -A temperature=0 \
  -A h5py_host_path=/weka/oe-adapt-default/finbarrt/scicode/test_data.h5 \
  -K max_model_len=131072 \
  -p urgent -w ai2/open-instruct-dev -B ai2/oe-adapt -c h100
```

Gemma-4-31b-it (reasoning, ~1h 48m on 2×H100):
```
uv run olmo-eval beaker launch -E scicode \
  -I finbarrt/olmo-eval-cu1281-trc290-amd64-sandbox-vllm \
  -K 'chat_template_kwargs={"enable_thinking":true}' \
  -A with_background=true -A max_tokens=16384 -A temperature=0.6 \
  -A h5py_host_path=/weka/oe-adapt-default/finbarrt/scicode/test_data.h5 \
  -m google/gemma-4-31b-it \
  -p urgent -w ai2/open-instruct-dev -c h100 -B ai2/oe-adapt -G 2
```

## Stacked PRs

The bake-once sandbox image / launch-host pre-bake mechanism has been split into a follow-up PR that depends on this one.


src/olmo_eval/inference/dispatch.py
```diff
@@ -112,6 +115,7 @@ async def _safe_process(self, item: T) -> tuple[R | None, Exception | None]:
             result = await self.process_fn(item)
             return (result, None)
         except Exception as e:
+            logger.warning("dispatch: process_fn raised %s: %s", type(e).__name__, e)
```
finbarrtimbers, line 118: Otherwise, this swallows the error.

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
```diff
@@ -0,0 +1,315 @@
...
+        provider: InferenceProvider,
+        args: dict[str, Any],
+        output_dir: str | None = None,
+        container_runtime: str = "podman",
+    ) -> ExternalEvalResult:
+        start_time = time.time()
+        sc_args = SciCodeConfig(**args)
+
+        if not isinstance(provider, VLLMServerProvider):
+            raise TypeError(f"SciCode requires VLLMServerProvider, got {type(provider).__name__}.")
```
undfined: I think this will raise because it's an `InstrumentedProvider`. 

Maybe we need something like:
```
class InstrumentedProvider:
    ...

    @property
    def base_provider(self) -> InferenceProvider:
        """Return the underlying non-instrumented provider.

        Unwraps nested InstrumentedProvider layers if present.
        """
        provider: InferenceProvider = self._provider
        while isinstance(provider, InstrumentedProvider):
            provider = provider.base_provider
        return provider
```

and then:

```
raw_provider = (
    provider.base_provider if isinstance(provider, InstrumentedProvider) else provider
)

if not isinstance(raw_provider, VLLMServerProvider):
    raise TypeError(
        f"SciCode requires VLLMServerProvider, got {type(raw_provider).__name__}."
    )
```

src/olmo_eval/harness/sandbox/image.py
```diff
@@ -194,12 +191,12 @@ def _resolve_swerex_image(
 RUN apt-get update && \\
     apt-get install -y --no-install-recommends curl git ca-certificates && \\
     rm -rf /var/lib/apt/lists/*
-ADD {PYTHON_STANDALONE_URL} /tmp/python.tar.gz
-RUN tar xzf /tmp/python.tar.gz -C /root && rm /tmp/python.tar.gz && \\
-    ln -sf /root/python/bin/python3 /root/python/bin/python && \\
-    /root/python/bin/pip install --no-cache-dir swe-rex uv
-ENV PATH="/root/python/bin:$PATH"
+COPY --from={UV_IMAGE} /uv /uvx /usr/local/bin/
+RUN uv venv /root/venv --python 3.11 && \\
```
undfined: Will this work with `requires-python = "==3.12.*"` ?

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
```diff
@@ -0,0 +1,315 @@
...
+from olmo_eval.inference.providers.vllm_server import VLLMServerProvider
+
+from . import loader as scicode_loader
+from . import prompts as scicode_prompts
+from . import verifier as scicode_verifier
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_H5PY_HOST_PATH = "/weka/oe-adapt-default/finbarrt/scicode/test_data.h5"
```
undfined: Can we make the default something more generic and have Ai2 users override in the launch command so we don't couple runner code with beaker knowledge?

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
```diff
@@ -0,0 +1,315 @@
...
+        container_runtime: str,
+    ) -> list[bool]:
+        if not scorable_indices:
+            return []
+
+        sandbox_config = SandboxConfig(
+            image=sc_args.sandbox_image,
+            mode=SandboxMode.DOCKER,
+            container_runtime=cast(ContainerRuntime, container_runtime),
+            capabilities=Capability.PYTHON,
```
undfined: This is poor interface design on my part so we should fix it, but I think the `execute_code()` calls in this  `_verify()` method will fail to route without either specifying `Capability.PYTHON` in the `execute_code()` call or by advertising `BASH` as a capability in the sandbox like below. Having language and capability arguments in the execute_* method can possibly be collapsed to simplify this. 

```capabilities=(Capability.PYTHON, Capability.BASH)```

undfined (changes requested):
I noticed a few cases that will likely raise exceptions that we should fix and just one hygiene request to avoid adding Ai2 paths in source when possible.

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
finbarrtimbers: Good point, done.

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
finbarrtimbers: Ah! Ok, ty. Done.

src/olmo_eval/harness/sandbox/image.py
finbarrtimbers: yes, it works fine.

src/olmo_eval/evals/external/benchmarks/scicode/eval.py
finbarrtimbers: Done, thank you!

undfined (approved):
Noice.

PR #154, opened 2026-04-24
Apply chat template in in-process vllm provider

finbarrtimbers:
## Summary
- The in-process `VLLMProvider.generate` was passing `req.prompt` directly to vLLM, which is empty for `RequestType.CHAT` requests (where only `req.messages` is populated). All chat-style tasks (e.g. `aime_2025:pass_at_32`) failed with `ValueError: The decoder prompt cannot be empty`.
- Fix: when the request is CHAT with messages, apply the tokenizer's chat template (`add_generation_prompt=True`) to build the prompt string before handing it to vLLM.
- Also bump pinned `vllm`/`transformers` versions so the provider can load models saved with `transformers>=5.4.0` (new `rope_parameters`/`layer_types` config shape).

## Test plan
- [x] `olmo-eval run -m /model --harness default -o provider.kind=vllm -t aime_2025:pass_at_32` on a Qwen3 4B base model. Previously failed with empty-prompt errors for all 30 instances; now completes with `pass_at_1:minerva_math_flex=0.2156`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,10 +41,11 @@ Issues = "https://github.com/allenai/olmo-eval/issues"
 
 [project.optional-dependencies]
 hf = [
-    "transformers~=4.57.3",
+    "transformers>=5.4.0",
 ]
 vllm = [
-    "vllm[runai]==0.13.0",  # runai extras enable S3 model loading
+    "vllm[runai]>=0.19.0",  # runai extras enable S3 model loading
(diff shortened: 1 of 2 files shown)


finbarrtimbers:
Closing — we're switching our caller to use provider.kind=vllm_server (the supported path for chat-format tasks), so this patch isn't needed for our use case. Happy to reopen if the team wants the in-process vllm provider to handle CHAT requests.

PR #217, opened 2026-06-16
Science-literature evals: citation scorer, ExpertQA, LitSearch

donovanr:
## Description

Adds a reusable citation-grounding scorer and two scientific-literature evals built on it: ExpertQA (attributed long-form QA) and LitSearch (literature retrieval, in two modes). Both benchmarks use a the scorer from `astabench_sqa` which has been refactored out.

## Citation-grounding scorer

`common/scorers/citation.py` (entry point `score_citations_for_sections`), extracted out of `astabench_sqa`, which now delegates to it. Produces claim-level citation precision/recall via an LLM judge (`gpt-5.5:medium` by default; override with `OLMO_EVAL_JUDGE`).

## ExpertQA

2,177 expert-written questions across 32 fields (`cmalaviya/expertqa`, config `main`). The model writes a cited long-form answer; we grade attribution and on-topic precision (the dataset's human labels grade its own pre-generated answers, so we score what the shared judges support without a reference):

- `citation_recall` (primary), `citation_precision` — claim-level attribution
- `answer_precision` — fraction of paragraphs the judge finds on-topic

```bash
uv run olmo-eval beaker launch \
  -n "roryd-olmo3-expertqa" \
  -m allenai/Olmo-3-7B-Instruct \
  -H default \
  -o provider.kind=vllm_server
  -o provider.max_model_len=16384 \
  -t expertqa@urgent \
  -w "ai2/olmo-eval-debug"
  -B "ai2/oe-other"
  --cluster ai2/ceres \
  --group roryd-sci-lit
```

This eval lives in the `science:research` and `science:judge` suites.

## LitSearch (two modes)

597 literature-search queries with gold Semantic Scholar corpus IDs (Ajith et
al., EMNLP 2024). There are two complementary tasks over this data:

### `litsearch_rerank` — fixed-corpus reranking (reproducible, judge-free, tool-free)

Each query carries a frozen pool of BM25-retrieved candidates; the model selects/ranks them and we score Recall@k over its own selection. Pools are built offline by `scripts/internal/build_litsearch_pools.py` (BM25 over LitSearch `corpus_clean`, 64,183 papers; 50 candidates/query, text baked in) and hosted on the Hub at `allenai/litsearch-rerank-pools`, so the eval needs no corpus download at run time. Metrics: recall@5 (primary), recall@20. In `science:research`, `science:nojudge`, `science:all`.

```bash
uv run olmo-eval beaker launch \
  -n "roryd-olmo3-litsearch-rerank" \
[...]

PR #238, opened 2026-07-02
Add SAGE scientific paper-retrieval benchmark (short_form + open_ended)

yilunzhao:
## Description

Adds **SAGE** ([arXiv:2602.05975](https://arxiv.org/abs/2602.05975), *SAGE: Benchmarking and Improving Retrieval for Deep Research Agents*), an agentic scientific paper-retrieval benchmark. Given a reasoning-intensive, **descriptive** query (venue, year, methods, what it cites…), an agent uses a paper-search tool to identify the target paper(s); we score the agent's **final answer** (title matching), not its trajectory. Both tasks live in the `science:research` suite.

## Dataset

`allenai/sage-retrieval` on the Hub (public; configs `short_form` (599) + `open_ended` (600), Semantic Scholar-enriched).

## `sage_short_form` — single-paper identification

One gold paper per query; the agent searches, then gives its single best answer.

- `exact_match` (primary) — normalized gold title present in the final answer.

## `sage_open_ended` — graded-set retrieval

A research question with a graded gold set (`most_relevant`, weight 2; `relevant`, weight 1); the agent lists the most relevant papers it found.

- `weighted_recall` (primary) — Σ(matched relevance) / Σ(total relevance).

```bash
olmo-eval run -m Qwen/Qwen3-8B \
  -H paper_search_agent -o provider.kind=vllm_server \
  -t sage_short_form -o limit=50 --num-gpus 1 --save-predictions
```

Needs an agentic paper-search harness (`paper_search_agent`, from #217, now on main).

## Scoring / matcher

Scoring uses a deterministic `NormalizedStringMatcher` (normalized gold-title substring), selected over LLM-judge alternatives on a human-labeled validation set:

- `sage_match_validation.py` ships 111 labeled cases plus an adversarial layer, with a CI regression test. NormalizedString scores **F1 0.948** — judge-free, reproducible, and adds no dependencies.
- A per-gold LLM judge scored marginally higher (0.974) but is costly and non-deterministic; its cheaper batched form (~0.88) underperforms the free matcher.

## Contents

- `evals/tasks/sage.py` — both tasks + inline scoring
- `evals/tasks/sage_match_validation.py` (+ `sage_matcher_validation.jsonl`) — matcher validation harness
- `scripts/internal/build_sage_dataset.py` — dataset build/enrichment
- `evals/suites/science.py` — registration under `science:research`
- `tests/evals/tasks/test_sage*.py` — fixture-based, no network


src/olmo_eval/evals/tasks/sage.py
```diff
@@ -0,0 +1,384 @@
...
+        )
+
+    async def score_responses(
+        self,
+        responses: Sequence[Response],
+        context: Any = None,
+    ) -> Sequence[Response]:
+        """Score each response by matching the final output against the gold paper."""
+        for response in responses:
+            output = response.outputs[0] if response.outputs else None
```
donovanr: Does this only look at outputs[0]? If num_samples > 1 the rest of the samples may get dropped / not recorded in metadata, and the response score ignores them. The framework normally aggregates across outputs (_apply_scorers → _aggregate_output_scores, base.py:585/666). LitSearch gets away with reading outputs[0] because it scores one trajectory per response; here I think if you're scoring each output's text the aggregation may matter.

scripts/internal/build_sage_dataset.py
```diff
@@ -0,0 +1,430 @@
+"""Offline builder for the SAGE retrieval dataset mirror.
```
donovanr, line 1: Do we want to call out the harness requirement up top, the way LitSearch does? Since SAGE grades the final answer instead of the trajectory, running it without `paper_search_agent` won't score zero. The model just answers from memory and you get numbers that look real but aren't. LitSearch at least logs a warning when the trajectory is missing; SAGE has no equivalent signal, so a clear note in the docstring may help with misleading results?

donovanr:
@yilunzhao I left a couple of comments. Also SAGE_REPO points at yilunzhao/sage-retrieval, should we change it to a allenai/ mirror in this PR rather than after, so the eval is reproducible the get-go?.

src/olmo_eval/evals/tasks/sage.py
yilunzhao: Good catch, fixed in dff29ea. Both tasks now score every output, write per-output `score:exact_match` / `score:weighted_recall` metadata (the keys pass@k expansion reads), and aggregate through `self._aggregate_output_scores` so `output_score_aggregation` is honored. Added multi-output tests covering both MAX and FIRST, plus an empty-outputs edge test.

scripts/internal/build_sage_dataset.py
yilunzhao: Agreed, and this failure mode is real: we ran into it ourselves during early baseline runs. Fixed in dff29ea: the sage.py module docstring now calls out the harness requirement the way LitSearch does, and both tasks log a warning when responses come back with no trajectory, pointing at `paper_search_agent`.

yilunzhao:
@donovanr Thanks for the review! All three points addressed in dff29ea: every output is now scored and aggregated through the framework path, the harness requirement is called out in the docstring plus a runtime warning, and the dataset is mirrored to `allenai/sage-retrieval` (public) with `SAGE_REPO` pointing at it. PR description updated accordingly.

src/olmo_eval/evals/tasks/sage.py
donovanr: Thanks, the multi-output scoring, tool-harness warning, and AllenAI dataset repo all look addressed.

One remaining loose end from the earlier review: `normalize_title` still intentionally skips article stripping and the docstring says SAGE’s original EM normalization is pending confirmation. Could you either confirm/update that normalization, or explicitly note that we’re deferring OT-2 and matching the task packet for now?

yilunzhao: Confirmed the normalization directly with the SAGE first author: `NormalizedStringMatcher` uses their reference EM implementation. The reference release (github.com/HughieHu/Sage) ships data + README only (no scoring code), and the paper (arXiv:2602.05975) defines short-form EM as whether the gold paper is "included in the output text or citations", which we implement as a normalized substring-inclusion check. Updated the docstring in a60bc00 to record this instead of deferring to an unconfirmed original. Resolves OT-2.

donovanr:
@yilunzhao ready to merge?

scripts/internal/build_sage_dataset.py
```diff
@@ -0,0 +1,430 @@
...
+from pathlib import Path
+from typing import Any
+from urllib.error import HTTPError, URLError
+from urllib.parse import quote
+from urllib.request import Request, urlopen
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+logger = logging.getLogger("build_sage_dataset")
+
+POOLS_REPO = "yilunzhao/sage-retrieval"  # HF repo set (OT-1)
```
undfined: I think Rory mentioned this as well but you want this to be your allenai/ dataset correct? Also I think POOLS_REPO is unused currently yes?

src/olmo_eval/evals/tasks/sage.py
```diff
@@ -0,0 +1,430 @@
...
+                "case_id": (
+                    doc.get("case_id")
+                    or doc.get("paper_id")
+                    or ground_truth.get("paperId")
+                    or f"sage_short_form_{index}"
+                ),
+                "domain": doc.get("domain", ""),
+                "index": index,
+            },
+        )
```
undfined, line 306: Claude scanned the short form dataset and indicated that there are multiple instances with the same paper_id (which seems fine) but the fallback to paper_id here may result in duplicates for case_id I believe. Does it make sense to include query_id as a first alternate to paper_id?

undfined (approved):
look great! 2 things to check before merging please.

yilunzhao:
Thanks @undfined @donovanr — addressed the last two items in 83f1b76.

**`case_id` → `query_id`** (undfined's `sage.py` comment): confirmed against the real dataset. `case_id` isn't present in the SAGE data at all (`0/599` short_form, `0/600` open_ended), so the fallback always fires, and `paper_id` has 15 collisions in short_form (`584` unique / `599`) — exactly the duplicate `case_id` you flagged. `query_id` is present and unique across both splits (`599/599`, `600/600`), so both tasks now prefer `query_id` ahead of `paper_id`. This also matches the existing convention in `litsearch_rerank` (`doc.get("query_id", index)`). Added two tests that lock in the priority.

**`POOLS_REPO` / allenai mirror** (undfined's `build_sage_dataset.py` comment): `POOLS_REPO` was unused anywhere in the repo — removed it. The builder's `SAGE_REPO` default now points at `allenai/sage-retrieval`, matching the eval's `data_source`, which already loads from there.

@donovanr ready to merge from my side — all review items resolved, gates green (ruff + ty clean, 19 SAGE tests pass).

PR #239, opened 2026-07-02
Harness robustness for agentic evals (tool-error handling + parser detection)

yilunzhao:
## Description

Two independent harness-robustness fixes surfaced while running agentic paper-retrieval evals ([SAGE](https://arxiv.org/abs/2602.05975), #238). Both benefit any agentic task (LitSearch, Dr. Tulu, SAGE), so they are split out from the benchmark PR.

## Agentic tool-error handling

`harness/scaffolds/openai_agents.py` re-raised `ModelBehaviorError` (e.g. a hallucinated tool name), which dropped the instance from the denominator — silently rewarding a model that hallucinates a tool. This catches it and returns a scored `[Tool error: …]` result, mirroring the existing `MaxTurnsExceeded` branch. In a real 50-instance run, 13 instances hallucinated a tool name; the fix keeps the denominator at 50/50 instead of inflating the score.

## Tool-call parser detection

`inference/providers/vllm_server_utils.py` now infers the tool-call parser from the model chat template (`<function=` present → `qwen3_coder`) instead of a brittle model-name check, keeping the existing name-based fallback and honoring the configured `trust_remote_code` / `revision`. This matters for models whose chat template emits XML-style tool calls (e.g. Qwen3.5-9B): the default `hermes` parser silently drops those, so no tool call executes.

## Self-correcting unknown tool calls

When the model calls a tool name that does not exist, the agents SDK raises `ModelBehaviorError` and the episode ends (with the fix above, scored 0 in the denominator). This is common in practice: in an ExpertQA probe, 3 of 5 instances died this way, and 1 of 10 even with tool names enumerated in the prompt. The scaffold now rewrites the unknown call to a hidden internal fallback tool whose result names the real tools, so the model can correct itself and continue; retries still consume the max_turns budget, and the fumble stays visible in the trajectory. The `ModelBehaviorError` catch remains as a last-resort backstop.

This applies to every task using the `openai_agents` scaffold, so agentic numbers may shift slightly upward relative to the old fail-the-instance behavior. Validated with unit tests (rewrite, hidden-tool wiring, reserved-name guard, kwargs forwarding) plus a live forced-hallucination probe showing the full correct-and-continue loop end to end.

### DIFF
--- a/src/olmo_eval/inference/providers/vllm_server_utils.py
+++ b/src/olmo_eval/inference/providers/vllm_server_utils.py
@@ -17,6 +17,7 @@ import sys
 import time
 from collections.abc import Callable, Generator
 from contextlib import contextmanager, suppress
+from functools import cache
 from typing import TYPE_CHECKING, Any
 
 from olmo_eval.common.debug import is_debug_provider
@@ -61,6 +62,29 @@ def _find_free_internal_port(exclude: set[int] | None = None) -> int:
 
 
 ProgressCallback = Callable[[str], None]
+
+
+@cache
+def _chat_template_has_function_tag(
+    model_name: str, trust_remote_code: bool = False, revision: str | None = None
+) -> bool | None:
+    """Return whether the model chat template contains '<function=', or None if unavailable."""
+    try:
+        from transformers import AutoTokenizer
+
+        tokenizer = AutoTokenizer.from_pretrained(
+            model_name, trust_remote_code=trust_remote_code, revision=revision
+        )
+        get_chat_template = getattr(tokenizer, "get_chat_template", None)
+        if callable(get_chat_template):
+            chat_template = get_chat_template()
+        else:
+            chat_template = getattr(tokenizer, "chat_template", None)
+        if chat_template is None:
+            return None
+        return "<function=" in str(chat_template)
+    except Exception:
+        return None
 
 
 def _wait_for_server(
@@ -154,7 +178,9 @@ def _wait_for_server(
     return False, last_error, None
 
 
-def _infer_tool_call_parser(model_name: str) -> str:
+def _infer_tool_call_parser(
+    model_name: str, *, trust_remote_code: bool = False, revision: str | None = None
+) -> str:
     """Infer the appropriate tool call parser based on model name.
 
     Args:
@@ -166,13 +192,18 @@ def _infer_tool_call_parser(model_name: str) -> str:
     model_lower = model_name.lower()
     if "llama" in model_lower:
         return "llama3_json"
-    elif "mistral" in model_lower:
+    if "mistral" in model_lower:
         return "mistral"
-    elif "olmo" in model_lower:
+    if "olmo" in model_lower:
         return "olmo3"
-    else:
-        # Default for Qwen and other models
-        return "hermes"
+
+    tag = _chat_template_has_function_tag(model_name, trust_remote_code, revision)
+    if tag:
+        return "qwen3_coder"
+    if tag is None and any(k in model_lower for k in ("qwen3-coder", "qwen3.5", "qwen3.6")):
+        return "qwen3_coder"
+    # Default for Qwen (e.g. Qwen3-8B) and other models
+    return "hermes"
 
 
 def _get_vllm_python() -> str:
@@ -304,7 +335,11 @@ def _build_server_command(
     if enable_auto_tool_choice:
         cmd.append("--enable-auto-tool-choice")
         # Auto-detect parser if not specified
-        parser = tool_call_parser or _infer_tool_call_parser(model_name)
+        parser = tool_call_parser or _infer_tool_call_parser(
+            tokenizer or model_name,
+            trust_remote_code=bool(kwargs.get("trust_remote_code", False)),
+            revision=kwargs.get("revision"),
+        )
         cmd.extend(["--tool-call-parser", parser])
 
         # OLMo3 requires custom chat template for tool calling (only when patch is applied)
--- a/src/olmo_eval/inference/providers/vllm_server_utils.py
+++ b/src/olmo_eval/inference/providers/vllm_server_utils.py
@@ -197,11 +197,18 @@ def _infer_tool_call_parser(
     if "olmo" in model_lower:
         return "olmo3"
 
+    name_suggests_xml = any(k in model_lower for k in ("qwen3-coder", "qwen3.5", "qwen3.6"))
     tag = _chat_template_has_function_tag(model_name, trust_remote_code, revision)
     if tag:
         return "qwen3_coder"
-    if tag is None and any(k in model_lower for k in ("qwen3-coder", "qwen3.5", "qwen3.6")):
+    if tag is None and name_suggests_xml:
         return "qwen3_coder"
+    if tag is False and name_suggests_xml:
+        logger.warning(
+            "Model name %s suggests XML-style tool calls but its chat template does not contain "
+            "'<function='; using 'hermes'; pass tool_call_parser explicitly to override.",
+            model_name,
+        )
     # Default for Qwen (e.g. Qwen3-8B) and other models
     return "hermes"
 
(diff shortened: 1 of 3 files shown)


src/olmo_eval/inference/providers/vllm_server_utils.py
```diff
@@ -166,13 +192,18 @@ def _infer_tool_call_parser(model_name: str) -> str:
...
+    if "olmo" in model_lower:
         return "olmo3"
-    else:
-        # Default for Qwen and other models
-        return "hermes"
+
+    tag = _chat_template_has_function_tag(model_name, trust_remote_code, revision)
+    if tag:
+        return "qwen3_coder"
+    if tag is None and any(k in model_lower for k in ("qwen3-coder", "qwen3.5", "qwen3.6")):
```
donovanr: claude was nit-picky about this -- not sure if I agree but surfacing it:

> The name heuristic only runs when the template lookup returns None, not False. A named qwen3-coder/qwen3.5/qwen3.6 model whose template loads but doesn't contain the literal <function= (e.g. the tag is built via a jinja macro or split string) falls through to hermes — silently dropping tool calls, which is the exact failure this PR fixes. A false downgrade here is worse than a false upgrade, so consider consulting the name list on both tag is None and tag is False. Lower priority since it's a heuristic-miss edge case, but cheap to harden.

src/olmo_eval/inference/providers/vllm_server_utils.py
```diff
@@ -63,6 +64,29 @@ def _find_free_internal_port(exclude: set[int] | None = None) -> int:
 ProgressCallback = Callable[[str], None]
 
 
+@cache
+def _chat_template_has_function_tag(
```
donovanr, line 68: can you add a test for all the branches of the logic here?

src/olmo_eval/inference/providers/vllm_server_utils.py
yilunzhao: We looked at this closely and decided to keep the decision logic but make the mismatch loud (c4ee7ee). The probe has three states: True and False both mean the template actually loaded, and in that case the template seems the more trustworthy signal about what format the model emits (a retemplated Qwen fine-tune that emits JSON tool calls would break if the name forced qwen3_coder). The name list stays as a fallback for None only, when the template is unavailable. What we added: when the template loads without `<function=` but the name suggests XML, we log a warning naming the model and pointing at the explicit `tool_call_parser` override, so the downgrade is no longer silent. Happy to flip to name-wins-on-False if you would rather.

src/olmo_eval/inference/providers/vllm_server_utils.py
yilunzhao: Added `tests/providers/test_vllm_server_parser.py` in c4ee7ee: 16 tests covering every branch of both functions (template with and without the literal, the `.chat_template` attribute fallback, tokenizer load failure, `trust_remote_code` / `revision` forwarding, the three name shortcuts skipping the probe entirely, the qwen3 name fallback when the probe returns None, and caplog assertions that the new warning fires only in the False plus qwen-name branch).

yilunzhao:
Heads up @donovanr, I added a third change to this PR and it is a behavior-level one, so I would appreciate extra scrutiny here.

**What it does:** when the model calls a tool name that does not exist, the SDK used to raise `ModelBehaviorError`, which ended the episode (after the earlier fix in this PR, the instance stays in the denominator and scores 0). Now the scaffold rewrites the unknown call to a hidden internal fallback tool whose result tells the model the real tool names, so it can correct itself and continue. The `ModelBehaviorError` catch stays as a last-resort backstop.

**Why:** this failure mode is common in practice. In an ExpertQA probe run, 3 of 5 instances died this way; even with tool names enumerated in the prompt it was still 1 of 10. Real agent deployments feed the error back rather than terminating, and the retries still consume the max_turns budget, so nothing is free.

**What changes for existing evals:** this applies to every task running through the `openai_agents` scaffold (litsearch, SAGE, etc.). Instances that previously scored 0 on a hallucinated tool name can now recover, so agentic numbers may shift slightly upward. The rewritten call and the corrective message are recorded in the trajectory, so the fumble stays visible.

**Validation:** unit tests cover the rewrite, the hidden-tool wiring, a reserved-name guard, and SDK kwargs forwarding. Live checks: a 10-instance regression run showed no regressions, and a forced-hallucination probe confirmed the full loop end to end (unknown call rewritten, corrective message returned, model retried with the correct name and completed normally).

Happy to split this into its own PR instead if you would rather keep this one to the two smaller fixes.

tests/core/harness/test_scaffold.py
```diff
@@ -42,3 +101,251 @@ async def run(
 
         # Clean up
         del SCAFFOLD_REGISTRY["custom"]
+
+
+class TestOpenAIAgentsToolCorrection:
```
donovanr: This currently breaks the default CI test command because `agents` / `openai` are optional deps and are not installed in the default groups.

Could you guard these like the other optional-dep tests? For example, `tests/integration/conftest.py` uses `pytest.importorskip("vllm")` before constructing vLLM-only fixtures. Since this file also contains non-Agents scaffold tests, We may want to use either a class-level skip guard for TestOpenAIAgentsToolCorrection, or move these tests into a dedicated OpenAI Agents test file with module-level `pytest.importorskip("agents")`.

donovanr:
Not opposed to the design, but I do think this should be reviewed separately rather than bundled with the parser/error-accounting fixes. Could we split this into its own PR, so we can get tylerm's opinion?

> Heads up @donovanr, I added a third change to this PR and it is a behavior-level one, so I would appreciate extra scrutiny here.
> 
> **What it does:** when the model calls a tool name that does not exist, the SDK used to raise `ModelBehaviorError`, which ended the episode (after the earlier fix in this PR, the instance stays in the denominator and scores 0). Now the scaffold rewrites the unknown call to a hidden internal fallback tool whose result tells the model the real tool names, so it can correct itself and continue. The `ModelBehaviorError` catch stays as a last-resort backstop.
> 
> **Why:** this failure mode is common in practice. In an ExpertQA probe run, 3 of 5 instances died this way; even with tool names enumerated in the prompt it was still 1 of 10. Real agent deployments feed the error back rather than terminating, and the retries still consume the max_turns budget, so nothing is free.
> 
> **What changes for existing evals:** this applies to every task running through the `openai_agents` scaffold (litsearch, SAGE, etc.). Instances that previously scored 0 on a hallucinated tool name can now recover, so agentic numbers may shift slightly upward. The rewritten call and the corrective message are recorded in the trajectory, so the fumble stays visible.
> 
> **Validation:** unit tests cover the rewrite, the hidden-tool wiring, a reserved-name guard, and SDK kwargs forwarding. Live checks: a 10-instance regression run showed no regressions, and a forced-hallucination probe confirmed the full loop end to end (unknown call rewritten, corrective message returned, model retried with the correct name and completed normally).
> 
> Happy to split this into its own PR instead if you would rather keep this one to the two smaller fixes.

tests/core/harness/test_scaffold.py
yilunzhao: Done in 87efbec: moved `TestOpenAIAgentsToolCorrection` and its Agents-only helpers into a dedicated `tests/core/harness/test_openai_agents_scaffold.py` with module-level `pytest.importorskip("agents")` / `("openai")`, so the module skips cleanly when those optional deps are absent. `test_scaffold.py` now keeps only the non-Agents `TestGetScaffold` and no longer imports agents/openai. Verified: `test_scaffold.py` still collects with the deps blocked, the new file skips, and the full `tests/core/harness` suite passes.

yilunzhao:
Done! Split it out as #252, stacked on this PR so its diff shows only the self-correction (GitHub will retarget it to `main` once this merges). This PR is now just the two smaller fixes: parser inference + error accounting. #252 also has field data from the full runs (misspelled-tool recoveries on LitSearch, hallucinated-name rescues on SAGE) for extra scrutiny.

undfined (approved):
lgtm

PR #240, opened 2026-07-02
Deepscholar-Bench

donovanr:
## Description

Adds DeepScholar-Bench as a sandboxed external evaluation, registered as
`deepscholar_bench`.

The benchmark retrieves prior academic work, generates a related-work section,
and evaluates it across seven organization, coverage, importance, and citation
metrics.

## Implementation

- Pins the validated upstream DeepScholar and LOTUS revisions.
- Supports local vLLM and external API providers through generated LOTUS
  configuration.
- Supports arXiv retrieval plus optional Semantic Scholar and Tavily backends.
- Applies compatibility fixes for citation lookup and organization scoring.
- Persists generation and evaluation artifacts, including partial runs.
- Reports both:
  - upstream parsed-query aggregates;
  - fixed-denominator metrics, where failed or missing queries score zero.
- Reports generation and scoring completeness in result metadata.

## Supporting reliability fixes

DeepScholar’s long-running sandbox workflow exposed several startup and control
failures that are included here because successful benchmark execution depends
on them:

- Sandbox streaming-control commands bypass a wedged in-container swe-rex
  server for local Docker/Podman runtimes.
- Failed streaming polls are detected without confusing a normal in-progress
  exit code with container failure.
- Beaker status reporting is nonblocking and nonfatal.
- Isolated vLLM environments expose adjacent console tools through `PATH`.

## Validation

Successful full 63-query Beaker runs:

- OLMo-3 7B Instruct: `01KXEDME5H6SPB432WBM7XE5ES`
- Qwen3.5 9B: `01KXEVZVCJHQMCK336MDZHD5HX`

Local validation:

- Focused DeepScholar and infrastructure tests: 42 passed
- Full non-integration unit suite: 1,882 passed, 4 skipped
- Ruff formatting: passed
- Ruff lint: passed
- Type checking: passed

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring


donovanr:
@undfined this is ready for a review when you get a chance

src/olmo_eval/evals/external/benchmarks/deepscholar/eval.py
```diff
@@ -0,0 +1,724 @@
...
+        provider: InferenceProvider,
+        args: dict[str, Any],
+        output_dir: str | None = None,
+        container_runtime: str = "podman",
+    ) -> ExternalEvalResult:
+        start_time = time.time()
+        ds_args = DeepScholarArgs.from_dict(args)
+        all_output: list[str] = []
+
+        provider_url = getattr(provider, "base_url", None) or "http://localhost:8000/v1"
```
undfined: This fallback makes API-backed providers without a `base_url` look like local providers. For example, `LiteLLMProvider("gpt-4o")` gets `provider_url = "http://localhost:8000/v1"`, so `_is_local_provider()` returns true and the run fails the local sandbox health check before generation. I think that contradicts the comment below that external API models should keep upstream behavior. Can we distinguish “missing base_url” from “local default URL” before calling `_is_local_provider()`?

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
```diff
@@ -0,0 +1,235 @@
...
+from typing import Any
+
+S2_SEARCH_URL = "https://api.semanticscholar.org/graph/v1/paper/search"
+S2_FIELDS = "title,abstract,url,authors,year,publicationDate,externalIds"
+REQUIRED_COLUMNS = ["title", "url", "snippet", "query", "context", "date"]
+
+# Hard wall-clock budget for a single S2 search across retries. The full validation
+# encountered frequent 429 responses, so each search needs bounded backoff rather
+# than retrying indefinitely.
+S2_SEARCH_BUDGET_S = 45.0
```
undfined: Nit: `_SEC` is a bit better

undfined (approved):
Looks good! A couple considerations.

src/olmo_eval/evals/external/benchmarks/deepscholar/eval.py
donovanr: Fixed by deciding locality from the provider's actual `base_url` before the local fallback, so a provider without one stays external:

```
base_url = getattr(provider, "base_url", None)
is_local = self._is_local_provider(provider, base_url or "")
provider_url = base_url or "http://localhost:8000/v1"
```

Now `LiteLLMProvider("gpt-4o")` is treated as external (agentic search, no health check) while local vLLM still resolves as local. `VLLMServerProvider` always assigns `self.base_url` (passed URL or its managed server's), and passing `""` rather than `None` keeps the `_server` check intact for a managed provider that hasn't set one yet.

Added a regression test covering all three states. Note the same pattern lives in `asta/eval.py` and `tau2/eval.py`; leaving those for a separate follow-up to keep this PR scoped and on track to merge quickly.

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
donovanr: Renamed to `S2_SEARCH_BUDGET_SEC`, and the `budget_s` param to `budget_sec` for consistency.

donovanr:
found a small formatting error for one of the sub-benchmarks, holding off on merging until that's added

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
```diff
@@ -0,0 +1,235 @@
...
+    """
+    import importlib
+
+    lm_attr = "LM"
+    lm_cls = getattr(importlib.import_module("lotus.models"), lm_attr)
+    init_name = "__init__"
+    original_init = getattr(lm_cls, init_name)
+
+    def init(self: Any, *args: Any, **kwargs: Any) -> None:
+        if kwargs.get("max_tokens") is None:
```
undfined, line 214: I think this also overrides the final generation budget. The pinned upstream `Configs` code removes `max_completion_tokens` before constructing `generation_lm`, so this wrapper supplies `stage_max_tokens` there as well. With the defaults, `max_tokens=10000` therefore becomes `4096`, making the advertised `max_tokens` option ineffective. Could we configure `generation_lm` explicitly with `ds_args.max_tokens`, while applying the stage budget only to the filter/search/taxonomize LMs?

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
```diff
@@ -0,0 +1,235 @@
...
+    unscorable. Non-arXiv S2 hits keep their S2 URL (usable as synthesis context,
+    but not counted as citations by this eval). The URL is never fetched, so this
+    does not reintroduce arXiv API calls.
+    """
+    title = paper.get("title") or ""
+    snippet = paper.get("abstract") or ""
+    # Emit a %Y-%m-%d date: downstream parses dates strictly and a bare year
+    # ("2006") raises, failing the search/filter step. Pad year-only to Jan 1.
+    date = paper.get("publicationDate") or ""
+    if not date and paper.get("year"):
```
undfined, line 57: This padding defeats the conservative same-year cutoff below. For example, an S2 result with only `year=2024` becomes `2024-01-01` and is accepted for a paper dated `2024-06-01`, even though the actual publication date is unknown and could be later. Could we preserve whether the source date was year-only and exclude same-year records before padding the date for downstream formatting?

undfined (approved):
2 suggestions before merging. Looks good otherwise!

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
donovanr: fixed

src/olmo_eval/evals/external/benchmarks/deepscholar/sandbox_search_shim.py
donovanr: fixed

PR #244, opened 2026-07-08
Compose CUDA_VISIBLE_DEVICES so concurrent runs don't collide on physical GPUs

yilunzhao:
## Summary

Concurrent `olmo-eval run` processes that each set a distinct external `CUDA_VISIBLE_DEVICES` collide on the same physical GPUs, so you cannot partition the machine across independent runs (e.g. 4 models, 2 GPUs each, on an 8-GPU host). This composes the planner indices through the inherited mask so each process stays on the GPUs it was given.

## Root cause

The GPU planner allocates indices relative to the process's visible set (`range(torch.cuda.device_count())`), then the worker writes them straight into the env:

```python
os.environ["CUDA_VISIBLE_DEVICES"] = ",".join(str(g) for g in gpu_ids)
```

CUDA always interprets `CUDA_VISIBLE_DEVICES` against the physical device ordering. So a process launched with an outer mask of `2,3` gets planner indices `[0,1]`, writes `"0,1"`, and lands on physical GPU 0,1 instead of 2,3. Two such processes both target physical 0,1 and one OOMs.

## Fix

New helper `resolve_visible_devices(gpu_ids)` in `gpu_planner.py`: if the process already has a `CUDA_VISIBLE_DEVICES` mask, map each relative index through it (`[0,1]` under mask `2,3` becomes `2,3`); otherwise emit the indices unchanged. Used at both sites that write `CUDA_VISIBLE_DEVICES` from planner indices (`workers.py`, `vllm_server_utils.py`).

The no-outer-mask path is byte-identical to before, so the normal single-process multi-worker allocation is unchanged.

## Verification

Two concurrent Qwen3.5-9B runs (tensor-parallel 2 each), launched with `CUDA_VISIBLE_DEVICES=0,1` and `CUDA_VISIBLE_DEVICES=2,3`:

- before: both loaded onto physical GPU 0,1; one crashed with CUDA OOM
- after: run A on physical GPU 0,1, run B on physical GPU 2,3, no collision

Tests in `tests/inference/test_gpu_planner.py` cover the identity (no-mask) non-regression, composition, reordered masks, whitespace, empty mask, and the out-of-range guard. `pytest tests/inference tests/runners` passes (99).

## Note

This touches the shared worker/server GPU setup, so flagging for your eyes. cc @roryd

### DIFF
--- a/src/olmo_eval/inference/gpu_planner.py
+++ b/src/olmo_eval/inference/gpu_planner.py
@@ -2,11 +2,42 @@
 
 from __future__ import annotations
 
+import logging
+import os
 from dataclasses import dataclass, field
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from olmo_eval.inference.providers.config import ProviderConfig
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_visible_devices(gpu_ids: list[int]) -> str:
+    """Map planner-relative GPU indices to a physical CUDA_VISIBLE_DEVICES string.
+
+    The planner allocates indices relative to the process's visible GPU set
+    (range(torch.cuda.device_count())). If the process already has a
+    CUDA_VISIBLE_DEVICES mask, those relative indices must be composed through it
+    so a worker does not clobber an outer mask; otherwise they are physical already.
+    """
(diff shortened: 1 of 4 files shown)


undfined (approved):
lgtm, thanks for the patch!

PR #249, opened 2026-07-14
Preserve agent work on max_turns: keep partial trajectory and force a tool-free final answer

yilunzhao:
## Problem

When an agent hits its `max_turns` budget, the `openai_agents` scaffold catches `MaxTurnsExceeded` and returns an **empty trajectory** plus `"[Max turns exceeded]"`. Every search and tool result the model actually produced is discarded, so the instance auto-scores 0 regardless of how much correct work was done. A model that searched hard for the full budget but did not finalize scores the same as one that did nothing.

Measured prevalence (share of instances that hit the cap and were zeroed): LitSearch 11-13% for the Qwen models; SAGE short_form 22-39% across Qwen3.5-9B/35B and Gemma.

## Fix

On `MaxTurnsExceeded`:
1. **Preserve the partial trajectory** from the exception's `run_data.new_items` (the SDK attaches `RunErrorDetails`), so scorers that read tool results (e.g. litsearch) see the searches that actually happened.
2. **Force a final answer**: run one tool-free generation (`tools=[]`, `tool_choice="none"`, `max_turns=1`) on the accumulated conversation plus an instruction to answer now from the gathered evidence, and use that text as `final_output` (for scorers that read the answer, e.g. sage/expertqa).
3. `max_turns_reached=True` is kept so downstream still knows the cap was hit.

**Never worse than before:** if the finalization call raises for any reason, the handler falls back to the previous empty result unchanged. The normal (non-capped) path is untouched and stays on non-streaming `Runner.run`.

## Validation

- Unit tests (new `tests/core/harness/test_openai_agents_scaffold.py`): capped run preserves tool results in the trajectory and returns the forced answer; finalization failure falls back to the old result; normal completion unchanged.
- End-to-end: re-ran only the capped instances of prior LitSearch/SAGE runs with this fix and merged by instance id. Recovered scores: LitSearch found_rate 9B 0.310 -> 0.356, 35B 0.320 -> 0.358; SAGE short_form exact_match 9B 0.45 -> 0.69, 35B 0.56 -> 0.76 (finished-run accuracy was ~90% all along; the cap discard was hiding it). SAGE open_ended barely moves (few caps), as expected.
- `ruff check` / `ruff format --check` clean; `ty` reports only pre-existing diagnostics.

cc @donovanr

### DIFF
--- a/src/olmo_eval/harness/scaffolds/openai_agents.py
+++ b/src/olmo_eval/harness/scaffolds/openai_agents.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 from collections.abc import Sequence
 from contextvars import ContextVar
+from dataclasses import replace
 from typing import TYPE_CHECKING, Any
 
 from olmo_eval.common.types import LMOutput, LMRequest, SamplingParams
@@ -22,6 +23,10 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 _current_binding: ContextVar[ExecutorBinding | None] = ContextVar("_current_binding", default=None)
+FORCED_FINAL_ANSWER_INSTRUCTION = (
+    "You have reached the maximum number of steps. Based on the information gathered so far, "
+    "provide your final answer now. Do not call any tools."
+)
 
 
 @register_scaffold("openai_agents")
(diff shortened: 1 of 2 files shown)


undfined (approved):
Nice!

PR #250, opened 2026-07-14
Add an in-process crawl4ai browse_webpage tool (+ presets, configurable fetch truncation)

yilunzhao:
## What

- New `browse_webpage` tool backed by [crawl4ai](https://docs.crawl4ai.com) (in-process headless browser, no hosted scrape service): lazy import, http(s)-only URL guard (rejects `file://` and other schemes before the browser starts), error sentinels mirroring `serper_fetch_webpage_content`, and the crawler's `error_message` surfaced on failures.
- Two presets: `dr_tulu_crawl4ai` (S2 + Serper search + crawl4ai browse) and `web_search_agent_crawl4ai` (Serper search + crawl4ai browse). Serper is still used for search; crawl4ai replaces only the fetch step and needs no API key.
- `OLMO_WEBPAGE_CONTENT_LIMIT` env var makes the fetched-page truncation limit configurable for BOTH fetch tools (default 4000 chars, `0` disables truncation).
- Packaging: `crawl4ai` optional extra; `crawl4ai.**` added to ty `allowed-unresolved-imports` so CI type-checking passes without the optional package installed.

## Why crawl4ai

- **Parity with DR Tulu's actual tool stack**: [open-instruct](https://github.com/allenai/open-instruct) ships `Crawl4AIBrowseTool`, exposing a `browse_webpage` tool backed by crawl4ai for DR Tulu. This PR mirrors that tool (same name, same backend), so models trained against open-instruct's agentic stack see the tool they were trained with at eval time.
- **No hosted-scrape dependency for fetch**: `serper_fetch_webpage_content` is credit-metered; we observed a mid-run quota exhaustion silently zeroing 33/50 instances of a baseline. crawl4ai fetches in-process with no key and no per-page cost.
- **Better page content**: real browser rendering (JS pages work) plus clean markdown extraction, instead of raw scrape text.
- **Enables fetch-length ablations**: with `OLMO_WEBPAGE_CONTENT_LIMIT`, on ExpertQA 32K-char fetches improved citation grounding for every searching model vs 8K (Gemma-4-26B +0.115 global_avg).

## Usage

```bash
pip install 'olmo-eval[crawl4ai]'
crawl4ai-setup   # one-time: provisions the headless browser

OLMO_WEBPAGE_CONTENT_LIMIT=32000 \
olmo-eval run -m Qwen/Qwen3.5-9B -H web_search_agent_crawl4ai -t expertqa -o limit=50
```

`SERPER_API_KEY` is still required (search); `S2_API_KEY` is recommended for `dr_tulu_crawl4ai` (keyless S2 is rate-limited to ~1 req/s process-wide).

## Tests

31 passing; crawl4ai fully mocked so CI needs no browser. Covers scheme rejection, error fidelity, truncation default/env/disable/invalid/boundary, and preset wiring.

## Note

Full SSRF hardening (DNS/redirect validation for private ranges) is out of scope; the scheme guard removes the local-file vector. Companion to #238/#239; used by the ExpertQA rewrite PR (linked next).

cc @donovanr

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,9 @@ Repository = "https://github.com/allenai/olmo-eval"
 Issues = "https://github.com/allenai/olmo-eval/issues"
 
 [project.optional-dependencies]
+crawl4ai = [
+    "crawl4ai>=0.8",
+]
 hf = [
     "transformers>=5.4.0",
 ]
@@ -239,6 +242,7 @@ allowed-unresolved-imports = [
     "cached_path.**",
     "instructions_registry.**",
     "instructions.**",
+    "crawl4ai.**",
 ]
 
 [tool.gantry]
--- a/src/olmo_eval/harness/presets.py
+++ b/src/olmo_eval/harness/presets.py
@@ -123,6 +123,34 @@ class HarnessPresets:
         )
 
     @lazy
+    def dr_tulu_crawl4ai(name: str) -> HarnessConfig:
+        """Dr. Tulu-style S2 plus web-search harness that browses with in-process
+        crawl4ai.
+
+        Install with ``pip install 'olmo-eval[crawl4ai]'``, then run
+        ``crawl4ai-setup`` once to provision the headless browser. ``S2_API_KEY``
+        is not required but is strongly recommended for throughput because
+        keyless S2 is rate-limited to about 1 request/second process-wide.
+        """
+        from .tools.search import crawl4ai_browse, semantic_scholar_search, serper_web_search
+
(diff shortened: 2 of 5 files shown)


undfined (approved):
Ok so this one is a bit more contentious. On the one hand the surface area for the crawlable urls is constrained by the evaluation datasets but this has the potential to cause havoc if a bunch of jobs are ran in parallel with the same seed such that start hammering a specific domain with a shared ip cidr range from our network. Probably ok but let's be cautious about the level of concurrency when running this particular setup on our infra.

PR #253, opened 2026-07-14
Date-constrain agentic retrieval to a per-instance cutoff

yilunzhao:
## What
Lets a task require that agentic retrieval surface only sources published on or before a per-instance
date. A per-instance cutoff (`Instance.metadata["retrieval_date_cutoff"]`) is threaded through the
asynq runner and the openai_agents scaffold into the shared search tools via a `ContextVar` (same
pattern as the scaffold's existing `_current_binding`).

## How
- `semantic_scholar_search`: sends S2's `publicationDateOrYear=:<ISO>` range + `publicationDate` field,
  and client-filters papers by full date (day-precise; year-only fallback for papers S2 has no full
  date for; undated dropped).
- `serper_web_search`: adds Google's `tbs` custom date range to the request.
- `serper_fetch_page`: unchanged (the URL is chosen from already-filtered search results).
- Default (no cutoff): every tool call is byte-identical to today, so all other tasks/presets are
  unaffected.

## Why
Faithful reproduction of date-constrained benchmarks. E.g. ResearchQA queries carry a survey cutoff D;
the paper constrains retrieval (date-constrained Google Scholar). Without this, agents cite work
published after D (a smoke showed ~8/10 answers citing post-D papers); with it, search results carry
zero post-D papers.

## Validation
- Unit: parse robustness, contextmanager nesting/exception-safety, **concurrency isolation** (two
  concurrently-live cutoffs don't leak across asyncio tasks), S2 day-precise filter (boundary kept,
  later/undated dropped), S2 no-cutoff invariance, serper tbs presence/absence. pytest 6.
- Behavioral (4-model researchqa:mini smoke): 0 over-cutoff paper-years across all search-tool results;
  answer-level post-D mentions dropped ~8/10 -> ~0-1/10 instances, residual traced to live fetched-page
  text / parametric memory (never a filtered search result). Coverage not degraded (no over-drop).
- Gates: ruff/format/ty green.

Independent of the researchqa task PR (disjoint files); the two compose when both land.

### DIFF
--- a/src/olmo_eval/harness/scaffolds/openai_agents.py
+++ b/src/olmo_eval/harness/scaffolds/openai_agents.py
@@ -14,6 +14,7 @@ from olmo_eval.harness.config import HarnessConfig
 from olmo_eval.harness.result import HarnessResult
 from olmo_eval.harness.scaffolds import Scaffold, register_scaffold
 from olmo_eval.harness.tools import Tool
+from olmo_eval.harness.tools.search import search_date_cutoff
 from olmo_eval.inference.base import InferenceProvider
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


donovanr (approved):
looks ok to me

PR #254, opened 2026-07-14
Add ResearchQA scholarly-QA benchmark (survey-mined rubric coverage)

yilunzhao:
## What
Adds ResearchQA as an agentic long-form scholarly-QA task. The model searches + reads sources,
then answers a research question (~250 words); the answer is scored by per-answer rubric
Coverage % against survey-mined rubric items. Registers `researchqa` (test) and `researchqa:mini`
(test_mini, 776 domain-balanced instances). Two new files only: `researchqa.py` (+461),
`test_researchqa.py` (+388).

- Paper: https://arxiv.org/abs/2509.00496
- Dataset: https://huggingface.co/datasets/realliyifei/ResearchQA
- Official code: https://github.com/realliyifei/ResearchQA

## Scoring (faithful port of the official scorer)
- Each rubric item rated by an LLM judge on the official 5-point ordinal {Not at all..Completely}
  using the verbatim official prompt, normalized `(x-1)/4`; items batched <=8/call; per-answer mean,
  then mean over instances. Per-rubric-type coverage tiers also reported.
- Deliberate deviation: an unparseable judge batch scores its items 0.0 and the instance is KEPT
  (official drops it), for a stable denominator.

## Answer extraction
- `extract_answer` strips a leading `<think>...</think>` reasoning block before judging (mirrors
  ExpertQA). Reasoning models (Qwen family) emit the block in the answer channel; without stripping
  the judge scored ~250 words of planning scaffold as part of the answer. On a 4-model smoke this
  corrected a model-order artifact (pre-strip the reasoning scaffold and max_turns caps understated
  the Qwen models specifically).

## Judge default
Defaults to `gpt-5.4-mini`; set `OLMO_EVAL_JUDGE=gpt-4.1-mini` to reproduce the paper's leaderboard.
Documented on the task class.

## Date-constrained retrieval
Each instance exposes `retrieval_date_cutoff`; when the date-constrained-retrieval harness change is
present, retrieval is limited to sources on/before the survey date (the paper uses date-constrained
Google Scholar). Harmless no-op without that change.

## Validation
Real-data smoke on `researchqa:mini` (4 models): schema/rubric parse over all 776 instances; judge
sanity (strong->1.0, garbage->0.0, 0 parse failures); agentic runs search + answer; one instance
independently re-judged and recomputed by hand (matched to one ordinal notch). Gates: ruff/format/ty
+ pytest 18.

### DIFF
--- a/src/olmo_eval/evals/tasks/researchqa.py
+++ b/src/olmo_eval/evals/tasks/researchqa.py
@@ -1,0 +1,461 @@
+"""ResearchQA agentic scholarly question answering (arXiv 2509.00496).
+
+The model searches the scholarly web and writes an approximately 250-word
+answer. Coverage is judged with the verbatim prompt from the official
+``compute_coverage.py`` scorer: rubric items are sent in batches of at most
+eight and receive one of five ordinal labels, normalized onto a 0--1 scale.
+The primary metric is an unweighted mean of per-answer rubric means. The paper
+reports this value multiplied by 100; for example, its best system's 75.29%
+coverage is ``0.7529`` here. Per-type metrics pool rubric items across answers,
+so answers with more items of a type receive proportionally more weight.
+
+Unlike the official scorer, which drops an entire answer after three malformed
+judge outputs, this task assigns the affected rubric items 0.0 and keeps the
+answer in the denominator. This prevents parse failures from inflating scores.
+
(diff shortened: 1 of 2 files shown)


donovanr (approved):
looks solid

PR #262, opened 2026-07-22
Add DeepResearch Bench (RACE report quality + FACT citation trustworthiness)

yilunzhao:
Adds **DeepResearch Bench** ([arXiv 2506.11763](https://arxiv.org/abs/2506.11763), [official repo](https://github.com/Ayanami0730/deep_research_bench), Apache-2.0): 100 PhD-level deep-research tasks (50 en / 50 zh, 22 topics). The model writes a citation-rich research report with web tools; scoring applies both official frameworks:

- **RACE** (primary, `race_overall` + 4 dimension metrics): judge compares the report against the official reference article under the task's pre-generated weighted criteria (verbatim bilingual prompts; score math ported faithfully incl. fuzzy criterion matching and `target/(target+reference)` normalization). Judge default `gpt-5.5` at the official `score` stage effort (medium).
- **FACT** (`fact_citation_accuracy`, `fact_avg_citations`, `fact_avg_effective_citations`): extract (fact, url) pairs -> group/dedup -> scrape -> support judgment, with the official stat semantics (unknown excluded; averages divide by citation-bearing tasks). Judge default `gpt-5.4-mini` at the official `fact` stage effort (low). Both judges overridable via `OLMO_EVAL_JUDGE` (`model` or `model:effort`).

**Dataset**: `allenai/deepresearch-bench` (configs `default`/`en`/`zh`), built by `scripts/internal/build_deepresearch_bench_dataset.py` from the official repo at pinned commit `469cce5`; `deepresearch_bench:en` selects the English subset.

**Deviations from the official pipeline** (documented in the module docstring): crawl4ai scraping instead of Jina Reader; deterministic report sanitation instead of the optional LLM ArticleCleaner; judge override env; the report-generation instruction is ours (official leaves system prompting unspecified); unconditional `clean_urls` + fragment-normalized grouping; 3 judge attempts (official RACE uses 10) with zero-and-keep-in-denominator on final failure.

## Hardening from a real 200-instance audit

Auditing a full 4-model x 50-task campaign drove three follow-up commits:

- `8bdb828` — deterministic sanitizer for model scaffold in scored reports: gemma-4 channel markers appeared in 50/50 of its reports (three "reports" were nothing but markers yet were judged), and one Qwen report ended in dangling tool-call tags. Also made the citation instruction copy-resistant (models echoed the literal example anchor text) and `clean_citation_url` now accepts CommonMark angle-bracket destinations. Validated in both directions against all 200 generations: exactly the 51 scaffold-affected reports change (pure deletions, zero legitimate prose lost) and a post-clean token census finds no remaining scaffold.
- `9cf070e` — FACT robustness: malformed validation items (missing `idx`/`result`) retry then fall back to all-unknown instead of erroring the instance (the official pipeline crashes outright on that input), and an unexpected FACT failure zeroes only the FACT channels instead of discarding the instance's already-computed RACE scores.
- `786e4c0` — judge reasoning effort mirrors the official stage configuration (`score`→medium, `fact`→low, from the reference `utils/api.py`) unless an explicit `:effort` suffix overrides it; the FACT stage is ~90% of judging volume, so this also cuts scoring cost substantially.

**Tests**: 91 unit tests, incl. SHA-256 pins of all 8 official prompt constants, differential fixtures for the RACE math, official-semantics tests for dedup/validation fallback paths, FACT denominator tests, and production-path sanitizer regressions.

## Campaign results

`deepresearch_bench:en`, all 50 tasks, `web_search_agent_crawl4ai`, `OLMO_EVAL_JUDGE=gpt-5.6-luna`:

| model | n | RACE overall | FACT citation acc | citations / effective | tools per task |
|---|---|---|---|---|---|
| Qwen3.5-35B-A3B | 23/50* | 0.448 | 0.849 | 20.4 / 17.3 | 14 search + 18 fetch |
| Qwen3.5-9B | 50/50 | 0.422 | 0.778 | 25.9 / 20.1 | 17 search + 16 fetch |
| gemma-4-26B-A4B | 46/50* | 0.393 | 0.824 | 6.4 / 5.3 | 2 search + 1 fetch |
| Olmo-3-7B | 50/50 | 0.387 | 0.139 | 4.2 / 0.6 | 0 (memory-only) |

\*35B and gemma tails were cut short by a judge-API quota outage; affected instances are excluded (not zero-filled). All 200 generations are archived and re-scorable offline without GPUs or search credits.

The FACT axis is doing exactly what the benchmark intends: Olmo writes fluent reports (RACE within 0.05 of the Qwens) but invents citations — 0.6 verifiable citations per task vs 17-20 for the Qwens.

## Repro

```
pip install 'olmo-eval[crawl4ai]' && crawl4ai-setup
olmo-eval run -m Qwen/Qwen3.5-9B -H web_search_agent_crawl4ai \
  -o provider.max_model_len=65536 -t deepresearch_bench:en -o limit=10
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,6 +178,8 @@ select = [
 # Verbatim prompt templates from asta-bench reference implementation
 "src/olmo_eval/evals/tasks/astabench_sqa.py" = ["E501"]
 "src/olmo_eval/common/scorers/citation.py" = ["E501"]
+# Verbatim bilingual prompt templates from DeepResearch Bench
+"src/olmo_eval/evals/tasks/deepresearch_bench.py" = ["E501"]
 # Verbatim prompt templates and gold snippets from scicode-bench reference implementation
 "src/olmo_eval/evals/external/benchmarks/scicode/prompts.py" = ["E501"]
 # Embedded fewshot data with long string literals
--- a/scripts/internal/build_deepresearch_bench_dataset.py
+++ b/scripts/internal/build_deepresearch_bench_dataset.py
@@ -1,0 +1,247 @@
+"""Offline builder for the DeepResearch Bench dataset mirror.
+
+Downloads the official query, criteria, and cleaned-reference JSONL files at a
+pinned upstream commit (or reads them from ``--source-dir``), joins them by task
+ID, writes local JSONL configs, and can push them to the HF Hub. This script is
+intentionally not imported or run by tests.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+from collections.abc import Iterable
+from pathlib import Path
+from typing import Any
+from urllib.error import HTTPError, URLError
+from urllib.request import Request, urlopen
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s %(levelname)s %(message)s")
+logger = logging.getLogger("build_deepresearch_bench_dataset")
+
+DEEPRESEARCH_BENCH_REPO = "allenai/deepresearch-bench"
+UPSTREAM_COMMIT = "469cce54ea7f6a63c163d3d9fec879cf289ec484"
+UPSTREAM_RAW_ROOT = (
+    f"https://raw.githubusercontent.com/Ayanami0730/deep_research_bench/{UPSTREAM_COMMIT}"
+)
+EXPECTED_IDS = set(range(1, 101))
+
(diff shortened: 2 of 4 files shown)


yilunzhao:

|---|---|---|---|
| this branch | 34 | 1882 | 26 |

PR #267, opened 2026-07-29
Add FrontierScience benchmark

### LINKED ISSUE #263: [FEATURE] Add FrontierScience benchmark
## Feature Description

Add support for FrontierScience (OpenAI, Jan 2026), a benchmark of expert-level scientific reasoning tasks across physics, chemistry, and biology. It has two tracks: Olympiad (100 gold-set short-answer problems, numeric/algebraic/fuzzy-string gradable) and Research (60 gold-set open-ended problems graded against 10-point rubrics via a model judge).

Dataset: https://huggingface.co/datasets/openai/frontierscience

## Use Case

Existing science benchmarks (GPQA, MMLU) are near-saturated by frontier models. FrontierScience is unsaturated (top model GPT-5.2: 77% Olympiad, 25% Research), giving us headroom to differentiate frontier and OLMo-class models on expert scientific reasoning. The Research track's rubric-based grading also gives per-step diagnostic signal rather than just final-answer correctness.

## Proposed Solution

- Olympiad track: standard generative task with a model-judge equivalence checker (given attempted + reference answer → CORRECT/INCORRECT). Judge prompt is in the paper (Appendix B).
- Research track: rubric-based model judge that returns total points earned (0–10); default success threshold = 7/10. Judge prompt also in Appendix B.
- Both use a strong judge model (paper uses GPT-5 high effort); make judge model configurable.
- Report accuracy overall and split by subject (physics/chem/bio).

## Alternatives Considered

- Olympiad-only (skip rubric grading) — simpler, but loses the open-ended Research signal that's the more interesting/unsaturated part.
- Reusing an existing rubric-judge harness if we already have one (e.g. HealthBench-style) rather than building bespoke grading.

## Additional Context

- Paper: FrontierScience (arXiv:2601.21165)
- Scoring config from paper: Olympiad averaged over 20 trials, Research over 30 trials.
- Text-only, no browsing / tools.

donovanr:
## Description

closes #263

Adds [FrontierScience](https://arxiv.org/abs/2601.21165) (OpenAI, Jan 2026) — expert-level scientific reasoning across physics, chemistry, and biology. Unlike GPQA and MMLU, this eval is not saturated: the paper's top model (GPT-5.2) scores 77% Olympiad and 25% Research, so it has headroom to hillclimb on.

### Tasks

| Task | Size | Primary metric |
| --- | --- | --- |
| `frontierscience_olympiad` | 100 | `accuracy` |
| `frontierscience_research` | 60 | `success_rate` (>= 7/10 rubric points) |

**Olympiad** — short-answer problems at IPhO / IChO / IBO level. A model judge compares the attempted answer against the reference (a number, expression, chemical formula, compound name, or entity) and returns CORRECT / INCORRECT.

**Research** — open-ended PhD-level research sub-tasks. Each carries a rubric totaling exactly 10 points that credits intermediate derivations, not just the final result. A model judge tallies the points earned. The paper marks >= 7 points a success, which is the primary metric here; `rubric_score` (mean fraction of points earned) is reported alongside it because it is far more sensitive on a benchmark where frontier models score around 25%.

Both tracks report their primary metric overall and per subject (`accuracy_physics`, `success_rate_biology`, ...). These are metrics within one task rather than separate subject-filtered tasks, so subject slices cost no extra inference and no extra judge calls.

### Judge

Judge prompts are reconstructed from the paper's Appendix B listings verbatim. The paper judges with GPT-5 at high reasoning effort. Following the `researchqa` precedent, the project default is the newer `gpt-5.5:high`, overridable via `OLMO_EVAL_JUDGE`:

```bash
OLMO_EVAL_JUDGE=gpt-5:high      # reproduce the published numbers
OLMO_EVAL_JUDGE=gpt-5-mini      # cheap iteration
```

Judge strictness does move the scores, so cross-model comparisons need to fix one judge spec. Both tasks declare `required_secrets = ("OPENAI_API_KEY",)`.

Judging is run concurrently under `context.scoring_concurrency`, with three retries and backoff on an unparseable verdict. An output whose verdict never parses scores 0.0 and stays in the denominator rather than being dropped, so parse failures cannot inflate a score.

### Prompting and answer extraction

The dataset's `problem` field already ends with the benchmark's own answer-format instruction, so the generation prompt is that field verbatim with nothing prepended.

- Olympiad: the response is cut down to the text after the last `FINAL ANSWER` marker the dataset asks for (tolerating `**FINAL ANSWER**`, a trailing colon, and case variation), falling back to the whole visible response if the marker is absent. Sending the full chain of thought would let the judge match on any expression mentioned mid-derivation.
- Research: graded whole, since the rubric credits the derivation.

Both strip a closed `<think>` block first.

### Trial averaging

The paper averages Olympiad over 20 independent trials and Research over 30. `:paper` variants request those sample counts. This needed one small framework addition — see below.

```bash
uv run olmo-eval run -m <model> -t frontierscience_olympiad:paper
```

### Suites

Adds a `frontierscience` convenience suite and puts both tracks in `science:judge`, and therefore `science:all`. FrontierScience is text-only with no tools, so it needs a judge but not an agentic harness and fits the existing judge/nojudge execution split cleanly.

## Changes to shared code paths

Two files outside the new task directory have been changed, but neither change should alter the behavior of any other existing task.

### `evals/tasks/common/base.py` (+3 lines)

```diff
     MAX = "max"
     FIRST = "first"
+    MEAN = "mean"
```
```diff
         if aggregation == OutputScoreAggregation.FIRST:
             return ordered_scores[0]
+        if aggregation == OutputScoreAggregation.MEAN:
+            return sum(ordered_scores) / len(ordered_scores)
         raise ValueError(f"Unsupported output_score_aggregation: {aggregation}")
```

**Why needed:** the paper's protocol is to average each problem over independent trials (20 Olympiad, 30 Research). The existing default `MAX` reports each problem's *best* sample, which on a 25%-accuracy benchmark at 30 trials is a wildly different number. `FIRST` throws away 29 of 30 samples. Neither is the metric the paper defines.

**Why here instead of inside the task:** `_aggregate_output_scores` is the single hook every task's per-instance collapse goes through, driven by the declarative `config.output_score_aggregation`. Computing the mean locally inside `frontierscience.py` would mean the task silently ignores that config field — so `-t frontierscience_research --overrides output_score_aggregation=first` would appear to work and do nothing. Adding the enum member keeps the choice declarative and overridable, which is what the field is for.

**Why it is safe:** purely additive. The `MAX` and `FIRST` branches are untouched, so no existing task's aggregation changes. The `ValueError` at the bottom enumerates the enum dynamically (`", ".join(option.value for option in OutputScoreAggregation)`), and no test hardcodes that option list, so the new member needs no other updates.

### `common/scorers/llm_judge.py` (+16 / -8)

This function caches an adaptive request shape across calls: newer models want `max_completion_tokens` instead of `max_tokens` and reject a non-default `temperature`, older ones reject `reasoning_effort`. On rejection it flips a closure-level flag so subsequent calls send the corrected shape.

#### Altered retry block

```diff
-        for _ in range(3):
+        for _ in range(4):
+            token_param = _token_param
+            send_temperature = _send_temperature
+            send_reasoning_effort = _send_reasoning_effort
+
             kwargs: dict[str, Any] = {"model": model, "messages": messages}
-            kwargs[_token_param] = max_tokens
-            if _send_temperature:
+            kwargs[token_param] = max_tokens
+            if send_temperature:
```
```diff
-                if _send_reasoning_effort and "reasoning_effort" in message:
+                if send_reasoning_effort and "reasoning_effort" in message:
                     _send_reasoning_effort = False
```

- `_token_param` / `_send_temperature` / `_send_reasoning_effort` (underscore-prefixed, `nonlocal`) remain the **shared cache across calls**. They are still the only things *written* — the `_send_reasoning_effort = False` on the line after the changed condition is unchanged, so the adaptation still sticks for every later call. That caching is the whole point of the mechanism and is preserved.
- `token_param` / `send_temperature` / `send_reasoning_effort` (no underscore) are new **per-attempt snapshots**, taken at the top of each loop iteration. They are only ever *read*.

The `except` block has to answer "did the API reject a parameter **I** just sent?" Reading the shared flags answers a different question: "does the shared state still say we send that parameter?" Those diverge the moment two judge calls are in flight:

1. Call A and call B both make their first attempt with `max_tokens`.
2. Both get back `Unsupported parameter: 'max_tokens' ... Use 'max_completion_tokens' instead`.
3. A's `except` runs first: `_token_param == "max_tokens"` is true, so it adapts, sets `_token_param = "max_completion_tokens"`, and retries successfully.
4. B's `except` runs: `_token_param == "max_tokens"` is now **false**. No other condition matches either, so `adapted` stays `False` and B hits `if not adapted: raise` — surfacing a `BadRequestError` for an error the function knows exactly how to handle.

Reading the snapshot instead makes B's check `token_param == "max_tokens"`, which is true because that is what B sent, so B adapts and retries like A. The `kwargs` construction reads the same snapshot so the shape sent and the shape tested cannot drift across the `await`.

**Why no existing task hits this:** every current judge task (`researchqa`, `astabench_sqa`, `expertqa`, the safety judge) awaits its judge calls one at a time. FrontierScience is the first to fan them out, and it reproduced on the very first real-judge run against a gpt-5-series model.

I did consider fixing this task-side by warming the judge with one sequential call before fanning out. That papers over a bug that will surface on the next task to parallelize a judge, so I fixed the function.

#### `range(3)` -> `range(4)`

Three parameters are adjustable, and each fix costs one failed attempt. So the loop needs three failure slots plus one slot for the corrected request to actually be *sent*. With `range(3)`, a model rejecting all three consumes every iteration on failures, exits the loop, and hits `raise last_error` — the third fix gets stored in the closure but never used by the call that discovered it.

#### Test coverage for both

`tests/core/test_llm_judge_scorer.py` gains a `TestBuildOpenAIJudgeFn` class with a stubbed `openai` module (injected via `sys.modules` so it runs whether or not the `clients` extra is installed). The race test uses an `asyncio.Barrier(2)` to hold both first attempts in flight, making the interleaving deterministic rather than timing-dependent. I confirmed it fails against the pre-fix code and passes after, and there is a companion test that an unrelated error (`rate limit exceeded`) still propagates rather than being swallowed by the retry loop.

## Results

| Model | max_tokens | Olympiad accuracy | Research success_rate | Research rubric_score |
| --- | --- | --- | --- | --- |
| Qwen3.5-35B-A3B | 120,000 | 0.5400 | 0.0500 | 0.3094 |
| Gemma4 26B-A4B | 120,000 | 0.4300 | 0.0667 | 0.3119 |
| GPT-OSS-20b | 120,000 | 0.4300 | 0.0333 | 0.3103 |
| Nemotron 3 Nano 30B-A3B | 120,000 | 0.4200 | 0.0500 | 0.2957 |
| Qwen3.5 9B Instruct | 120,000 | 0.3500 | 0.0167 | 0.2391 |
| Olmo-3 7B Think | 60,000 | 0.1600 | 0.0000 | 0.1777 |
| GLM-4.1V-9B-Thinking | 60,000 | 0.1400 | 0.0000 | 0.1802 |
| OLMo-3 7B Instruct | 32,768 | 0.0500 | 0.0000 | 0.1983 |

These runs were generated with

```
uv run olmo-eval beaker launch \
    --yes --name fs-small \
    --model allenai/Olmo-3-7B-Instruct \
    --model allenai/Olmo-3-7B-Think \
    --model zai-org/GLM-4.1V-9B-Thinking \
    --task frontierscience_olympiad \
    --task frontierscience_research \
    --harness default \
    -o provider.max_model_len=40960 \
    -o provider.kwargs.startup_timeout=1800 \
    -o provider.kwargs.language_model_only=true \
    --cluster ai2/ceres --workspace ai2/olmo-eval-debug \
    --priority urgent --timeout 12h --gpus 1 --no-follow

uv run olmo-eval beaker launch \
    --yes --name fs-large \
    --model google/gemma-4-26B-A4B-it \
    --model nvidia/NVIDIA-Nemotron-3-Nano-30B-A3B-BF16 \
    --task frontierscience_olympiad \
    --task frontierscience_research \
    --harness default \
    -o provider.max_model_len=40960 \
    -o provider.kwargs.startup_timeout=1800 \
    -o provider.kwargs.language_model_only=true \
    -o provider.kwargs.tensor_parallel_size=2 \
    --cluster ai2/ceres --workspace ai2/olmo-eval-debug \
    --priority urgent --timeout 12h --gpus 2 --no-follow

uv run olmo-eval beaker launch \
    --yes --name fs2-qwen35 \
    --model Qwen/Qwen3.5-35B-A3B \
    --task frontierscience_olympiad -o max_tokens=120000 \
    --task frontierscience_research -o max_tokens=120
    --harness default \
    -o provider.max_model_len=131072 \
    -o provider.kwargs.startup_timeout=1800 \
    -o provider.kwargs.language_model_only=true \
    -o provider.kwargs.tensor_parallel_size=2 \
    -o provider.kwargs.gdn_prefill_backend=triton \
    --cluster ai2/ceres --workspace ai2/olmo-eval-debug \
    --priority urgent --timeout 12h --gpus 2 --no-follow

uv run olmo-eval beaker launch \
    --yes --name fs2-gptoss \
    --model openai/gpt-oss-20b \
    --task frontierscience_olympiad -o max_tokens=120000 \
    --task frontierscience_research -o max_tokens=120
    --harness default \
    -o provider.max_model_len=131072 \
    -o provider.kwargs.startup_timeout=1800 \
    -o provider.kwargs.language_model_only=true \
    --cluster ai2/ceres --workspace ai2/olmo-eval-debug \
    --priority urgent --timeout 12h --gpus 1 --no-follow

uv run olmo-eval beaker launch \
    --yes --name fs2-qwen9 \
    --model Qwen/Qwen3.5-9B \
    --task frontierscience_olympiad -o max_tokens=120000 \
    --task frontierscience_research -o max_tokens=120
    --harness default \
    -o provider.max_model_len=131072 \
    -o provider.kwargs.startup_timeout=1800 \
    -o provider.kwargs.language_model_only=true \
    -o provider.kwargs.gdn_prefill_backend=triton \
    --cluster ai2/ceres --workspace ai2/olmo-eval-deb
    --priority urgent --timeout 12h --gpus 1 --no-follow
```

## Type of Change

- [ ] Documentation update

## Testing

`1915 passed, 7 skipped`. Also clean on `ruff check`, `ruff format --check`, and `ty check src/ alembic/`.

**`tests/evals/tasks/test_frontierscience.py`** (62 tests): registration and metric wiring; `process_doc` against the real HF schema, including unique ids where `task_group_id` is not; judge prompts rendered against the paper's listings; LaTeX braces in substituted values surviving `str.format`; `FINAL ANSWER` extraction including markdown-wrapped and repeated markers; verdict parsing (case, markdown emphasis, backticks, `7/10`, out-of-range clamping, no verdict at all); the 7/10 threshold boundary at 6.9 / 7 / 7.1; subject metrics averaging only their own slice; retry-then-zero and retry-then-recover; multi-trial averaging; and a runner-shaped group covering judge concurrency bounded across both tracks, run-scoped limiter lifetime, partial vs total judge failure, and cancellation propagation.

**`tests/core/test_llm_judge_scorer.py`**: three tests for the judge-builder fix — see the shared-code section above for what they pin down.

**End-to-end against the live judge**: a correct olympiad answer scores 1.0, a wrong one 0.0, and a junk research answer earns 0/10 after the judge walks the rubric item by item.

One existing test needed updating: `test_science_judge_contains_judge_tasks` asserted an exact task set that the `science:judge` addition changes.

## Checklist

## Notes for review

Three things that might be worth a second opinion:

1. **Default judge is `gpt-5.5:high`, not the paper's `gpt-5:high`.** This follows the what we did with `researchqa`, which also deliberately defaults off-paper and documents the override. The trade-off is that out-of-the-box numbers will not match the published leaderboard. Could bump it to 5.6 at this point.
2. **`:paper` variants use `temperature=1.0`.** The paper says "20 independent trials" but never states a temperature. At 0.0 the trials would be identical, and 1.0 is the API default for the models it evaluates. This is a reconstruction, flagged in a comment.
3. **Subject reporting as metrics rather than tasks.** GPQA uses subject-filtered *tasks*; I used per-subject metrics inside one task to avoid paying for inference and judge calls four times over. Consistent with how `researchqa` slices per rubric type.

Cost note: a full run uses a lot of generation / judge calls, since we score 20 attempts on each olympiad questions and 30 attempts per research question. So the `:paper` variants are 2000 (Olympiad) and 1800 (Research) generation / judge calls. The default setting is just one generation / judge call per question, to save time / money.


src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -0,0 +1,640 @@
...
+    matches = list(_RESEARCH_VERDICT.finditer(raw))
+    if not matches:
+        return None
+    points = float(matches[-1].group(1))
+    return min(max(points, 0.0), FRONTIERSCIENCE_RUBRIC_TOTAL)
+
+
+def build_frontierscience_judge_fn(*, scorer_name: str, max_tokens: int) -> JudgeFn:
+    """Build the judge from ``$OLMO_EVAL_JUDGE`` or the task-local default."""
+    spec = os.getenv("OLMO_EVAL_JUDGE", FRONTIERSCIENCE_DEFAULT_JUDGE_SPEC)
```
undfined: Could we include the judge model, reasoning effort, and token budget in the serialized task configuration?

Right now `OLMO_EVAL_JUDGE` is read only at scoring time. Runs using `gpt-5:high` and `gpt-5.5:high` therefore have the same task hash and artifact filename, even though the judge choice can change the score. This also makes it difficult to tell which judge produced a stored result.

A dated model snapshot would also make the default more reproducible than the mutable `gpt-5.5` alias.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -0,0 +1,640 @@
...
+
+async def judge_with_retries(
+    judge_fn: JudgeFn,
+    prompt: str,
+    parse: Callable[[str], float | None],
+) -> tuple[float | None, str]:
+    """Call the judge until its verdict parses, returning the value and last reply."""
+    raw = ""
+    for attempt in range(FRONTIERSCIENCE_JUDGE_ATTEMPTS):
+        raw = await judge_fn(prompt)
```
undfined: Could we retry transient API errors here, such as rate limits, timeouts, and 5xx responses?

The current loop only retries responses with an unparseable result but not API errors from what I can tell. With one sample, the runner records the instance as zero. With multiple samples, the failed sample is excluded from the mean. In both cases, judge infrastructure changes the benchmark score.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -0,0 +1,640 @@
...
+    def _record(
+        self,
+        output: LMOutput,
+        *,
+        scores: dict[str, float],
+        raw_judge_response: str,
+        parse_error: bool,
+    ) -> None:
+        """Persist per-output judge details and score channels."""
+        output.metadata["frontierscience_raw_judge_response"] = raw_judge_response
```
undfined: These metadata fields are not included in saved predictions. `build_predictions()` currently keeps `score:*`, `scoring_errors`, and `execution_result`, but drops the raw judge response and parse-error flag.

It doesn't look like a parse failure can be distinguished from a genuine zero after the run. Could we persist at least the parsed outcome, rubric points, parse status, and judge configuration or something like this?

undfined (approved):
A few small considerations. Looks good!

src/olmo_eval/evals/tasks/frontierscience.py
donovanr: Judge model, reasoning effort, and token budget are now conditionally serialized in `TaskConfig`, so they affect the task hash and stored results without re-hashing non-judge tasks. `OLMO_EVAL_JUDGE` is resolved when the task is instantiated rather than during scoring. I also pinned the default to `gpt-5.5-2026-04-23:high` (verified that snapshot through the OpenAI Models API).

src/olmo_eval/evals/tasks/frontierscience.py
donovanr: Now each attempt wraps the judge call with `retry_with_backoff`, covering rate limits, timeouts, connection failures, and retryable 5xx responses while failing fast on non-retryable errors. API retries and parse retries have independent budgets, with tests for transient recovery and non-retryable behavior.

src/olmo_eval/evals/tasks/frontierscience.py
donovanr: Saved predictions now include a compact per-output judge_result: `Olympiad` stores the parsed outcome, `Research` stores rubric points, and both store parse status. Raw judge text is retained only when parsing fails to avoid bloating prediction artifacts, especially for `:paper` runs. Judge model, effort, and token budget are stored in the serialized task configuration in `results.json` and represented by the task hash/artifact filename rather than duplicated on every output.

PR #268, opened 2026-08-03
Make a failed run fail loudly: exit codes, version pinning, and launch diagnostics

yilunzhao:
Six independent fixes to how a run is launched, identified, and reported. Each one came out of a real failure during a multi-system evaluation sweep; none changes what any task measures.

| commit | problem it fixes |
|---|---|
| `abf5232` | A run that scored **zero** instances exited 0 and looked successful. A whole batch of experiments failed silently this way before it was noticed. |
| `5833684` / `170c41d` | The launched code version was neither pinnable nor visible. Two full sweeps were later found to have run a different commit than intended, and there was no record in the run itself to catch it. |
| `1c52aa2` | The isolated vLLM venv was not on the server subprocess `PATH`, so the server came up against the wrong interpreter. |
| `acf01ae` | Rich markup swallowed the dirty-worktree warning, so the one message telling you the launch was not reproducible never reached the console. |
| `dcf2361` | Semantic Scholar requests all went to the first configured key instead of spreading across every key, so concurrency was capped by one key's rate limit. |

## Verification

Test counts are stated against `origin/main` rather than as absolutes, because `main` is not currently green.

- `origin/main`: **36 failed, 1789 passed, 26 skipped**
- this branch: **36 failed, 1816 passed, 26 skipped**

The failing sets are **identical** — compared name by name, not just by count. Nothing is introduced and nothing pre-existing is fixed. The 27 additional passing tests are this branch's own.

`tests/analysis/test_eval_power.py` and `tests/analysis/test_pairwise.py` are excluded from both runs: they fail at collection on `numpy.in1d`, removed in NumPy 2.0, identically on untouched `main`.

## Scope

These commits were extracted one at a time from a longer working branch and each was applied to `main` on its own, so this branch contains nothing but the six fixes above. A seventh related commit (FlashInfer prebuilt kernels) is not here — it depends on the `python` provider work and belongs on top of that instead.

### DIFF
--- a/src/olmo_eval/cli/run/__init__.py
+++ b/src/olmo_eval/cli/run/__init__.py
@@ -8,6 +8,8 @@ Configuration parsing, storage setup, and runner creation are delegated to:
 - options.py: Option decorators for logical grouping
 """
 
+from typing import Any
+
 import click
 
 from olmo_eval.cli.run.options import (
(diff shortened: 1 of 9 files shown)


yilunzhao:

|---|---|---|---|
| this branch | 34 | 1818 | 26 |

PR #269, opened 2026-08-03
Build the ResearchQA prompt from the harness's real tool list

yilunzhao:
The ResearchQA prompt hardcoded an "Available tools" block naming `serper_google_webpage_search` and `serper_fetch_webpage_content`, and instructed the model to search and fetch with them.

Under `paper_search_agent`, which exposes only `semantic_scholar_snippet_search`, the model obeys the prompt over the tool schema and calls a tool that does not exist. The `openai_agents` scaffold then ends the episode with

```
[Tool error: Tool serper_google_webpage_search not found in agent openai_agents]
```

and an empty trajectory. Every instance scores near zero — not because the model is weak on the task, but because it never got to attempt it. This is the failure mode where a benchmark returns a plausible-looking low number while measuring nothing.

## The change

Carry the harness's real tool names into `TaskConfig`, and build the prompt's tool list and workflow steps from them. When the harness exposes no tools, fall back to the previous hardcoded list.

`dr_tulu` exposes exactly the three previously hardcoded tools, so its prompt and every non-agentic prompt stay **byte-identical**. A test pins that, so this cannot silently change the established configuration.

## Verification

Measured on Qwen3.5-35B-A3B with `paper_search_agent`, 10 instances:

- **before**: 10/10 answers were the tool-not-found error, `turns=[]`
- **after**: 10/10 are ~2k-character cited answers whose trajectories contain only `semantic_scholar_snippet_search` calls

Test counts are stated against `origin/main` rather than as absolutes, because `main` is not currently green:

- `origin/main`: **36 failed, 1789 passed, 26 skipped**
- this branch: **36 failed, 1792 passed, 26 skipped** (stable across two runs)

The failing sets are identical, compared name by name rather than by count. `tests/analysis/test_eval_power.py` and `tests/analysis/test_pairwise.py` are excluded from both: they fail at collection on `numpy.in1d`, removed in NumPy 2.0, identically on untouched `main`.

## Scope

One commit, extracted onto `main` on its own. The working branch it came from carries it behind a merge commit; the merge contributes no changes of its own, so only the content commit is here.

### DIFF
--- a/src/olmo_eval/evals/tasks/common/base.py
+++ b/src/olmo_eval/evals/tasks/common/base.py
@@ -165,6 +165,12 @@ class TaskConfig:
     #: beaker launcher mounts each as the user-scoped secret ``{user}_{NAME}``.
     required_secrets: tuple[str, ...] = ()
 
+    #: Tool names the configured harness actually exposes to the agent, empty when
+    #: it exposes none. The runner fills this in from the harness config so a task
+    #: whose prompt names tools can describe what is really callable instead of a
+    #: hardcoded list.
+    harness_tool_names: tuple[str, ...] = ()
+
     def __post_init__(self) -> None:
         """Validate scheduler-only sandbox allocation hints."""
         if isinstance(self.output_score_aggregation, str):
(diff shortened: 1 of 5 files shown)


yilunzhao:

|---|---|---|---|
| this branch | 34 | 1794 | 26 |

PR #275, opened 2026-08-06
Add StrongReject

mgmorgan23:
## Description

Add the StrongReject eval from the safety eval repository and include it in the safety suite. 
This PR also changes the default temperature for the safety suite tasks from 0.7 to 0.6, moving from parity with the original safety suite to parity with the other posttraining evals, and removes the hard-coding of add_special_tokens = False in the vllm_server provider so that it will honor the add_bos_token flag (which defaults to false, so this will only impact runs that directly set it).

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update

## Testing

- [X] Unit tests pass locally (`pytest tests/ --ignore=tests/integration/`)

## Checklist

- [X] My code follows the project's style guidelines
- [X] I have performed a self-review of my code
- [X] I have added/updated documentation as needed
- [X] My changes generate no new warnings
- [X] Any dependent changes have been merged and published


mgmorgan23:
This PR adds StrongReject to the safety suite with a thinking, instruct, and base version. Final parity testing is here: https://docs.google.com/spreadsheets/d/1lZo3lhDbVjXhDOgGxT4gZm50wJZGcBNox7nfLEpS_5o/edit?gid=1089743412#gid=1089743412. 

The biggest infrastructure change was adding `enable_lora` and `lora_modules` to the vllm_server provider in order to load and run the StrongReject judge

abhishekraok:
The new task looks sound, but a few things change existing numbers or invert scoring:

1. WMDP scores unparseable output 1.0 (wmdp.py:96). A model that never emits a clean A–D letter now reads as maximally safe (was 0.0). Base models with max_tokens=1 will score near 100%.
2. StrongReject scores empty responses 0.0 — i.e. fully jailbroken — when an empty answer should grade as a refusal (~1.0). Also the opposite convention from the WMDP change above.
3. Every existing judge variant's numbers silently shift, four ways: judge_sampling temp 0.7→0.6; wildguard judge now sees vanilla_prompt instead of the prompt the model answered (DAN/wildjailbreak); dropping truncate_prompt_tokens=32600 means long thinking responses can overflow the judge's 32k context → HTTP 400 → scored 0.0; and add_bos_token now sets add_special_tokens on all string completions. If intentional, please call these out in the description — pre/post-PR runs won't be comparable.
4. strongreject:openai_judge doesn't run the StrongReject rubric — it falls back to the standard yes/no judge grading against the encoded (base64/rot13) prompt, so that variant is meaningless for exactly the subsets the benchmark targets.
5. LoRA routing: all requests go to lora_modules[0]; base model and other adapters are unreachable. Consider a single lora_module arg or raising on >1.

mgmorgan23:
Thank you so much! I made some modifications, please review and let me know if everything is addressed!

-  This is intentional – WMDP captures whether a model has dangerous dual-use knowledge so unparsable answers and refusals are all desired responses (matches safety-eval implementation). Parsing errors are still captured by the SafetyErrorMetric which reports separately. The base models should be evaluated on the base suite, where WMDP uses logprobs. The scorer had some documentation for this but I explicitly added that this differs from the other tasks
- This is intentional to be in line with previous implementations of strong-reject – the safety evals (with the exception of WMDP) penalize empty responses. This was set up on safety-eval to punish degenerate reasoning and is expanded slightly here to include any model errors that return blank responses.
- The judge_sampling was intentional -- the safety repository used a temperature of 0.7 but all other post-training evals were done with a temperature of 0.6, so this brings the safety suite in line with that. The vanilla_prompt change was to correct an earlier mistake in the implementation -- they now match the old safety suite. add_bos_token was added to safety suite calls in order to ensure the local judges are loaded correctly -- I verified that the token level requests sent to the judges matched the original from the safety suite only when it was set, and the strong reject judge (which is based on a gemma model) especially had degenerate output when it was not set. It was intended only to impact the local auxiliary judge. The truncation on the Wildguard judge was not removed -- it was moved to the prompt formatting stage to ensure the truncation was only on the model response and not on the prompt formatting. I verified that the longest prompt that hits Wildguard is 1277 tokens (WildGuard formatting is 116), and the truncation leaves 2640 tokens of space
- The format (not using the rubric) was intentional so the openai judge implementations all matched each other. I added a prompt catch so the openai judges now use vanilla/forbidden prompts if provided instead of encoded ones to bring it more in line with the wildguard and strongreject judges.
- I added an error call if more than one LoRa layer is passed!

abhishekraok (approved):
Thanks! One small ask: note the 0.7→0.6 temperature change and the add_bos_token→add_special_tokens widening in the PR description, since anyone with external configs or pre-PR baselines will see a shift. Otherwise LGTM.

mgmorgan23:
Thank you! I added the comments to the PR description

PR #276, opened 2026-08-10
Add ruler-plus extending RULER evals to 2m tokens

jopetty:
Adds ruler_lc_* tasks and suites mirroring the existing RULER tasks/suites but reading data from a local Weka path which has context sizes up to 2097152 tokens (vs. the original 131072 cap).

Demo run: https://beaker.org/orgs/ai2/workspaces/linear-rnns/work/01KZPE3PYQN125A0Q52SMR3QJK?taskId=01KZPE3PYXRXJG0DASG01GYB6G&jobId=01KZPE3Q2C997APBS38DFVEA9N


src/olmo_eval/data/ruler_lc_loader.py
```diff
@@ -0,0 +1,33 @@
...
+ruler_loader.load_ruler_dataset is reused as-is once the data root is resolved.
+"""
+
+import os
+
+RULER_LC_DATA_DIR_ENV_VAR = "OLMO_EVAL_RULER_LC_DATA_DIR"
+
+# Ai2 cluster path where jopetty/RULER's generate-data.sh was run with its
+# default output_dir (see olmo-eval's RULER long-context integration notes).
+_DEFAULT_RULER_LC_DATA_DIR = "/weka/oe-training-default/jacksonp/RULER/scripts/benchmark_root/data"
```
undfined: Let's consider one of two options:

1) Rename the task in a way that indicates it is internal only given the data lives on our private storage layer or...
2) Rework this to make it easy for someone to generate the data and point it at their local path

I kind of like the latter but for expediency 1 is probably better. If you go that route can add a task to come back to this later?

src/olmo_eval/evals/tasks/ruler_lc.py
```diff
@@ -0,0 +1,63 @@
+"""RULER long-context (ruler-lc) tasks.
```
undfined: Nit: Maybe consider a name like ruler-plus or something similar since ruler _is_ a long-context bench.

undfined (approved):
Overall lgtm! 2 things to consider in comments.

src/olmo_eval/evals/tasks/ruler_lc.py
jopetty: Makes sense, changed to `ruler-plus`

src/olmo_eval/data/ruler_lc_loader.py
jopetty: This is now changed to load data from `allenai/ruler-plus` on HF; I'm not sure if that has an impact on whether or not it should be renamed further, since it's technically a private repo rn but in theory we can make it public easily without exposing anything internal

undfined (changes requested):
Let's try and reduce the I/O and memory footprint a bit. 

- snapshot_download() fetches the entire ~109 GB repository for every task, even though each task needs only one shard. Can we download just the requested task/context file when possible?

- The loader reads the full multi-GB shard before applying limit, so small runs still incur the full memory and I/O cost. Can we apply the limit while reading the data? Have we run the full suite without exhausting our system memory?

jopetty:
[6aad682](https://github.com/allenai/olmo-eval/pull/276/commits/6aad682936a0c966822899f474fdbb45b21fea37) addresses these points:
1. it replaces `snapshot_download()` (which pulled the entire ~109GB `allenai/ruler-plus` repo) with `hf_hub_download()` for just the single shard a task actually needs (`data/{size}/{task}/validation.jsonl`)
2. `load_ruler_plus_shard()` now streams the JSONL line-by-line and reservoir-samples (Algorithm R) up to the configured limit, instead of parsing every record into a list before subsampling. Also skips json.loads entirely on lines that don't get selected into the sample.

In local testing, setting `--limit 5` drops the memory usage from 8GB to ~500MB.

PR #278, opened 2026-08-11
Add multimodal evaluation

jason718:
## Description

Adds multimodal (vision-language) evaluation to olmo-eval: **25 image tasks** across five families, **5 suites**, and **two VLM inference paths**. Everything is aligned to the `mm_olmo` reference implementation and verified numerically against it — see [Results](#results).

The task/harness/provider split is unchanged: image tasks attach images to `LMRequest.images`, and the provider decides how to render them. No changes to the text-only evaluation paths beyond the shared provider files.

### Tasks

| Family | Suite | Tasks |
|---|---|---|
| **Image-QA** | `molmo2_imageqa` | `chart_qa`, `vqa2`, `doc_qa`, `info_qa`, `text_vqa`, `real_world_qa`, `mmmu`, `mmmu_pro`, `math_vista`, `countbench_qa`, `pixmo_count`, `ai2d`, `charxiv_descriptive`, `charxiv_reasoning` |
| **Dense caption** | `molmo2_imageqa_caption` | the above plus `dense_caption` (GPT-4o recall+consistency judge); `dense_caption_captioner` for style_and_length_v2 captioner checkpoints |
| **Pointing** | `molmo2_pointing` | `pixmo_points_eval`, `sa_co_gold_subset` |
| **Pointing (`_mp`)** | `molmo2_pointing_mp` | `pixmo_points_eval_mp`, `sa_co_gold_subset_mp`, `sa_co_gold_point_4k_mp` (+ `sa_co_gold_point_mp`, registered but outside the suite) |
| **Multi-image** | `molmo2_multiimage` | `muir_bench`, `mmiu`, `blink` |

Image-QA primaries are all 0-1 so the suite averages them; `dense_caption` reports 0-100, so `molmo2_imageqa_caption` is display-only. Pointing scores point-in-mask precision/recall/f1 by maximum bipartite matching rather than VQA-style answers, which is why it is a separate family.

`mmmu_pro` is deliberately decoupled from `mmmu` — its own prompts, parser and scorer, faithful to the official mmmu-pro repo. One task covers all three official settings and emits the SOTA-reported `overall = (standard-10 + vision)/2`.

### Prompting follows the checkpoint

mm_olmo's `DataFormatter` builds the prompt from **two settings recorded in each checkpoint's own config**, so the same benchmark asks different things of different models:

- `prompt_templates` — `none` uses the bare lowercased label; `uber_model_v2` samples one of 43 templates, seeded per example.
- `system_prompt` — `demo_or_style_v2` adds nothing; `style_and_length_v2` prefixes the style (`"pointing: "`, `"point_count: "`).

`common/pointing_prompts.py` reproduces both, including mm_olmo's exact per-example RNG (`RandomState((691203*195172 + idx) % (2**32-1))`). The `_mp` task variants route the bare label through it; the pre-built-question benchmarks apply the style prefix via `StylePrefixMixin`. Defaults are unchanged (`demo_or_style_v2` adds no prefix), so only an explicit `-o system_prompt_style=style_and_length*` alters existing runs.

This matters: on a stage-1 checkpoint the mismatch cost `dense_caption` 7.1 recall points and turned 3.4% of caption answers into `<points .../>` markup.

### Inference

**`HuggingFaceProvider`** gains an image-text-to-text path: `AutoProcessor` + `AutoModelForImageTextToText`, transformers-5 load shims for Molmo2 remote code, fp32-weights/bf16-autocast to match `mm_olmo`'s `amp_bf16` numerics, and pickle-safe lazy image decode. It also loads consolidated OLMo-core exports directly.

**`olmo_core_vlm`** is a new provider for OLMo-core `MultimodalLM` checkpoints in three on-disk formats: raw trainer saves, consolidated safetensors exports, and mm_olmo trainer checkpoints (config-translated and key-remapped at load). It drives its own decode loop with batched variable-length KV-cached decoding, since OLMo-core's vision branch has no multimodal generation module. `-m` with `provider.kind=olmo_core` auto-detects multimodal checkpoints.

---

## Running it

```bash
uv sync --extra hf                    # HF multimodal path
uv sync --extra hf --extra olmo_core  # + OLMo-core MultimodalLM checkpoints

export MOLMO_DATA_DIR=/weka/oe-training-default/mm-olmo   # read-only; loaders raise rather than build caches
export OPENAI_API_KEY=...                                 # dense_caption, charxiv, math_vista judges

uv run olmo-eval run -m molmo2-4b -t molmo2_imageqa
uv run olmo-eval suite inspect molmo2_pointing_mp
```

For an OLMo-core checkpoint, force the provider and pass the checkpoint's prompt family:

```bash
uv run olmo-eval run --harness default -o provider.kind=olmo_core \
    -m /weka/path/to/stepN \
    -t pixmo_points_eval_mp -o prompt_templates=none -o system_prompt_style=style_and_length_v2
```

`math_vista:offline` avoids the OpenAI dependency. Full end-user docs are in the README under **Multimodal Evaluation**.

---

## Results

Verified against `mm_olmo` on Molmo2-4B (fp32 + bf16 autocast, `max_crops=24`).

**`_mp` pointing, reproducing mm_olmo runs of the same protocol** — prompts byte-identical (3856/3856 and 1215/1215):

| Benchmark | this PR | mm_olmo | Δ |
|---|---|---|---|
| `sa_co_gold_subset_mp` (4B-Pretrain) | 0.6631 | 0.6658 | −0.0027 |
| `pixmo_points_eval_mp` (4B-Pretrain) | 0.8152 | 0.8083 | +0.0069 |

**Original pointing / image-QA parity:** SACoGold weighted f1 0.6909 vs 0.6922 (all 18 metrics within 0.0025); PixMo-Points f1 0.8314 vs 0.8290. MMMU-Pro full run (5190 ex): overall 26.01 (standard-10 35.20 / standard-4 44.62 / vision 16.82).

**Exactness checks:** prompt construction token-exact with the released Molmo2 processor (765/765 `input_ids`); the OLMo-core → Molmo2 weight conversion bit-exact against an independently converted twin (706/706 tensors); KV-cached decoding token-for-token identical to no-cache decoding.

---

## Bugs this surfaced

Running real OLMo-core multimodal checkpoints end-to-end exposed several latent defects, each fixed here:

| Bug | Symptom |
|---|---|
| `_embed` ignored split-vocab `extra_weight` | CUDA device-side assert on every image token |
| flex-attention checkpoints couldn't carry a KV cache | 30× slowdown (425s → 14s / 4 examples) |
| image tasks never declared `pillow` | all 19 image tasks failed on Beaker |
| tqdm lazy-lock race in parallel task prep | nondeterministic task failure |
| provider-init timeout hardcoded at 900s | every worker timed out under concurrent checkpoint loads |
| pointing/judge scorer deps undeclared | **silent 0.0000 with status `Success`** |
| mm_olmo untied LM head silently tied to embeddings | Molmo2-8B scored 0.12 vs 0.83 on pixmo, never emitting EOS |

The two silent ones are worth review attention: a task whose every instance fails to score is currently indistinguishable from one that legitimately scored zero, and the LM-head substitution loaded and ran without error.

## Type of Change

## Testing

- [x] Unit tests pass locally (`pytest tests/ --ignore=tests/integration/`) — **2054 passed, 9 skipped**
- [x] New tests added — all CPU-only, no GPU or network: multimodal provider paths, image-QA request/prompt parity vs reference dumps, scoring, MMMU-Pro parser, pointing geometry and metrics, mm_olmo prompt-parity guard, and the mm_olmo key remap
- [x] Instance counts verified against mm_olmo for every task added
- [x] Lint + format clean; `ty` clean under CI's `--no-group vllm`

Integration-level verification (real checkpoints, real data) was done out-of-band against `mm_olmo` — see [Results](#results).

## Checklist

- [x] I have added/updated documentation as needed — README **Multimodal Evaluation** section


undfined:
Some high level thoughts before we decompose this a bit.

1) We probably want a cleaner package layout for vision stuff. Maybe roughly something like this:

```
evals/vision/
├── tasks/
│   ├── base.py
│   ├── single_image.py
│   ├── multi_image.py
│   └── pointing.py
├── scoring/
│   ├── vqa.py
│   ├── multiple_choice.py
│   ├── pointing.py
│   └── judges.py
├── benchmarks/
│   ├── charxiv/
│   ├── mmmu/
│   └── ...
└── data/
    ├── images.py
    └── paths.py
```

2) Can we add an abstraction for handling batch scoring instead of overriding behavior for charxiv? I think something roughly like this could be cleaner if it's not too onerous:

```
class ResponseBatchScorer:
    async def score_responses(
        self,
        responses: Sequence[Response],
        context: ScoringContext,
    ) -> None: ...
```

3) Provider classes have a lot of responsibilities. Let's try to decompose them a bit (olmo-core vl in particular) maybe something like:

```
inference/providers/olmo_core_vlm/
├── provider.py
├── checkpoint.py
├── conversion.py
├── preprocessing.py
├── generation.py
└── cache.py
```

4) (From Sol, can be a follow up possibly but might make your life easier) Runtime requirements should be declarative
The branch has several kinds of requirements:
Python packages.
Model/provider capabilities.
Local benchmark data.
External API credentials.
Judge caches.
GPU/runtime requirements.
These are currently spread among optional dependencies, task docstrings, environment-variable reads, provider presets, and README instructions.
Tasks and providers could expose structured requirements that can be validated before launching:
```
RuntimeRequirements(
    packages=(...),
    secrets=(...),
    data_paths=(...),
    capabilities=("images", "multi_image"),
)
```

This would let the CLI fail during planning rather than during dataset loading or scoring. It would also make suite requirements computable from their constituent tasks.

5) Consider consolidating class construction into some sort of base class for the vision task classes:
ImageQATask, MultiImageQATask, PointingTask, and DenseCaptionEval each implement similar mechanics for caching things, applying limits, resolving lazy images, constructing chat requests, and attaching image inputs. 
Maybe something like:

```
class ImageTask(Task):
    def build_instances(self) -> Iterator[Instance]: ...
    def resolve_images(self, instance: Instance) -> tuple[Image, ...]: ...
    def format_messages(self, instance: Instance) -> tuple[Message, ...]: ...
```

6) Consider something like the following so that we centralize how we describe image references:

```
class ImageReference(ABC):
    """A lazily resolved reference to an image."""

    @abstractmethod
    def load(self) -> Image.Image:
        """Load and return the referenced image."""

    @abstractmethod
    def describe(self) -> str:
        """Return a useful source description for logging and errors."""


@dataclass(frozen=True, slots=True)
class PathImageReference(ImageReference):
    path: Path

    def load(self) -> Image.Image:
        return Image.open(self.path)

    def describe(self) -> str:
        return str(self.path)
```

We can then have different concrete classes like InMemoryImageReference or w/e

jason718:
Per the review, this PR is now split into a five-PR stack to `main`, restructured into the proposed `evals/vision/` layout as it goes:

1. #316 — Multimodal inference: image requests, HF VLM path, OLMo-core VLM provider (as a package, per #3)
2. #317 — Vision task framework (`evals/vision/`, `VisionTask` per #5) + captioning
3. #322 — Pointing + counting
4. #323 — Image QA
5. #324 — Multi-image + docs

Each PR is a self-contained milestone, stacked strictly serially, and AST-verified content-identical to the code reviewed here (183/183 top-level symbols; the only changed definitions are the VisionTask consolidation and added `required_secrets`). Two amendments to the proposed layout — flat benchmark modules, and no re-export of vision scorers through `common/scorers/__init__` (package cycle) — are argued in #317 for sign-off.

The remaining review asks (#2 ResponseBatchScorer, #3b provider-class decomposition, #4 RuntimeRequirements, #6 ImageReference laziness) are deliberate follow-ups; issues to be filed once the stack lands. This PR stays open as the tracking umbrella until then.

jason718:
Follow-up issues from the review of this stack, linked here before the last PR merges (as requested on #324):

- #392 — `ResponseBatchScorer`: let a scorer grade several responses per call (CharXiv's official grader batches 5 questions per GPT call; ours makes one call per question)
- #393 — Decompose `OlmoCoreVLMProvider` into collaborators (cross-references allenai/OLMo-core#867)
- #394 — `RuntimeRequirements`: unify task dependencies, secrets, data paths and capabilities
- #395 — `ImageReference`: make image laziness a declared contract
- #379 — Consolidate checkpoints before eval instead of loading DCP directly
- #381 — Vision tasks ship Ai2 Weka paths as source defaults
- #402 — Port mm_olmo's `GroundingEvaluatorV5` and the balanced SA-Co v5 benchmark

Stack status: #316, #317, #322 and #323 are merged; #324 is the last.

PR #280, opened 2026-08-14
Add n-gram copying loss/bpb eval for C4

jopetty:
Adds a bits-per-byte metrics restricted to k-gram sequences which have been repeated earlier in the same document, allowing us to measure how well models can copy previously-seen spans of text.

```
x y x y x y   input sequence
0 0 1 2 3 4   n-gram repetition score
F F T T T T   mask @ k = 1
F F F T T T   mask @ k = 2
F F F F T T   mask @ k = 3
F F F F F T   mask @ k = 4
```

Adds `NGramCopyingBPBMetricInstanceAvg` (average of per-document averages) and `NGramCopyingBPBMetricByteAvg` (average across all tokens in corpus) at `k=1...5` and wires them into `c4/c4_1k/c4_10k/c4_100k` as a new `:ngram_copying_bpb` variant.

### DIFF
--- a/src/olmo_eval/common/metrics/ngram_copying.py
+++ b/src/olmo_eval/common/metrics/ngram_copying.py
@@ -1,0 +1,105 @@
+"""Metrics for n-gram copying: bits per byte restricted to repeated-span positions."""
+
+import math
+from collections.abc import Sequence
+from dataclasses import dataclass, field
+
+from olmo_eval.common.scorers import NGramCopyingBPBScorer
+from olmo_eval.common.types import Response
+
+from .base import Metric
+
+#: Default set of n-gram-length thresholds. Add new values here to extend coverage.
+NGRAM_COPYING_K_VALUES: tuple[int, ...] = (1, 2, 3, 4, 5)
+
+
+def _ngram_copying_masked_totals(
+    response: Response, scorer: NGramCopyingBPBScorer
+) -> tuple[float, int] | None:
+    """Look up the masked (logprob, bytes) totals for a response's document output."""
+    if not response.outputs:
+        return None
+    return scorer.masked_totals(response.outputs[0])
+
+
+@dataclass(frozen=True, slots=True)
+class NGramCopyingBPBMetricInstanceAvg(Metric):
+    """Arithmetic mean of per-document BPB restricted to length-k+ repeated n-grams.
+
+    Documents with no positions meeting the threshold are excluded from the average.
+    """
+
+    name: str = field(init=False)
+    k: int = 1
+    scorer: NGramCopyingBPBScorer = field(init=False)
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "name", f"ngram_copying_bpb_instance_avg_k{self.k}")
+        object.__setattr__(self, "scorer", NGramCopyingBPBScorer(k=self.k))
+
+    def compute(self, responses: Sequence[Response]) -> float:
+        values = [value for r in responses if (value := self.compute_instance(r)) is not None]
+        return sum(values) / len(values) if values else 0.0
+
+    def compute_instance(self, response: Response) -> float | None:
+        totals = _ngram_copying_masked_totals(response, self.scorer)
+        if totals is None:
+            return None
+        total_logprob, total_bytes = totals
+        if total_bytes == 0:
+            return None
+        return -total_logprob / (total_bytes * math.log(2))
+
+    def supports_pairwise_scorer_fallback(self) -> bool:
+        return False
+
+    def pairwise_higher_is_better(self) -> bool:
+        return False
+
+
+@dataclass(frozen=True, slots=True)
+class NGramCopyingBPBMetricByteAvg(Metric):
+    """Byte-weighted (corpus-level) BPB restricted to length-k+ repeated n-grams.
+
+    Documents with no positions meeting the threshold are excluded from the sum.
+    """
+
+    name: str = field(init=False)
+    k: int = 1
+    scorer: NGramCopyingBPBScorer = field(init=False)
+
+    def __post_init__(self) -> None:
+        object.__setattr__(self, "name", f"ngram_copying_bpb_byte_avg_k{self.k}")
+        object.__setattr__(self, "scorer", NGramCopyingBPBScorer(k=self.k))
+
+    def compute(self, responses: Sequence[Response]) -> float:
+        total_logprob = 0.0
+        total_bytes = 0
+        for response in responses:
+            totals = _ngram_copying_masked_totals(response, self.scorer)
+            if totals is None:
+                continue
+            logprob, num_bytes = totals
+            if num_bytes == 0:
+                continue
+            total_logprob += logprob
+            total_bytes += num_bytes
+
+        if total_bytes == 0:
+            return 0.0
+        return -total_logprob / (total_bytes * math.log(2))
+
+    def compute_instance(self, response: Response) -> float | None:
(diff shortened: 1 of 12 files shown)


src/olmo_eval/common/metrics/ngram_copying.py
```diff
@@ -0,0 +1,105 @@
...
+    if not response.outputs:
+        return None
+    return scorer.masked_totals(response.outputs[0])
+
+
+@dataclass(frozen=True, slots=True)
+class NGramCopyingBPBMetricInstanceAvg(Metric):
+    """Arithmetic mean of per-document BPB restricted to length-k+ repeated n-grams.
+
+    Documents with no positions meeting the threshold are excluded from the average.
```
viking-sudo-rm: Could we just do one aggregate average weighted by the overall number of "copy bytes", i.e., bytes contributed by copy tokens per doc? So if there are three documents with 5, 3, and 0 copy bytes respectively, the denominator in the average would be 8.

viking-sudo-rm: I think this makes more sense because it would avoid overweighting tokens fro documents with a small number of copy tokens. It would also naturally handle the edge case where the document has no copy tokens.

src/olmo_eval/common/metrics/ngram_copying.py
```diff
@@ -0,0 +1,105 @@
...
+        return sum(values) / len(values) if values else 0.0
+
+    def compute_instance(self, response: Response) -> float | None:
+        totals = _ngram_copying_masked_totals(response, self.scorer)
+        if totals is None:
+            return None
+        total_logprob, total_bytes = totals
+        if total_bytes == 0:
+            return None
+        return -total_logprob / (total_bytes * math.log(2))
```
viking-sudo-rm: With the changes I proposed above, `total_bytes` would be the number of copying bytes. So basically the number of copy tokens times some tokenizer-specific constant

src/olmo_eval/common/metrics/ngram_copying.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def supports_pairwise_scorer_fallback(self) -> bool:
+        return False
+
+    def pairwise_higher_is_better(self) -> bool:
+        return False
+
+
+@dataclass(frozen=True, slots=True)
+class NGramCopyingBPBMetricByteAvg(Metric):
```
viking-sudo-rm, line 26: This is probably just my lack of familiarity with the new olmo-eval framework, but I don't understand why we have this class and also the other one (feel free to resolve if this is just my ignorance)

viking-sudo-rm (changes requested):
The core per-document logic in src/olmo_eval/common/scorers/ngram_copying.py looks solid.

When aggregating across document, it appears the current approach is to weight all documents evenly. It would make more sense to me to sum log probs and bytes independently across documents and then divide. This will naturally handle the case where some document contributes 0 bytes.

jopetty:
@viking-sudo-rm The `...ByteAvg` version does what you (I think) are requesting, i.e., averaging over all relevant tokens irrespective of document boundary while the `...InstanceAvg` version first averaged within-documents; I removed the `...InstanceAvg` version here, which I think resolves your issues?

src/olmo_eval/common/metrics/ngram_copying.py
jopetty: The `...ByteAvg` metric does this, but I agree that this version is maybe not a useful signal

src/olmo_eval/common/metrics/ngram_copying.py
jopetty: See comment above

src/olmo_eval/common/scorers/ngram_copying.py
```diff
@@ -0,0 +1,81 @@
...
+        total_logprob = 0.0
+        total_bytes = 0
+        for entry, matched in zip(output.logprobs, mask, strict=True):
+            if not matched:
+                continue
+            total_logprob += entry.get("logprob", 0.0)
+            token_bytes = entry.get("bytes")
+            total_bytes += (
+                len(token_bytes) if token_bytes is not None else len(entry["token"].encode("utf-8"))
+            )
```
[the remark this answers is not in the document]
jopetty: This implementation follows the same as in the normal `BitsPerByteScorer`; see eg https://github.com/allenai/olmo-eval/blob/1a737f9534e765dca6251a278eba0e222be9d8e6/src/olmo_eval/common/scorers/base.py#L225

src/olmo_eval/common/metrics/ngram_copying.py
```diff
@@ -0,0 +1,105 @@
...
+        object.__setattr__(self, "scorer", NGramCopyingBPBScorer(k=self.k))
+
+    def compute(self, responses: Sequence[Response]) -> float:
+        values = [value for r in responses if (value := self.compute_instance(r)) is not None]
+        return sum(values) / len(values) if values else 0.0
+
+    def compute_instance(self, response: Response) -> float | None:
+        totals = _ngram_copying_masked_totals(response, self.scorer)
+        if totals is None:
+            return None
```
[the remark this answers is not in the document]
jopetty: Renaming the object to `_scorer` doesn't fix this, and in any case the base class of `Scorer` already handles this by overriding `__call__(self)` to permit exactly this behavior. No changes are necessary. https://github.com/allenai/olmo-eval/blob/1a737f9534e765dca6251a278eba0e222be9d8e6/src/olmo_eval/common/scorers/base.py#L29-L35

src/olmo_eval/common/metrics/ngram_copying.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+def _ngram_copying_masked_totals(
+    response: Response, scorer: NGramCopyingBPBScorer
+) -> tuple[float, int] | None:
+    """Look up the masked (logprob, bytes) totals for a response's document output."""
+    if not response.outputs:
+        return None
+    return scorer.masked_totals(response.outputs[0])
+
```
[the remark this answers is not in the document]
jopetty: > and doesn’t check output.logprobs

Not true, it does check within the method body of `scorer.masked_totals(...)`: https://github.com/allenai/olmo-eval/pull/280/changes/BASE..5f695133feabc6b568fb0725f57a705f1e95e17b#diff-234cea571b88e32ede62ea36527012defbd644b2b61a5b3ca7881b78d66b01f2R54-R55

> which can select the wrong continuation for multi-output responses (gold_idx) 

This metric is only ever attached to whole-document loglikelihood tasks, not sampling evals, so this concern doesn't make sense.

viking-sudo-rm:
LGTM! I'm not familiar with olmo-eval so I will hold off on merging in case @undfined has any opinions

undfined (approved):
Looks solid! One thing to consider before merging:

We might want to use token IDs rather than decoded token strings here. I had claude scan the C4 10k subset and found that Unicode fragments that decode to � (our tokenizer) produce 717 false k=5 matches (~0.94%) and inflate the byte denominator by ~0.87%. Not a ton of false pos cases but worth considering whether we care about that noise. Preserving token IDs and raw token bytes in LogProbEntry could avoid treating distinct tokens as repeats and keep BPB normalization more accurate.

jopetty:
@undfined makes sense! [24ad336](https://github.com/allenai/olmo-eval/pull/280/commits/24ad33609b6aebfcf4da0e5541f01bb510d6f98c) addresses this

PR #295, opened 2026-08-17
Adding CTC Suite (upcoming release) to olmo eval

PrasannS:
## Description

Added 22 tasks from the CTC-Suite (coming out publically soon) in case OLMO long context folks (Yashas, Jackson, ...) would be interested in using it.

## Type of Change

New tasks in suite.

## Testing

No extensive testing

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,6 +158,9 @@ vllm = [
 [tool.ruff]
 line-length = 100
 target-version = "py312"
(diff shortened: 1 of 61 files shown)


YashasSamaga:
What is the difference between these two PRs:
1. https://github.com/allenai/olmo-eval/pull/289
2. https://github.com/allenai/olmo-eval/pull/295

PR #296, opened 2026-08-17
Fix IFEval loose scoring: build each verifier once, not per variant

abhishekraok:
## Summary

`IFEvalScorer` rebuilt the instruction verifier for the strict check and again for each of the 8 loose variants. Several `ifbench` verifiers (e.g. `count:word_count_range`) draw **random** parameters for kwargs the dataset leaves unset, so the nine checks could each run against different criteria. Measured on 150 identical responses:

- loose accuracy below strict (0.375 vs 0.400 prompt-level) — impossible, since loose's variant set includes the raw response
- score drift across reruns of byte-identical output (spread 0.0084 on `inst_level_strict_acc`)

Fix: build each verifier once (`_build_instruction`) and reuse it for the strict check and all loose variants.

## Notes for review

- oe-eval reuses one instance within its loose test but builds a *separate* one for strict, so the reference can itself produce loose < strict on random-kwarg prompts. This PR shares one instance across both — slightly stronger than oe-eval; expected values unchanged.
- Reuse is safe: of all 83 verifier classes, only `EndChecker` mutates state in `check_following`, and idempotently.
- **Numbers shift**: affects `ifeval_ood`, live in the `ifbench` suite. Zero effect on `allenai/IFBench_test` (kwargs fully specified); ~0.8pt on `google/IFEval`-style prompts.

## Test plan

- New regression test: `count:word_count_range` with unset kwargs, 200 trials, asserts strict ⇒ loose. Fails on pre-fix code, passes on the fix.
- ruff, ty, and the full unit suite pass (1860 passed, 7 skipped).

### DIFF
--- a/src/olmo_eval/common/scorers/ifeval.py
+++ b/src/olmo_eval/common/scorers/ifeval.py
@@ -36,13 +36,18 @@ def _loose_response_variants(response: str) -> list[str]:
     ]
 
 
-def _check_one(
+def _build_instruction(
     instruction_cls: Any,
     instruction_id: str,
     kwargs: dict[str, Any],
     prompt: str,
-    response: str,
-) -> bool:
+) -> Any:
+    """Instantiate and configure a single instruction verifier.
+
+    Built once per instruction and reused for the strict check and every loose
+    variant: verifiers draw random parameters for unspecified kwargs, so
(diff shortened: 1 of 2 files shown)


undfined (approved):
lgtm!

PR #297, opened 2026-08-17
Add single-turn IFEval task

abhishekraok:
## Summary

Adds `google/IFEval` (541 prompts) as an `ifeval` task. It is in the post-training `OLMO_3` suite in open-instruct but had no counterpart here — only the OOD slice (`ifeval_ood`) existed. Mirrors `ifeval::hamish_zs_reasoning_deepseek`, the Olmo 3 post-training regime: chat format, temperature 0.6 / top-p 0.95 with generation bounded by the model context (`max_tokens=None`), `<think>` reasoning stripped before verification (as oe-eval's model-level `r1_style` does), primary metric `prompt_level_loose_acc`. The vendored `ifbench` registry covers all 25 instruction IDs the dataset uses (verified: 25/25, none missing), so the existing `IFEvalScorer` and its four metrics apply unchanged.

## Parity

Against the paper's own oe-eval generations: Olmo-3-7B-Think 539/541 and Olmo-3-7B-Instruct-SFT 538/541 on all four metrics, prompts byte-identical (see the parity comment below for the 5 explained disagreements). Earlier greedy `ifeval::tulu` checks, which established prompt/dataset/scorer equivalence, follow.

## Parity with oe-eval `ifeval::tulu` (original, greedy)

Verified against `oe-eval-internal` directly, not by inspection. Checks run on the merged state of this PR and #314.

**Config** — matches on every field:

| field | oe-eval `ifeval::tulu` | this task |
|---|---|---|
| format | `use_chat_format: True` | `RequestType.CHAT` |
| max tokens | `max_gen_toks: 2048` | `max_tokens=2048` |
| sampling | `do_sample: False, temperature: 0.0` | same |
| stops | `stop_sequences: []` | none |
| split | `train` | `train` |
| primary metric | `prompt_level_loose_acc` | `prompt_level_loose_acc` |

**Dataset** — oe-eval reads `HuggingFaceH4/ifeval`, this reads `google/IFEval`. They are content-identical: 541 rows, same 24 kwargs schema keys, `key`/`prompt`/`instruction_id_list` equal on every row. Raw `kwargs` differ on all 541 rows through HF null-padding alone, and **0/541 differ after the None-strip this task applies**.

**Prompts and scorer verdicts** — drove `oe_eval.dependencies.ifeval.utils.process_results` and this PR's scorer over the same inputs, all 541 prompts × 6 response shapes (compliant, empty, markdown-wrapped, refusal, prompt echo, long filler):

```
(prompt, response) pairs compared : 3234
instruction verdicts compared     : 4980
prompt mismatches                 : 0/541
strict verdict mismatches         : 0/3234
loose  verdict mismatches         : 0/3234
```

Stable across three runs. Details and the two excluded rows are in [this comment](https://github.com/allenai/olmo-eval/pull/297#issuecomment-5415061638).

## Known caveat, not specific to this port

Two dataset rows (keys 1122, 1129) specify a non-alphabetic `letter` for `keywords:letter_frequency`. Every IFEval verifier — `ifbench` and oe-eval's alike — rejects those and substitutes `random.choice(string.ascii_letters)`, so those rows score a random letter and flip between runs. Bounded at 2/541 prompts (~0.37 points of prompt-level noise) and present in any IFEval harness. Filed with the other upstream defects in #320; excluded from the comparison above.

## Scope

No reasoning-mode variant here. oe-eval runs this under `ifeval::hamish_zs_reasoning_deepseek` (temp 0.6 / top-p 0.95, uncapped budget). That regime should land once as a shared variant across all post-training tasks rather than as a one-off here, so this PR deliberately doesn't set the convention.

## Not covered

An end-to-end score cell against a live model. With prompts and scorer verdicts exact, a GPU run would be confirming the generation path rather than the port.

## Test plan
- [x] 9 task tests added (`tests/evals/tasks/test_ifeval.py`): config, null-padded kwargs strip, single-user-message formatting, verbatim extraction, scoring end to end
- [x] `ruff check`, `ruff format --check`, `ty check` clean
- [x] Full suite: 1966 passed, 7 skipped
- [x] Config, dataset, prompt and scorer parity with oe-eval (above)
- [ ] Optional: end-to-end score cell vs `ifeval::tulu` on a fixed checkpoint


abhishekraok:
## CPU parity against oe-eval's own code (no GPU required)

Verified this port directly against `oe-eval-internal` rather than by inspection. All checks run on the merged state of this PR and #314.

### Config matches `ifeval::tulu`

`primary_metric: prompt_level_loose_acc`, `split: train`, `use_chat_format: True`, `max_gen_toks: 2048`, `do_sample: False`, `temperature: 0.0`, `aggregation_levels: [prompt, inst]`, `strictness_levels: [strict, loose]` — each matches the entry in `oe_eval/configs/tasks.py`.

### Dataset

oe-eval reads `HuggingFaceH4/ifeval`; this task reads `google/IFEval`. **They are content-identical**: 541 rows each, same 24 kwargs schema keys, and `key` / `prompt` / `instruction_id_list` match on every row. Raw `kwargs` differ on all 541 rows purely through HF null-padding, and **0/541 differ after the None-strip this task already applies**. The dataset choice is not a parity risk.

### Prompt construction

oe-eval builds `query = doc["prompt"]` through a no-op template, then wraps it as a single user message. Compared against this task's `format_request` output on all 541 docs: **0 mismatches**.

### Scorer verdicts vs oe-eval's own `process_results`

Drove `oe_eval.dependencies.ifeval.utils.process_results` and this PR's `IFEvalScorer` over the same inputs — all 541 real prompts × 6 response shapes (compliant, empty, markdown-wrapped, refusal, prompt echo, long filler):

```
(prompt, response) pairs compared : 3234
instruction verdicts compared     : 4980
prompt mismatches                 : 0/541
strict verdict mismatches         : 0/3234
loose  verdict mismatches         : 0/3234
```

Stable across three runs.

### Two dataset rows are inherently nondeterministic — in both harnesses

Excluded from the above, and worth knowing for *any* IFEval number regardless of harness:

| key | `letter` | effect |
|---|---|---|
| 1122 | `'#'` | non-alphabetic |
| 1129 | `'!'` | non-alphabetic |

`LetterFrequencyChecker.build_description` rejects any `letter` outside `a`–`z` and substitutes `random.choice(string.ascii_letters)`, so it scores a *randomly chosen* letter's frequency instead of the one the prompt asks for. Both oe-eval's vendored verifier and `ifbench` do this. Verified by repeated builds: oe-eval drew `x, c, t, i, c, t` on consecutive calls to the same row, flipping the verdict between runs.

Effect is bounded — 2/541 prompts, so at most ~0.37 points of prompt-level noise — but it is not a port difference, and it is why the comparison above excludes those rows. Worth fixing upstream in the dataset alongside the `VGraf/ifeval_mt` row noted on #314.

### What this does not cover

An end-to-end score cell against a real model. Prompts and scorer verdicts are now exact, so a GPU run would be confirming the generation path rather than the port.

abhishekraok:
**GPU parity evidence** — allenai/Olmo-3-7B-Instruct, run from `run/ifeval-parity` = this PR's head `d591a43` (code unchanged since). Beaker: olmo-eval `01M0X8S1SKTMBZV28P9H4GZAEA` (results `01M0X9ZRP2B6EC355EX0Z87GYC`), oe-eval `ifeval::tulu` `01M0X8SBHBMM94M935Y9NZBE15` (results `01M0X9VEEWF84NSZGK9930JJSW`).

**ifeval (google/IFEval, n=541, greedy):**

| metric | oe-eval | olmo-eval (this PR) |
|---|---|---|
| prompt_level_loose (primary) | 0.852126 | **0.852126** (equal) |
| prompt_level_strict | 0.828096 | 0.829945 |
| inst_level_strict | 0.884892 | 0.882494 |
| inst_level_loose | 0.901679 | 0.899281 |

**Paired scorer check** (oe-eval's stored generations through this PR's `IFEvalScorer`): agreement **539/541 (99.63%)** on all four metrics. The 2 disagreements are docs 19 and 22 — the two known-defective `google/IFEval` rows (#320) whose `letter` kwarg is non-alphabetic (`#`, `!`). Root cause of the divergence, verified in source: **oe-eval's vendored verifier silently replaces a non-alphabetic letter with `random.choice(string.ascii_letters)`** (`dependencies/ifeval/instructions.py` ~L1414), so its verdict on those rows is random and unrelated to the prompt (doc 22's prompt explicitly asks for six `!`); the `ifbench` package counts the literal character, which is what the prompt asks. So 539/541 is the *maximum achievable* agreement — the two remaining rows are oe-eval coin flips, and this PR's behavior on them is the faithful one. The identical run-level primary (461/541 both) is those two flips cancelling.

**Also first-ever execution of the multi-turn tasks** (no oe-eval counterpart run; validation, not parity): `ifeval_mt_wildchat_unused_withRewrite` prompt strict/loose 0.8067/0.8191 (n=1774, incl. the #325 row handled cleanly), `ifeval_mt_ood_wildchat_unused_withRewrite` 0.4946/0.5494 (n=1387).

**Not covered:** `ifeval_ood` — blocked on olmo-eval by the `IFBench_test2` token policy; oe-eval reference banked (prompt loose 0.310 / strict 0.273, n=300, results `01M0X9N1R1VDM1YEH9J16NTFGX`) and pairs against olmo-eval once the `IFBench_test` dataset switch lands.

src/olmo_eval/evals/tasks/ifeval.py
```diff
@@ -0,0 +1,86 @@
...
+    split = Split.TRAIN
+    metrics = (
+        IFEvalPromptStrictAccuracy(),
+        IFEvalPromptLooseAccuracy(),
+        IFEvalInstStrictAccuracy(),
+        IFEvalInstLooseAccuracy(),
+    )
+    primary_metric = _PRIMARY_METRIC
+    sampling_params = SamplingParams(
+        max_tokens=2048,
```
mgmorgan23, line 49: IFEval was functionally run with overrides for max_tokens (32768), temperature (0.6), and top_p (0.95) to match the other post-training tasks

mgmorgan23:
Overall this looks good to me – I think like MMLU it might be best to use pytest for the tests to stay consistent. I also would suggest running a couple more parity tests on additional models for full verification!

src/olmo_eval/evals/tasks/ifeval.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Confirmed from the PAPER runs' stored config: `ifeval::hamish_zs_reasoning_deepseek` — T=0.6, top-p 0.95, `max_gen_toks` = model `max_length` = 32768 (i.e. context-bounded), plus model-level `process_output: r1_style` for both Instruct and Think models. Switched the task to that regime in 583f997 (`max_tokens=None` is the context-bounded form, same as the other post-training ports) and added `<think>` stripping before verification.

abhishekraok:

Tests are pytest now (583f997).

**Parity against the paper's own oe-eval generations** (prompts byte-identical to the logged requests; r1-style stripping reproduces oe-eval's scored text 541/541):

| run | all-4-metric agreement | prompt_level_loose |
|---|---|---|
| Olmo-3-7B-Think PAPER run1 (`01KAD4RGRYERJZY4SHEPXYPBX4`) | 539/541 | oe 0.7930 · ours 0.7948 |
| Olmo-3-7B-Instruct-SFT PAPER run1 (`01KACQQ78HTSZ9JB0NSMMESTD7`) | 538/541 | 0.8096 both |

The 5 disagreements are environment, not port logic: keys 1122/1129 are the random-letter rows (#320); key 279 is unseeded `langdetect` in `english_lowercase` (measured 29/40 "en" vs 11/40 "af" on that text); key 3407 is NLTK — oe-eval pins 3.8.1, this repo has 3.9.x, and 3.9 splits `unfolds—SHINING` on the em-dash where 3.8.1 doesn't, changing a `capital_word_frequency` count (reproduced under 3.8.1).

Without the `<think>` stripping, the Think run agrees on only 289/541 — `ifeval_ood`/`ifeval_mt` still score raw text, tracked in #348.

mgmorgan23 (approved):
LGTM!

PR #302, opened 2026-08-18
Make the counting and pointing benchmarks follow the checkpoint's prompt family too

jason718:
Follow-up to #301, which fixed `dense_caption`. The same gap exists on every other benchmark that
supplies its own question.

mm_olmo's `DataFormatter` keys the `"<style>:"` prefix off the example's **style**, not off whether
the benchmark supplied a ready-made question (`olmo/data/data_formatter.py`: styles `pointing`,
`point_count`, … under `style_and_length`/`_v2`; nothing under `demo_or_style_v*`). Only the `_mp`
variants honoured that, because they route through `build_pointing_prompt` — so
`-o system_prompt_style=style_and_length_v2`, the override a pretrain checkpoint is meant to pass,
was **silently inert** on four tasks:

| task | mm_olmo style | evidence |
|---|---|---|
| `pixmo_count` | `point_count` | PixMoCount message style |
| `countbench_qa` | `point_count` | `CountBenchQaConfig.format_example` sets it explicitly |
| `pixmo_points_eval` | `pointing` | `pixmo_datasets.py:789` |
| `sa_co_gold_subset` | `pointing` | `academic_datasets.py:1766` |

A `style_and_length_v2` checkpoint was trained on `"point_count: how many cats?"` and
`"pointing: the dog"`; these tasks were sending the bare question.

## Why it's worth fixing

On a 4B Molmo2 stage-1 checkpoint, the equivalent mismatch on `dense_caption` (#301) cost **7.1
recall points** (38.18 → 45.28) and turned **3.4% of caption answers into `<points .../>` markup**,
with some replies as short as three words. It presented as a model regression: two stage-1
checkpoints that looked 3–4 points apart were within 0.05 recall of each other once both were scored
correctly. The counting and pointing benchmarks have the same exposure for any pretrain checkpoint.

## The change

- `system_prefix(system_prompt, style=POINTING_STYLE)` — style-aware; the default keeps existing
  callers unchanged.
- `apply_style_prefix(question, system_prompt, style)` — prefixes a pre-built question.
- `StylePrefixMixin` — gives a task the same `system_prompt_style` override the `_mp` tasks already
  had; each task declares its mm_olmo `style`.

**Backwards compatible.** Defaults stay `demo_or_style_v2`, which adds no prefix, so every existing
run keeps its current prompts. Only an explicit `-o system_prompt_style=style_and_length*` changes
behaviour.

## Verification

Prefix resolution checked per task for the default and for `style_and_length_v2` (`point_count: Q`,
`pointing: Q`, bare by default), and #301's `dense_caption` behaviour re-checked on top of this.
`make lint` clean (520 files). `make type-check` reports one diagnostic, pre-existing and in
`providers/olmo_core.py:219` — the one new diagnostic this introduced (the mixin referencing
`self.config`) is resolved with a `TYPE_CHECKING` annotation.

### DIFF
--- a/src/olmo_eval/common/pointing_prompts.py
+++ b/src/olmo_eval/common/pointing_prompts.py
@@ -84,6 +84,8 @@ _STYLE_PREFIX_STYLES = ("style_and_length", "style_and_length_v2")
 
 #: The formatter's style name for these tasks, used as the system prefix.
 POINTING_STYLE = "pointing"
+#: The formatter's style name for the counting benchmarks (PixMo-Count, CountBenchQA).
+POINT_COUNT_STYLE = "point_count"
 
 
 def pointing_rng(index: int) -> np.random.RandomState:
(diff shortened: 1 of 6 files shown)


jason718:
Measured the effect after merging, on the same 4B Molmo2 stage-1 checkpoint. Correcting an
overstatement in the PR description: I wrote that this would be "worth several points on the
counting and pointing numbers", extrapolating from the caption result. It isn't — for counting the
gain is real but an order of magnitude smaller.

| benchmark | bare question | with the family prefix | Δ |
|---|---|---|---|
| `pixmo_count` (correct) | 0.8741 | **0.8815** | +0.74 pts |
| `countbench_qa` (correct) | 0.9592 | **0.9612** | +0.20 pts |
| `dense_caption` (recall), for comparison | 38.18 | 45.28 | +7.09 pts |

Verified the prefix reaches the model: 100% of 400 sampled `pixmo_count` prompts now begin with
`point_count:` (e.g. `point_count: How many people do you see?`), so these are real deltas rather
than a no-op.

Plausible reason for the gap: a counting question states the task unambiguously even without the
style prefix, whereas a bare `"What can be seen in this image?"` left the model choosing between
captioning and pointing — 3.4% of caption answers came back as `<points .../>` markup. The
correctness and consistency argument for this change stands; the magnitude claim for counting did
not.

`sa_co_gold_point_4k_mp` numbers to follow.

PR #303, opened 2026-08-19
Fix ladder dataset resolution and vLLM runtime installs

AkshitaB:
## Summary

- use fully qualified Hugging Face IDs for GSM8K, CommonsenseQA, HumanEval, PIQA, SocialIQA, and WinoGrande
- use PyPI as the primary provider-package index and the runtime PyTorch CUDA index as an extra index
- allow vLLM to resolve its required CUTLASS DSL, torchvision, and torchaudio versions instead of constraining those to the base image

These changes make fresh, uncached ladder eval launches reliable and support the CUDA 13 / vLLM 0.26 runtime stack.

## Validation

- ============================= test session starts ==============================
platform darwin -- Python 3.14.6, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/akshitab/local/code/scaling-ladders/olmo-eval
configfile: pyproject.toml
plugins: anyio-4.9.0, cov-7.1.0
collected 102 items

tests/launch/test_beaker.py ............................................ [ 43%]
..........................................................               [100%]

============================= 102 passed in 1.36s ============================== (102 passed)
- Ruff on all changed Python files
-

### DIFF
--- a/src/olmo_eval/launch/beaker/launcher.py
+++ b/src/olmo_eval/launch/beaker/launcher.py
@@ -894,7 +894,11 @@ class BeakerLauncher:
         # an otherwise compatible vLLM impossible to resolve.
         constraints = "/tmp/cuda-constraints.txt"
         steps.append(
-            f"uv pip freeze -q | grep -E '^(torch|nvidia-)' "
+            # Match torch itself, but not companion packages such as
+            # torchvision or torchaudio. Provider releases pin those companions
+            # alongside their Torch generation and must be allowed to replace
+            # the versions inherited from the base image.
+            f"uv pip freeze -q | grep -E '^(torch(==| @ )|nvidia-)' "
             f"| grep -vE '^nvidia-cutlass-dsl' > {constraints}"
         )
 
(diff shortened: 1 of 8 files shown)


src/olmo_eval/launch/beaker/launcher.py
```diff
@@ -872,7 +872,9 @@ def provider_package_spec(package: str) -> str:
 
             _, flags = parse_install_spec(package)
             extra_flags: list[str] = []
-            if "--index-url" not in flags and "--extra-index-url" not in flags:
+            if "--index-url" not in flags:
+                extra_flags.extend(["--index-url", "https://pypi.org/simple"])
```
jopetty: Minor:

Do we ever set `UV_DEFAULT_INDEX` or `UV_INDEX_URL` / is there a worry that a job might get launched with those set? If so, this might cause a silent deviation from expected behavior.

jopetty:
@AkshitaB are the vLLM CUTLASS changes intended for B300 compatibility? If so, does a job launch successfully when this is required?

AkshitaB:
> @AkshitaB are the vLLM CUTLASS changes intended for B300 compatibility? If so, does a job launch successfully when this is required?

@jopetty Yes, I minimally tested with the following:
 - CUDA 12.8, Torch 2.10, vLLM 0.19.1.
  - CUDA 13, Torch 2.11, vLLM 0.26.

abhishekraok:
Note I merged in #312 which is a subset of this change as I needed it for a few evals.

PR #308, opened 2026-08-22
Add OMEGA-500 task

abhishekraok:
Ports OMEGA-500 from oe-eval as 0-shot chat with the boxed-answer template, temp 0.6 / top-p 0.95 / 131k tokens (mirrors `omega_500:0-shot-chat_deepseek`, the OLMO_3 suite entry). Adds `common/answer_extraction.py`, a shared port of oe-eval's regex-cascade `extract_answer`, for reuse by other reasoning-regime tasks (GPQA CoT already builds on it in #309). Metrics: `exact_match` (format-demoted) and `exact_match_flex` (primary).

- Prompts byte-identical on all 500 docs; golds identical
- Scorer verdicts identical on 544 synthetic responses run through oe-eval's own ExactMatch metric
- 10 unit tests; ruff/ty/pytest clean

**Before undraft:** GPU parity run vs oe-eval on a fixed checkpoint.

### DIFF
--- a/src/olmo_eval/evals/extract/answer_format.py
+++ b/src/olmo_eval/evals/extract/answer_format.py
@@ -21,6 +21,10 @@ class ExtractedAnswer(NamedTuple):
     format_correct: float
 
 
+# The full parameter set mirrors the reference implementation's
+# ``extract_answer`` contract. ``answer_regexes_templates`` is used by the
+# GPQA and MMLU chain-of-thought tasks; the ``use_last_*`` flags have no
+# caller yet but keep the cascade's selection semantics explicit.
 def extract_answer_with_format(
     continuation: str,
     *,
--- a/src/olmo_eval/evals/tasks/omega_500.py
+++ b/src/olmo_eval/evals/tasks/omega_500.py
@@ -15,18 +15,18 @@ from collections.abc import Iterator
 from dataclasses import dataclass
 from typing import Any
 
-from olmo_eval.common.answer_extraction import ExtractedAnswer, extract_answer_with_format
+from olmo_eval.common.formatters import ChatFormatter
 from olmo_eval.common.metrics import AccuracyMetric
 from olmo_eval.common.scorers.base import Scorer
 from olmo_eval.common.types import (
     Instance,
     LMOutput,
     LMRequest,
-    RequestType,
     SamplingParams,
     Split,
 )
 from olmo_eval.data import DataSource
+from olmo_eval.evals.extract import ExtractedAnswer, extract_answer_with_format
 from olmo_eval.evals.tasks.common import Task, register
 
 _BOXED_SUFFIX = "\n\nPresent the answer in LaTex format: \\boxed{Your answer}"
@@ -36,6 +36,13 @@ _TEMPLATE_SUFFIX = (
     '\\boxed{answer}." where answer is just the final solution that solves the '
     'problem. E.g., if the answer is 6, conclude with "Therefore, the final '
     'answer is \\boxed{6}."'
+)
+
+# The suffix lives in the formatter (braces escaped for str.format) so
+# register_variant can swap the prompt template; see the rlzero/midtrain
+# regimes upstream, which use different templates and stop sequences.
+_FORMATTER = ChatFormatter(
+    user_template="{question}" + _TEMPLATE_SUFFIX.replace("{", "{{").replace("}", "}}")
 )
 
 _ANSWER_FORMAT_REGEX = r"Therefore, the final answer is \\boxed\{(.*)\}"
@@ -87,11 +94,21 @@ class OmegaExactMatchScorer(Scorer):
     the requested format; the flex form accepts it regardless.
     """
 
-    name: str = "exact_match"
+    name: str = ""
     flex: bool = False
+
+    def __post_init__(self) -> None:
+        # The name keys scorer storage and deduplication, so it is derived
+        # from ``flex`` unless set explicitly — a flex scorer left with the
+        # strict default name would be silently dropped as a duplicate.
+        if not self.name:
+            object.__setattr__(self, "name", "exact_match_flex" if self.flex else "exact_match")
 
     def score(self, instance: Instance, output: LMOutput) -> float:
         answer, format_correct = _extract(output.text or "")
+        # Whether the answer was stated in the requested format, independent
+        # of correctness — distinguishes "wrong" from "badly formatted".
+        output.metadata["answer_format_correct"] = format_correct
         if not self.flex and format_correct < _FORMAT_CORRECT_CUTOFF:
             answer = ""
         gold = str(instance.gold_answer or "")
@@ -99,13 +116,14 @@ class OmegaExactMatchScorer(Scorer):
 
 
 _STRICT = OmegaExactMatchScorer()
-_FLEX = OmegaExactMatchScorer(name="exact_match_flex", flex=True)
+_FLEX = OmegaExactMatchScorer(flex=True)
 
 
 @register("omega_500")
 class Omega500(Task):
-    data_source = DataSource(path="saumyamalik/omega-500", split="train")
+    data_source = DataSource(path="saumyamalik/omega-500")
     split = Split.TRAIN
+    formatter = _FORMATTER
     metrics = (
         AccuracyMetric(name="exact_match", scorer=_STRICT),
         AccuracyMetric(name="exact_match_flex", scorer=_FLEX),
@@ -120,17 +138,13 @@ class Omega500(Task):
     )
 
     @property
-    def request_type(self) -> RequestType:
-        return RequestType.CHAT
-
-    @property
     def instances(self) -> Iterator[Instance]:
         yield from self._load_instances_cached()
 
     def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance | None:
         question = str(doc["messages"][0]["content"]).replace(_BOXED_SUFFIX, "")
         return Instance(
-            question=question + _TEMPLATE_SUFFIX,
+            question=question,
             gold_answer=str(doc["ground_truth"]),
             metadata={
                 "id": doc.get("index", index),
@@ -139,10 +153,8 @@ class Omega500(Task):
         )
 
     def format_request(self, instance: Instance) -> LMRequest:
-        return LMRequest(
-            request_type=RequestType.CHAT,
-            messages=({"role": "user", "content": instance.question},),
-        )
+        assert self.config.formatter is not None
+        return self.config.formatter.format(instance, self.get_fewshot())
 
     def extract_answer(self, output: LMOutput) -> str:
         return _extract(output.text or "").answer
--- a/src/olmo_eval/evals/tasks/omega_500.py
+++ b/src/olmo_eval/evals/tasks/omega_500.py
@@ -108,8 +108,10 @@ class Omega500(Task):
         AccuracyMetric(name="exact_match_flex", scorer=_FLEX),
     )
     primary_metric = AccuracyMetric(name="exact_match_flex", scorer=_FLEX)
+    # max_tokens=None generates to the model's context limit, matching the
+    # reference regime's effective behavior on any context size.
     sampling_params = SamplingParams(
-        max_tokens=131072,
+        max_tokens=None,
         temperature=0.6,
         top_p=0.95,
     )
--- a/src/olmo_eval/evals/tasks/omega_500.py
+++ b/src/olmo_eval/evals/tasks/omega_500.py
@@ -1,0 +1,143 @@
+"""OMEGA-500: out-of-distribution math reasoning.
+
+OMEGA (https://arxiv.org/abs/2506.18880) evaluates exploratory,
+compositional, and transformative generalization in math. OMEGA-500 is a
+500-problem subset spanning the benchmark's sub-categories. Answers are
+extracted from the response via a regex cascade and compared to the ground
+truth, with a strict metric that demotes answers not stated in the
+requested format and a flex metric that accepts them.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Iterator
+from dataclasses import dataclass
+from typing import Any
+
+from olmo_eval.common.answer_extraction import ExtractedAnswer, extract_answer_with_format
+from olmo_eval.common.metrics import AccuracyMetric
+from olmo_eval.common.scorers.base import Scorer
+from olmo_eval.common.types import (
+    Instance,
+    LMOutput,
+    LMRequest,
+    RequestType,
+    SamplingParams,
+    Split,
+)
+from olmo_eval.data import DataSource
+from olmo_eval.evals.tasks.common import Task, register
+
+_BOXED_SUFFIX = "\n\nPresent the answer in LaTex format: \\boxed{Your answer}"
+
+_TEMPLATE_SUFFIX = (
+    '\n\nShow your work and conclude with "Therefore, the final answer is '
+    '\\boxed{answer}." where answer is just the final solution that solves the '
+    'problem. E.g., if the answer is 6, conclude with "Therefore, the final '
+    'answer is \\boxed{6}."'
+)
+
+_ANSWER_FORMAT_REGEX = r"Therefore, the final answer is \\boxed\{(.*)\}"
+_PREFIX_REGEXES = (
+    r"(?i)Therefore,? the final answer is",
+    r"(?i)Therefore,? the answer is",
+    r"(?i)the final answer is",
+    r"(?i)the answer is",
+    r"(?i)answer is",
+    r"(?i)answer is:",
+    r"(?i)answer:",
+)
+_ANSWER_REGEXES = (r"[\s\S]*?\\boxed\{(.*?)\}[\s\S]*?", r"(.*)\.?")
+_FORMAT_CORRECT_CUTOFF = 0.4
+
+# Mathy delimiters stripped when they surround the whole answer text.
+_DELIMITERS_TO_STRIP = (
+    ("$", "$"),
+    ("\\(", "\\)"),
+    ("**", "**"),
+    ("***", "***"),
+    ("\\[", "\\]"),
+    ("\\[\n", "\n\\]"),
+)
+
+
+def _extract(continuation: str) -> ExtractedAnswer:
+    """Normalize a continuation and extract the final answer."""
+    output = re.sub(r"(\d),(\d)", r"\1\2", continuation)
+    res = re.sub(r"\.\s*$", "", output).strip()
+    for left, right in _DELIMITERS_TO_STRIP:
+        res = re.sub(f"{re.escape(left)}(.*){re.escape(right)}", "\\1", res).strip()
+    return extract_answer_with_format(
+        res,
+        answer_format_regex=_ANSWER_FORMAT_REGEX,
+        answer_regexes=_ANSWER_REGEXES,
+        prefix_regexes=_PREFIX_REGEXES,
+    )
+
+
+@dataclass(frozen=True, slots=True)
+class OmegaExactMatchScorer(Scorer):
+    """Case-insensitive exact match on the extracted answer.
+
+    The strict form treats an answer as missing when it was not stated in
+    the requested format; the flex form accepts it regardless.
+    """
+
+    name: str = "exact_match"
+    flex: bool = False
+
+    def score(self, instance: Instance, output: LMOutput) -> float:
+        answer, format_correct = _extract(output.text or "")
+        if not self.flex and format_correct < _FORMAT_CORRECT_CUTOFF:
+            answer = ""
+        gold = str(instance.gold_answer or "")
+        return 1.0 if answer.lower() == gold.lower() else 0.0
+
+
+_STRICT = OmegaExactMatchScorer()
+_FLEX = OmegaExactMatchScorer(name="exact_match_flex", flex=True)
+
+
+@register("omega_500")
+class Omega500(Task):
+    data_source = DataSource(path="saumyamalik/omega-500", split="train")
+    split = Split.TRAIN
+    metrics = (
+        AccuracyMetric(name="exact_match", scorer=_STRICT),
+        AccuracyMetric(name="exact_match_flex", scorer=_FLEX),
+    )
+    primary_metric = AccuracyMetric(name="exact_match_flex", scorer=_FLEX)
+    sampling_params = SamplingParams(
+        max_tokens=131072,
+        temperature=0.6,
+        top_p=0.95,
+    )
+
+    @property
+    def request_type(self) -> RequestType:
+        return RequestType.CHAT
+
+    @property
+    def instances(self) -> Iterator[Instance]:
+        yield from self._load_instances_cached()
+
+    def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance | None:
+        question = str(doc["messages"][0]["content"]).replace(_BOXED_SUFFIX, "")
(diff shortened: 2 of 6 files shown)


abhishekraok:
Third and final omega parity cell — Qwen/Qwen3-8B (official), n=500:

| harness | budget | exact_match_flex | at ceiling |
|---|---|---|---|
| oe-eval | ~39,864 effective | 0.464 | 0% |
| olmo-eval (this PR) | 32,768 | 0.462 | 5.0% |

diff 0.002, SE 0.0315, z=0.06 → agree. (oe-eval mean output 16,450 tokens vs olmo-eval median 15,063 — same generation behavior.) The 5% cap asymmetry is bounded by the budget-decomposition result above (≲1 point, ns, and directionally it would only *lower* the capped side, which scored equal).

Complete merge-gate summary for this PR:

| model | oe-eval | olmo-eval | z | note |
|---|---|---|---|---|
| Olmo-3-7B-Instruct (official) | 0.274 flex | 0.304 flex | ≈1.1 SE | both ~uncapped |
| dense Olmo3-7B SFT | 0.100 | 0.086 | 0.76 | budget-matched @32k |
| Qwen3-8B (official) | 0.464 | 0.462 | 0.06 | 5% cap on olmo-eval side |

All three agree, across three models and two budget regimes, on top of the byte-identical prompt/scorer verification (all 500 docs, 544 scorer cases) already in the PR description. Parity evidence complete.

abhishekraok:
**Paired offline scorer check — exact agreement on both models, zero sampling noise.**

Took oe-eval's own stored omega_500 generations and scored them with this PR's `OmegaExactMatchScorer` (strict + flex), comparing per-instance against oe-eval's own verdicts. Identical inputs both sides:

| model | n | strict agree | flex agree | oe (strict/flex) | this PR (strict/flex) |
|---|---|---|---|---|---|
| Olmo-3-7B-Instruct | 500 | 500/500 (100%) | 500/500 (100%) | 0.2600 / 0.2740 | 0.2600 / 0.2740 |
| Qwen3-8B | 500 | 500/500 (100%) | 500/500 (100%) | 0.4640 / 0.4640 | 0.4640 / 0.4640 |

Aggregates match to four decimals because every per-instance verdict matches — this exercises the full regex cascade in `common/answer_extraction.py` (answer-format regex → prefix regexes → bare fallbacks, plus comma normalization and delimiter stripping) against 1,000 real reasoning traces including the truncated/degenerate tail. Much stronger than the sampled cross-harness cells above, which at n=500 and T=0.6 carry SE ≈ 0.018.

**Incidental performance note:** scoring is CPU-cheap for normal traces but not for very long ones — the 500 Olmo traces scored in 1 second, the 500 Qwen traces (mean ~16.4k tokens, max ~40k) took 964 seconds. The `(.*)\.?` and `[\s\S]*?\\boxed\{(.*?)\}[\s\S]*?` fallback patterns are the likely cost on long inputs. Not a correctness issue and not blocking, but worth knowing before anyone scores a large reasoning-heavy suite on CPU, and worth a look if scoring ever lands on the critical path.

abhishekraok:
## Parity evidence (complete)

Consolidates the two per-cell comments previously posted here plus the paired scorer check.

### Cross-harness score parity — `omega_500:0-shot-chat_deepseek` vs `omega_500`, n=500, `exact_match_flex`

| model | oe-eval | olmo-eval | z | budgets | at ceiling (oe / olmo) |
|---|---|---|---|---|---|
| Olmo-3-7B-Instruct (official) | 0.274 | 0.304 | ≈1.1 SE | both ~uncapped | — |
| dense Olmo3-7B SFT | 0.100 | 0.086 | 0.76 | ~31,819 vs 32,768 (matched) | 10.8% / 7.6% |
| Qwen3-8B (official) | 0.464 | 0.462 | 0.06 | ~39,864 vs 32,768 | 0% / 5.0% |

All three agree. n=500 at T=0.6 is inherently a wide cell (SE ≈ 0.018–0.031).

Budget note: oe-eval's *effective* omega budget under `--max-length 32768` is ~31,819 tokens, readable from its own predictions (54/500 items sit exactly there) — so olmo-eval @32,768 is the matched setting, not a handicap. On Qwen the generation behavior matches directly: oe-eval mean output 16,450 tokens vs olmo-eval median 15,063.

### Runs

| cell | harness | experiment | result dataset |
|---|---|---|---|
| Olmo-3-7B-Instruct | oe-eval | `01M0MYKGN52QMVWXJD9KW9PT3D` | `01M0NKZ01X8CARTWD7R46ZSZHX` |
| Olmo-3-7B-Instruct | olmo-eval | `01M0Q1YB4N3WA1YZTH92Q6DE9P` | — |
| Qwen3-8B | oe-eval | `01M0P4FSQ51SQ0BG66BQY804D6` | `01M0SG1YQ7ZX4ZKJEVCSF7KWE7` |
| Qwen3-8B | olmo-eval | `01M0R6SP8ERWSY5RNYVXFGE33G` | — |

The two result datasets hold the per-instance predictions the paired check below ran on, so the 500/500 agreement is reproducible from them without a GPU. Spot-checked `01M0NKZ01X8CARTWD7R46ZSZHX`: `exact_match 0.26 / exact_match_flex 0.274`, `num_instances 500`, `max_tokens_reached 0.0` — matching the table above. The budget-matched dense-SFT cell came from the cross-architecture SFT session and its run IDs aren't recorded here.

### Generation budget is not a confound

Three-way decomposition on the dense SFT model:

| setting | exact_match_flex |
|---|---|
| olmo-eval @8,192 | 0.078 |
| olmo-eval @32,768 | 0.086 |
| oe-eval @~31,819 | 0.100 |

Budget effect z=0.46, harness effect z=0.76, combined z=1.22 — all ns. **Quadrupling the generation budget buys ≲1 point, indistinguishable from zero**: runaway items don't convert to correct however long they run. That is the empirical justification for capped runs (previously argued only from first principles), and it makes capped cells legitimate parity evidence with a truncation audit attached. It also bounds the 5% cap asymmetry on the Qwen cell — directionally that would only *lower* the capped side, which scored equal.

### Paired offline scorer check — zero sampling noise

oe-eval's own stored generations scored with this PR's `OmegaExactMatchScorer` (strict + flex), compared per instance against oe-eval's own verdicts. Identical inputs both sides:

| model | n | strict agree | flex agree | oe (strict/flex) | this PR (strict/flex) |
|---|---|---|---|---|---|
| Olmo-3-7B-Instruct | 500 | 500/500 (100%) | 500/500 (100%) | 0.2600 / 0.2740 | 0.2600 / 0.2740 |
| Qwen3-8B | 500 | 500/500 (100%) | 500/500 (100%) | 0.4640 / 0.4640 | 0.4640 / 0.4640 |

Aggregates match to four decimals because every per-instance verdict matches. This exercises the full regex cascade in `common/answer_extraction.py` — answer-format regex → prefix regexes → bare fallbacks, plus comma normalization and delimiter stripping — against 1,000 real reasoning traces including the truncated/degenerate tail. Much stronger than the sampled cells above.

### Incidental: scoring cost on long traces

Scoring is CPU-cheap for normal traces but not for very long ones — the 500 Olmo traces scored in ~1 second, the 500 Qwen traces (mean ~16.4k tokens, max ~40k) took ~964 seconds. The `(.*)\.?` and `[\s\S]*?\\boxed\{(.*?)\}[\s\S]*?` fallback patterns are the likely cost on long inputs. Not a correctness issue and not blocking, but worth knowing before scoring a large reasoning-heavy suite on CPU, and worth a look if scoring ever lands on the critical path.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+    split = Split.TRAIN
+    metrics = (
+        AccuracyMetric(name="exact_match", scorer=_STRICT),
+        AccuracyMetric(name="exact_match_flex", scorer=_FLEX),
+    )
+    primary_metric = AccuracyMetric(name="exact_match_flex", scorer=_FLEX)
+    # max_tokens=None generates to the model's context limit, matching the
+    # reference regime's effective behavior on any context size.
+    sampling_params = SamplingParams(
+        max_tokens=None,
```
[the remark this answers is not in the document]
abhishekraok: Confirmed — good catch, and it was broader than this task. Fixed in #319.

Verified the mechanism: `describe_request()` builds generation kwargs *without* calling `_resolve_max_tokens()`, and `runners/asynq/processing.py:110` traces every request before generation, so `params.max_tokens <= 0` raises `TypeError: '<=' not supported between instances of 'NoneType' and 'int'` before inference starts. The generate path was fine because it resolves first.

This hits every task using `max_tokens=None`, not just OMEGA — also #307 `popqa:chat` and #309 `gpqa_*:cot`. It went unnoticed because the parity runs used `vllm_server`.

Fixed in the provider rather than by reverting to a concrete cap: `max_tokens=None` is deliberate here (it avoids the fixed-131k footgun in #310, which silently 400s every request on models with smaller context). #319 lets `None` through the guard while still rejecting a concrete non-positive limit, with a regression test.

src/olmo_eval/evals/extract/answer_format.py
```diff
@@ -0,0 +1,104 @@
...
+                            answer = group
+                else:
+                    answer = res
+                if answer:
+                    return ExtractedAnswer(answer, 1.0 if idx == 0 else 0.5)
+
+    for idx, answer_regex in enumerate(answer_regexes):
+        ans_match = re.findall(answer_regex, continuation)
+        if ans_match:
+            answer = ans_match[-1] if use_last_raw_match else ans_match[0]
```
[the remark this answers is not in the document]
abhishekraok: Accurate description of the behavior, but this is a deliberate port and oe-eval does exactly the same thing — so changing it here would break the parity this PR exists to establish.

oe-eval's `extract_answer` (`oe_eval/tasks/utils.py`) has the identical final fallback, with no empty-match filtering:

```python
for idx, answer_regex in enumerate(answer_regexes):
    ans_match = re.findall(answer_regex, continuation)
    if ans_match:
        answer_string = ans_match[-1] if use_last_raw_match else ans_match[0]
```

Ran both implementations on your example:

| input | oe-eval | this PR |
|---|---|---|
| `42` | `('', 0.1)` | `('', 0.1)` |

So a bare unformatted `42` scores 0 on both metrics in oe-eval too. Filtering the empty match would make olmo-eval score responses oe-eval scores 0, and the 500/500 per-instance agreement reported above would no longer hold.

Worth fixing — but upstream, so both harnesses move together. Filing it there rather than diverging here.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+_ANSWER_FORMAT_REGEX = r"Therefore, the final answer is \\boxed\{(.*)\}"
+_PREFIX_REGEXES = (
+    r"(?i)Therefore,? the final answer is",
+    r"(?i)Therefore,? the answer is",
+    r"(?i)the final answer is",
+    r"(?i)the answer is",
+    r"(?i)answer is",
+    r"(?i)answer is:",
```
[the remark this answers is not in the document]
abhishekraok: Same situation as the empty-match finding: real behavior, faithfully ported, and oe-eval's prefix list has the identical order.

From `oe_eval/configs/tasks.py`, `omega_500`'s `answer_prefix_regexes`:

```
"(?i)Therefore,? the final answer is", "(?i)Therefore,? the answer is",
"(?i)the final answer is", "(?i)the answer is", "(?i)answer is",
"(?i)answer is:", "(?i)answer:"
```

`the answer is` precedes `answer is:` there too, so the colon-bearing entries are equally unreachable upstream. Ran both on your example:

| input | oe-eval | this PR |
|---|---|---|
| `The answer is: 42` | `(': 42', 0.5)` | `(': 42', 0.5)` |
| `The answer is 42` | `('42', 0.5)` | `('42', 0.5)` |

Both extract the leading colon and both fail the comparison. Reordering the list here would change scores relative to oe-eval on exactly the responses this PR is being certified against.

Also worth fixing upstream rather than diverging. Same for the `answer_regexes` and the 0.4 cutoff — all copied verbatim.

src/olmo_eval/evals/extract/answer_format.py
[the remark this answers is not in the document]
undfined: A note on this; I've typically matched and then versioned as a v2 or something with fixes ignoring oe-eval backporting.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+    metrics = (
+        AccuracyMetric(name="exact_match", scorer=_STRICT),
+        AccuracyMetric(name="exact_match_flex", scorer=_FLEX),
+    )
+    primary_metric = AccuracyMetric(name="exact_match_flex", scorer=_FLEX)
+    # max_tokens=None generates to the model's context limit, matching the
+    # reference regime's effective behavior on any context size.
+    sampling_params = SamplingParams(
+        max_tokens=None,
+        temperature=0.6,
```
undfined, line 140: These sampling params look like they come from omega_500:0-shot-chat_deepseek config. oe-eval's plain omega_500:0-shot-chat uses temperature=0.7 and no top_p.

Should we rename this to reflect the deepseek regime?

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+        return RequestType.CHAT
+
+    @property
+    def instances(self) -> Iterator[Instance]:
+        yield from self._load_instances_cached()
+
+    def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance | None:
+        question = str(doc["messages"][0]["content"]).replace(_BOXED_SUFFIX, "")
+        return Instance(
+            question=question + _TEMPLATE_SUFFIX,
```
undfined: The prompt suffix is concatenated onto the question inside process_doc, and the regexes are module-level constants that _STRICT and _FLEX close over. Both are fixed at
  import time, so register_variant cannot override either one.

  That blocks the other three upstream configs:
  - 0-shot-chat needs a different temperature
  - olmo3:midtrain needs stop sequences
  - rlzero needs a different prompt template and a different set of regexes

aime.py:32 handles this by putting the CoT suffix in ChatFormatter(user_template=...). register_variant can override a formatter. Moving the suffix there now would make those variants a config change later instead of a refactor.

src/olmo_eval/evals/extract/answer_format.py
```diff
@@ -0,0 +1,104 @@
+"""Regex-cascade answer extraction for generative tasks.
```
undfined, line 1: This module does answer extraction, but it is not in the answer extraction package.

src/olmo_eval/evals/extract/ already holds extract_math_answer, extract_mcq_answer, and MathExtractor, and exports them through __all__. This new module is in common/ and is not exported anywhere.

Suggest moving it to evals/extract/answer_format.py and adding it to that package's __all__.

src/olmo_eval/evals/extract/answer_format.py
```diff
@@ -0,0 +1,104 @@
...
+    answer: str
+    format_correct: float
+
+
+def extract_answer_with_format(
+    continuation: str,
+    *,
+    answer_format_regex: str | None = None,
+    answer_regexes: Sequence[str] | None = None,
+    answer_regexes_templates: Sequence[str] | None = None,
```
undfined, line 33: Three parameters here have no callers.

omega_500.py:74 is the only call site. It passes answer_format_regex, answer_regexes, and prefix_regexes. It never passes answer_regexes_templates, and it never overrides use_last_prefix_match or use_last_raw_match.

  The template branch at lines 62-76 is about fifteen lines that nothing can currently reach.

  If future ported tasks will need these let's add a comment otherwise maybe we can drop this.

src/olmo_eval/evals/extract/answer_format.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class ExtractedAnswer(NamedTuple):
+    """An extracted answer string and its format-correctness score."""
+
+    answer: str
+    format_correct: float
+
+
+def extract_answer_with_format(
```
undfined, line 28: This function has no direct tests. It is only tested through the omega scorers, which exercise one path through it.

  Not covered:
  - the 0.2 and 0.1 fallback scores, as distinct from each other
  - use_last_prefix_match=False
  - use_last_raw_match=False
  - the entire answer_regexes_templates branch

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+
+    The strict form treats an answer as missing when it was not stated in
+    the requested format; the flex form accepts it regardless.
+    """
+
+    name: str = "exact_match"
+    flex: bool = False
+
+    def score(self, instance: Instance, output: LMOutput) -> float:
+        answer, format_correct = _extract(output.text or "")
```
undfined, line 108: _extract returns format_correct, and this line uses it only to decide the strict/flex cutoff. Then it is discarded.

oe-eval reports this value as a metric (omega.py:75). It answers a specific question: when the strict score is low, was the model wrong, or did it just format the answer differently?

Writing it to output.metadata would make it available in results.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+        return 1.0 if answer.lower() == gold.lower() else 0.0
+
+
+_STRICT = OmegaExactMatchScorer()
+_FLEX = OmegaExactMatchScorer(name="exact_match_flex", flex=True)
+
+
+@register("omega_500")
+class Omega500(Task):
+    data_source = DataSource(path="saumyamalik/omega-500", split="train")
```
undfined: `split="train"` here has no effect.

TaskConfig.get_data_source (common/base.py:212-218) always calls with_split(self.split.value), so the class-level split = Split.TRAIN on line 108 overwrites this argument every time.

Both values are "train", so nothing is broken. But the argument looks like it controls the split, and it does not. Suggest removing it.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -0,0 +1,148 @@
...
+@dataclass(frozen=True, slots=True)
+class OmegaExactMatchScorer(Scorer):
+    """Case-insensitive exact match on the extracted answer.
+
+    The strict form treats an answer as missing when it was not stated in
+    the requested format; the flex form accepts it regardless.
+    """
+
+    name: str = "exact_match"
+    flex: bool = False
```
undfined: name and flex are separate fields, and every caller has to set both correctly.

OmegaExactMatchScorer(flex=True) gets the default name "exact_match". Task._get_scorers (common/base.py:554) keys scorers by name and skips duplicates, so that scorer would be dropped and its metric would report 0.0. No error, no warning.

Lines 101-102 set both fields correctly, so this is not a bug today but likely a foot gun down the road.

Consider setting name from flex in __post_init__, or delete the flex field and derive the behavior from the name.

undfined (approved):
A few suggestions but otherwise looks good.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Fixed in a838ccd — the argument is gone.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Fixed in a838ccd: the scorer now writes `answer_format_correct` to `output.metadata`, so results carry the wrong-vs-badly-formatted distinction the reference reports.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Fixed in a838ccd: `__post_init__` derives the name from `flex` unless set explicitly, so `OmegaExactMatchScorer(flex=True)` now gets `exact_match_flex` instead of silently colliding with the strict scorer and being dropped.

src/olmo_eval/evals/extract/answer_format.py
abhishekraok: Fixed in a838ccd: `tests/evals/extract/test_answer_format.py` covers every cascade stage's format score (1.0 / 0.5 / 0.2 / 0.1 as distinct cases), both `use_last_*` flags in both settings, and the templates branch.

src/olmo_eval/evals/extract/answer_format.py
abhishekraok: Fixed in a838ccd: moved to `evals/extract/answer_format.py` and exported through the package `__all__`. The stacked PRs (#309, #321) are rebased over the move with their imports updated.

src/olmo_eval/evals/extract/answer_format.py
abhishekraok: Half right — kept with a comment (a838ccd). `answer_regexes_templates` does have callers: the stacked GPQA (#309) and MMLU (#321) CoT tasks pass `OLMO_3_ANSWER_REGEX_TEMPLATES` through it; reviewing #308 in isolation hides them. The `use_last_*` flags genuinely have no caller overriding the defaults (ours or oe-eval's configs) — kept to mirror the reference `extract_answer` contract, noted as such, and now covered by direct tests either way.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Adopted in a838ccd, aime-style: the CoT suffix now lives in `ChatFormatter(user_template=...)` (braces escaped), `process_doc` stores the bare question, and `format_request` defers to `config.formatter` — so the upstream 0-shot-chat / midtrain / rlzero regimes become `register_variant` config (template, sampling, stops) rather than a refactor. The regexes stay module constants for now; the rlzero variant can pass its own set to the extractor when it lands. Emitted requests verified byte-identical on all 500 docs before/after, so the parity certification stands.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Keeping the name and documenting the regime rather than renaming, unless you feel strongly. Reasons: (1) precedent — merged #307's `popqa:chat` mirrors `popqa::hamish_zs_reasoning_deepseek` without carrying the deepseek name; (2) this config is the OLMO_3 suite entry, i.e. the setting the published numbers and the parity certification use, which is what an unqualified `omega_500` should mean here; (3) with the formatter refactor above, plain `omega_500:0shot-chat` (temp 0.7, no top_p) becomes a straightforward variant if anyone needs it, which resolves the ambiguity in the direction of more tasks rather than a rename. Docstring now states which upstream config the defaults mirror.

src/olmo_eval/evals/extract/answer_format.py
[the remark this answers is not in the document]
abhishekraok: That pattern fits here well — matched v1 for comparability, then a `:v2` variant with the extraction fixes, no oe-eval backporting required. Noted it on #320 as the disposition for the two extraction bugs.

src/olmo_eval/evals/extract/answer_format.py
abhishekraok: Behavior-preservation evidence for the move + formatter refactor, all three cascade paths re-verified against real stored generations at the new heads: omega paired scorer 500/500 strict and flex (aggregates equal, 0.4640); gpqa paired scorer 448/448 on both models (templates path, aggregates equal to five decimals); mmlu:cot 3,207/3,207 prompts+golds byte-identical, scorer 56/56. Emitted omega requests also byte-identical on all 500 docs.

PR #309, opened 2026-08-22
Add GPQA 0-shot CoT tasks (post-training regime)

abhishekraok:
Stacked on #308 (uses the shared answer-extraction cascade; will retarget to main once #308 merges).

Registers `gpqa_{diamond,main,extended}:cot` matching oe-eval's `gpqa:0shot_cot::qwen3-instruct` (the OLMO_3 suite entry): inline choices, boxed-letter instruction, temp 0.6 / top-p 0.95 / 131k tokens, OLMO_3 answer-regex templates. The regime's choice shuffle (seed 111 + index, oe-eval's list order) differs from the base gpqa tasks' shuffle, so this is a separate task class with its own doc processing — base `gpqa_*` numbers are unaffected.

- Prompts and gold letters identical on 16 synthetic docs covering the preprocessing edge cases ([title] markers, bracket stripping, double spaces)
- Scorer verdicts identical on 64 response/label cases
- 90 gpqa tests pass; ruff/ty clean

**Before undraft:** accept the `Idavidrein/gpqa` HF gate on the eval account, re-run parity on real docs, then GPU parity run.


abhishekraok:
## Verification (complete)

Consolidates the three comments previously posted here. One of them advanced a claim that a later comment retracted; the retracted claim is simply gone rather than restated and corrected.

### Prompt/gold parity against oe-eval, real data

Prompts and gold letters are byte-identical to oe-eval's `gpqa:0shot_cot::qwen3-instruct` construction on **all 448 `gpqa_main` and 198 `gpqa_diamond` docs** (the `Idavidrein/gpqa` gate is accepted). Shuffle seed, preprocessing and template all match.

Note the shuffle here deliberately differs from base `gpqa_diamond`: seed 111 + index over oe-eval's list order, which is what the post-training regime uses. Base GPQA doc processing is unaffected — verified identical to `main` for all three subsets.

### Production runs

OLMoE3-KDA MoE via the #312 path, run `01M0R4PSF4QZCJQ4T8NH7DH4A4` (from the cross-architecture SFT session):

| task | score | n | median tokens | truncated @8192 |
|---|---|---|---|---|
| `gpqa_main:cot` | 0.3058 | 448 | 435 | 0.0% |
| `gpqa_diamond:cot` | 0.2980 | 198 | — | 0.5% |

Both clean end-to-end. Cross-harness score cells for `:cot` are not yet run.

### What the CoT regime actually buys

Paired McNemar comparison against the non-CoT MCQ variant over identical items, six model×subset pairs: **no model shows a significant CoT uplift on either subset** (closest is dense diamond, χ²=3.52 < 3.84). CoT does not score higher in aggregate.

What it does do is reshuffle *which* items are correct without changing how many. Dense on `gpqa_main`: 83 items flip wrong→right, 79 flip right→wrong — 36% of items change state for a net +4, and the pattern is near-symmetric for every model×subset. The two regimes are not measuring the same thing item-for-item even when aggregates agree, so a model selected on MCQ is not necessarily the model you would select on CoT. That, plus matching other labs' published numbers, is the argument for these variants — not an uplift.

### Scope caveat for the parity work generally

Aggregate equality can hide large per-item disagreement — observed elsewhere in this effort: two models tying at 59/198 while agreeing on only 16 of the 59. The cross-harness parity claimed on #307/#308/#309 is **score parity**, which is what these merge gates assert. Item-level agreement of generations across harnesses is untested and not expected at T=0.6. (For #307's and #308's *scorers*, the paired offline checks on those PRs do establish per-instance agreement — a different and stronger claim about the scorer alone.)

abhishekraok:
**GPU parity — the gap this PR was missing is now closed.** Olmo-3-7B-Instruct, gpqa_main, n=448 both sides:

| harness | spec | exact_match |
|---|---|---|
| oe-eval | `gpqa:0shot_cot::qwen3-instruct` | 0.42857 |
| olmo-eval | `gpqa_main:cot` | 0.41964 |

diff 0.009, z ≈ 0.27 → agree. (Beaker: olmo-eval `01M0X48ANW8WH0XM1AXBY8XCVV`, oe-eval `01M0X48YNCBYV7FZHPMT97EAND`, results `01M0X6HRZQPXGY8QD65D42WH3W`.) Same run also produced `gpqa_diamond:cot` = 0.43939.

**Paired scorer check on the same 448 generations** — oe-eval's stored output scored by this PR's `GPQACoTExactMatchScorer`, compared per-instance to oe-eval's own verdicts:

```
per-instance agreement : 448/448 = 100.00%
oe-eval                : 0.42857
this PR's scorer       : 0.42857
paired diff            : 0.00000
```

That's the strong form: n=448 at T=0.6 gives SE_diff ≈ 0.033, so the aggregate cell alone only bounds gross error, while the paired check shows the two extractors never disagree on a single real generation. Note `answer_format_correct` averages 0.51 on this run — about half the responses don't use the requested phrasing and are recovered by lower-priority templates in the cascade — so the agreement is exercising the fallback tiers, not just the happy path.

Qwen3-8B pair is still running; will add if it changes anything.

abhishekraok:
Second model cell — Qwen/Qwen3-8B, gpqa_main, n=448 (Beaker: olmo-eval `01M0X48JSV8XZWGZE7RT8ENETG` results `01M0X4MSK10S17EZKCQMGWKGFD`; oe-eval `01M0X4901ZXVV89ACDEGXWZ3QZ` results `01M0X97G7JWE00EFSEE6J1ZK1N`):

| harness | exact_match |
|---|---|
| oe-eval | 0.56920 |
| olmo-eval | 0.58482 |

diff 0.0156, z ≈ 0.47 → agree. Paired scorer check on oe-eval's 448 stored generations: **448/448 agreement**, aggregates equal to five decimals. Same run also produced `gpqa_diamond:cot` = 0.61111 (no oe-eval counterpart run for diamond).

Cumulative for this PR: 2 models × aggregate agreement (z = 0.27, 0.47), 896/896 paired scorer agreement, prompts byte-identical on all 646 real docs. Evidence complete.

src/olmo_eval/evals/tasks/gpqa.py
```diff
@@ -234,6 +240,115 @@ def extract_answer(self, output: LMOutput) -> str | None:
...
+)
+
+
+def _extract_cot_answer(text: str) -> str:
+    answer = extract_answer_with_format(
+        text,
+        answer_format_regex=_COT_ANSWER_FORMAT_REGEX,
+        answer_regexes=_COT_ANSWER_REGEXES,
+        answer_regexes_templates=OLMO_3_ANSWER_REGEX_TEMPLATES,
+    ).answer
```
undfined: Could we preserve `format_correct` from `extract_answer_with_format()` instead of discarding it here? The reference GPQA evaluation reports answer-format correctness, and the parity results suggest it is a useful diagnostic—roughly half the responses relied on fallback extraction. At minimum, it would be helpful to store this value in the output metadata, consistent with the OMEGA-500 scorer.

src/olmo_eval/evals/tasks/gpqa.py
```diff
@@ -272,6 +387,20 @@ def _clean_text(text: str) -> str:
     register_variant(_subset, "mc", formatter=MultipleChoiceFormatter(), metrics=_DEFAULT_METRICS)
     register_variant(_subset, "bpb", formatter=PPLFormatter(), metrics=(BPBMetricInstanceAvg(),))
 
+    # 0-shot CoT (post-training regime): needs its own doc processing, so it is
```
undfined, line 393: Small documentation thought: since the task count and supported variants are already changing, should we avoid hard-coding an exact count in the module docstring? Describing the naming scheme and available variants without a total might keep it from becoming stale as more GPQA configurations are added.

undfined (approved):

donovanr:
Didn't spat anything that Tyler didn't – looks pretty good to me.

donovanr (approved):

src/olmo_eval/evals/tasks/gpqa.py
abhishekraok: Done — the scorer now records `answer_format_correct` in output metadata, same as the omega_500 scorer. Scores are unchanged (the extractor already computed the value; it was just being dropped). Test added covering both the requested-format and fallback-extraction cases.

src/olmo_eval/evals/tasks/gpqa.py
abhishekraok: Done — the count is gone; the docstring now describes the naming scheme (`gpqa_{subset}`, `gpqa_{subset}_{subject}`) and lists the `:mc` / `:bpb` / `:cot` variants without a total.

PR #312, opened 2026-08-23
Exclude Torch companion packages and nvidia-cutlass-dsl from CUDA constraints

abhishekraok:
`/tmp/cuda-constraints.txt` froze everything matching `^(torch|nvidia-)`, which also catches `torchvision`/`torchaudio` and `nvidia-cutlass-dsl` — pinning them to the base image and breaking any provider release that swaps torch at runtime (OLMoE3/KDA plugin stack: `failed to remove .../torchgen`). Now only `torch` itself (`torch==` / `torch @ `) and `nvidia-*` are constrained.

The first two commits are @akshitab's, cherry-picked from #303, which targets `jacksonp/ladder` and so cannot reach `main`; `launcher.py` here is byte-identical to it. The third adds a regression test that fails against the old pattern.

Validated on the OLMoE3-KDA MoE: Beaker `01M0R4PSF4QZCJQ4T8NH7DH4A4` — torch 2.10.0+cu128 swap completed, plugin registered, no `torchgen` failure. That run predates the cherry-pick rework, which left the emitted install command byte-identical.

### DIFF
--- a/src/olmo_eval/launch/beaker/launcher.py
+++ b/src/olmo_eval/launch/beaker/launcher.py
@@ -892,7 +892,11 @@ class BeakerLauncher:
         # an otherwise compatible vLLM impossible to resolve.
         constraints = "/tmp/cuda-constraints.txt"
         steps.append(
-            f"uv pip freeze -q | grep -E '^(torch|nvidia-)' "
+            # Match torch itself, but not companion packages such as
+            # torchvision or torchaudio. Provider releases pin those companions
+            # alongside their Torch generation and must be allowed to replace
+            # the versions inherited from the base image.
+            f"uv pip freeze -q | grep -E '^(torch(==| @ )|nvidia-)' "
             f"| grep -vE '^nvidia-cutlass-dsl' > {constraints}"
         )
 
--- a/src/olmo_eval/launch/beaker/launcher.py
+++ b/src/olmo_eval/launch/beaker/launcher.py
@@ -886,9 +886,15 @@ class BeakerLauncher:
         if runtime_torch_version:
             steps.append(build_install_command(runtime_torch_spec(), force_reinstall=True))
 
-        # Generate constraints from pre-installed CUDA packages to prevent uv from changing them
+        # Generate constraints from pre-installed CUDA packages to prevent uv from changing them.
+        # CUTLASS DSL is the exception: vLLM pins this package independently of
+        # Torch's CUDA package set, so constraining an image's older build makes
+        # an otherwise compatible vLLM impossible to resolve.
         constraints = "/tmp/cuda-constraints.txt"
-        steps.append(f"uv pip freeze -q | grep -E '^(torch|nvidia-)' > {constraints}")
+        steps.append(
+            f"uv pip freeze -q | grep -E '^(torch|nvidia-)' "
+            f"| grep -vE '^nvidia-cutlass-dsl' > {constraints}"
+        )
 
         # Install vLLM in isolated venv when requested (for server mode)
         if use_isolated_vllm_venv:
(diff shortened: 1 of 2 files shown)


abhishekraok:
Validated against a real plugin-stack model (Beaker `01M0R4PSF4QZCJQ4T8NH7DH4A4`, run by the cross-architecture SFT eval session): OLMoE3 latent-KDA MoE (`Olmo3MoeForCausalLM`, 18.5B total / 1.3B active) on image `akshitab/olmo-core-tch2110cu128-rma-2026-08-04` with `OLMO_EVAL_RUNTIME_TORCH_VERSION=2.10.0+cu128`, installing vllm 0.19.1 + flash-linear-attention 0.4.1 + ai2-olmo-core over the constraints file. The runtime torch swap completed (`+ torch==2.10.0+cu128`) and the plugin registered (`Registered Olmo3MoeForCausalLM with the OLMoE3 vLLM plugin`) — no `failed to remove .../torchgen`, which is exactly the failure the old `^(torch|nvidia-)` pattern caused. First run of this path on anything other than the dangling original commit, confirming the reimplementation is behaviorally equivalent for the case the fix exists for.

(If you see `flash-attn 99.0.0` in that job's logs: a deliberate stub wheel on their side to evict the image's torch-2.11-built flash-attn; vLLM falls back cleanly. Unrelated to this PR.)

undfined:
Can you confirm whether this is a partial duplicate of: https://github.com/allenai/olmo-eval/pull/303

abhishekraok:
@undfined Confirmed — byte-identical to #303's hunk, not just partial.

My provenance claim was wrong: `97166c2` is not dangling. @AkshitaB's fix is alive in #303 as `3f9d764` + `c1fd424`; it just hasn't reached `main` because #303 targets `jacksonp/ladder`. Apologies for the mischaracterization.

This branch now cherry-picks Akshita's two commits with `-x` instead of reimplementing them, so `launcher.py` matches #303 byte-for-byte, plus one commit adding a regression test. Happy to close in favor of her splitting the hunk out of #303 — the only cost is the MoE staying blocked on `main` until then.

abhishekraok:
Confirmed, and it's a stronger overlap than partial — the constraints change here was **byte-identical** to the one in #303. Thanks for catching it.

To be precise about the relationship: #303 carries @akshitab's `3f9d764` ("Allow vLLM to select CUTLASS DSL") and `c1fd424` ("Exclude Torch companion packages from CUDA constraints"). This PR previously reimplemented that change and described it as recovering a dangling commit that "never landed on any branch" — that was wrong, and I've corrected the description. The change was never dangling; it lives in #303 on `akshitab/ladder-eval-runtime-fixes`.

I've rewritten this branch to cherry-pick Akshita's two commits unmodified rather than reimplement them, so authorship is preserved and `launcher.py` here is now byte-identical to #303's.

What's left that isn't duplicated: this targets `main`, and #303 targets `jacksonp/ladder`. The fix can't reach `main` via #303 without also taking that branch's six HF dataset-ID renames and the provider index-url change, which are independent and need their own review. Without this on `main`, the OLMoE3/KDA MoE can't run any migrated task, which blocks the oe-eval migration work. When #303 later rebases onto `main`, this hunk drops out of its diff on its own.

I also added the one thing neither PR had on `main`: a regression test for the constraints pattern, adapted from #303's assertions. It fails against the old `^(torch|nvidia-)` and passes against the new one.

Happy to close this instead if you'd rather Akshita split the launcher fix out of #303 onto `main` directly — that's cleaner provenance, at the cost of the MoE evals staying blocked until then. Your call.

abhishekraok:
I'm going to merge this in to unblock many of my evals.

PR #316, opened 2026-08-25
Multimodal eval (1/5): image requests, HF VLM path, OLMo-core VLM provider 

jason718:
## Description

First of five PRs splitting #278 into reviewable milestones.
**Milestone: any VLM checkpoint — released HF or raw OLMo-core — can be prompted with images.**
No tasks or benchmarks yet; those follow in #317 (framework + captioning) → #322 (pointing + counting) → #323 (image QA) → #324 (multi-image + docs).

### Core plumbing

- `LMRequest.images` and harness pass-through — tasks attach lazy image references (paths / picklable callables), providers resolve them at inference time
- **Image-request contract**: providers declare `supports_images`; the harness rejects image requests aimed at a provider that cannot render them, so a vision task on a text provider fails loudly instead of scoring text-only completions (mock opts in so vision suites stay dry-runnable)
- Provider kwargs passthrough in `ProviderConfig.from_dict` — unknown keys now reach the constructor for every provider kind, so a typo'd YAML key surfaces instead of being dropped silently (tested)
- Completion prompts survive harness-injected system messages: both multimodal providers build chat turns via a shared `chat_messages_for_request`, which appends `request.prompt` as a user turn whenever no user turn exists

### HF multimodal provider

`huggingface.py` gains an image-text-to-text path: `AutoProcessor` + `AutoModelForImageTextToText`, transformers-5 shims for the Molmo2 remote code, and fp32-weights/bf16-autocast matching mm_olmo's `amp_bf16` numerics. Presets: `molmo2-4b`, `molmo2-8b`, `molmo2-o-7b` (all mirror mm_olmo's official eval settings). The text path is untouched: `_init_text` is the old constructor verbatim, and every shim runs only inside the multimodal init.

### OLMo-core VLM provider — as a package

`inference/providers/olmo_core_vlm/`, split by responsibility:

| module | contents |
|---|---|
| `checkpoint.py` | format detection (raw DCP / consolidated export / mm_olmo trainer), config building, weight loading + key remap, untied-LM-head load guard |
| `preprocessing.py` | tokenizer / max-crops / max-length hints, Molmo2 image special tokens |
| `cache.py` | cache-capable dense-SDPA backend; the decode attention mask is a module-level helper with direct tests (causal prefill, image `or_mask`, left padding, decode) |
| `provider.py` | `OlmoCoreVLMProvider` — batched variable-length KV-cached decode loop |

The split is AST-verified content-identical to the reviewed #278 code. A prompt exceeding `max_length` raises (multimodal prompts cannot be truncated without corrupting the image-token layout) rather than scoring an empty output. `logprobs` mirrors `OlmoCoreProvider`'s boundary handling (per-request limits, empty/over-limit continuations) and emits `token_id`/`bytes` entries like the other providers.

### Dependency isolation (the part to scrutinize)

The multimodal classes are in **no ai2-olmo-core release** (2.6.0 ships `nn.vision` but rejects vision-branch configs and lacks `molmo2_loader`), so the provider has its own opt-in extra, and the text side is untouched:

- `olmo_core` extra: **unchanged** — `ai2-olmo-core==2.4.0`, with `torchao==0.17.0` pinned explicitly so the conflicting fork below cannot drag it down
- **New `olmo_core_vlm` extra**: `ai2-olmo-core @ git+…OLMo-core.git@2dad929a` (vision branch, pinned by SHA), declared **conflicting** with `olmo_core`/`openhands` — the two OLMo-cores can never coexist in one install. Beaker maps `provider.kind=olmo_core_vlm` to this extra
- `hf` extra: + `einops`, `torchvision` (required at import by the Molmo2 remote code); + `pycocotools`/`scipy` arrive with the pointing family in #322
- The default **vllm group is now self-contained** rather than routed through the conflict-gated vllm extra — routed through the extra, uv 0.10 resolved the default group to *nothing* and `uv sync` silently installed no vllm
- **Resolution parity with main, verified per profile via `uv export`**: `default+olmo_core` (the Beaker text-eval profile) is byte-identical to main's resolution; the CI profile differs only by pure additions; numpy/pandas/starlette are at main's exact versions
- `kind=olmo_core` still auto-detects multimodal checkpoints (returns to the text path on any read error) but logs a WARNING steering to the explicit kind, since remote jobs must install the matching extra

### Runner fixes (from running this at scale)

- Eager `tqdm.get_lock()` before parallel task preparation (lazily-created class lock races)
- `--secret-env` mappings satisfy `required_secrets` validation (previously only the user-scoped secret was accepted)
- `olmo_core_vlm` in `_SEQUENTIAL_ONLY`, package-override normalization for both OLMo-core kinds, mm_olmo unpickle shim is idempotent

## Testing

Everything below runs on CPU with no API keys:

```bash
uv sync --frozen
uv run ruff check src/ tests/ && uv run ruff format --check src/ tests/
uv run --frozen --no-group vllm ty check src/ alembic/     # what CI runs
uv run pytest tests/ --ignore=tests/integration -q         # 2,335 passed
```

Targeted suites: harness image pass-through + `supports_images` rejection (`tests/core/harness/test_harness.py`), HF multimodal shims (`tests/providers/test_huggingface_multimodal.py`), provider unit tests incl. key-remap regression, unpickle-shim idempotence, decode-mask cases, split-vocab embedding equivalence, logprob boundaries, untied-head guard, and `chat_messages_for_request` (`tests/providers/test_olmo_core_provider.py`).

GPU verification: checkpoint load + image pointing + text generation through the **declared** `olmo_core_vlm` extra (no overrides); out-of-band vs mm_olmo on real checkpoints — prompt construction token-exact (765/765 input_ids), KV-cached decode token-identical to uncached. End-to-end: the full stack ran caption/pointing/counting on a fresh stage-1 checkpoint on Beaker (5 tasks, 8,831 instances, zero failures).

## Review history

Three rounds, all findings addressed with fixes + tests: 7 from @undfined (shim idempotence, logprob boundaries, package normalization, embedding memory, batching allowlist, mask-helper extraction, field placement), 2 from @abhishekraok (installable extra, completion-prompt preservation), 7 from @gregorybchris's review agent (image-request contract, silent skip, lock fork, reroute warning, kwargs test, timeout parsing, logprob fields).

## Type of Change

## Checklist

Part of #278 (tracking). Next: #317 — vision task framework + captioning.


src/olmo_eval/inference/providers/olmo_core_vlm/checkpoint.py
```diff
@@ -0,0 +1,441 @@
...
+
+    mm_olmo distributed checkpoints pickle a custom ``_StorageInfo`` dataclass
+    (fields ``relative_path`` / ``offset`` / ``length``, mirroring torch's) into
+    ``model_and_optim/.metadata``. The mm_olmo package isn't installed in the
+    eval environment, so provide a minimal class with the same module path for
+    pickle to resolve. No-op when the real package (or the shim) is importable.
+    """
+    import importlib.util
+
+    try:
```
undfined, line 330: This crashes on the second invocation. The shim installed below has `__spec__ = None`, so `find_spec()` raises `ValueError` before reaching the existing `sys.modules` check.

I reproduced it with two consecutive calls:

```text
first ok
ValueError: olmo.train.remote_filesystem.__spec__ is None
```

This is reachable when multiple mm_olmo providers are constructed in one process, including a `ReplicaSet` with multiple instances. Check `sys.modules` first and catch `ValueError` defensively:

```python
module_name = "olmo.train.remote_filesystem"
if module_name in sys.modules:
    return

try:
    if importlib.util.find_spec(module_name) is not None:
        return
except (ImportError, ModuleNotFoundError, ValueError):
    pass
```

Please add a regression test that calls the helper twice.

src/olmo_eval/inference/providers/olmo_core_vlm/provider.py
```diff
@@ -0,0 +1,873 @@
...
+            request_outputs: list[LMOutput] = []
+            cont_prompts = request.continuation_prompts
+            for i, continuation in enumerate(request.continuations or ()):
+                prompt = cont_prompts[i] if cont_prompts else request.prompt
+                context_ids, continuation_ids = encode_context_and_continuation(
+                    self.tokenizer, prompt, continuation
+                )
+                if not context_ids:
+                    context_ids = [self.tokenizer.eos_token_id]
+                full_ids = (context_ids + continuation_ids)[-(self.max_length + 1) :]
```
undfined: `full_ids` is truncated to the context window, but `n_cont` still describes the original continuation. If the continuation exceeds the window, `cont_logits` has fewer rows than `cont_tensor` and `torch.gather()` fails with an internal shape error.

Empty continuations also fail because `-0` selects every logit row before the greedy comparison. This method additionally ignores `request.max_length`, unlike `OlmoCoreProvider`.

Could this mirror the text provider's boundary handling?

- Resolve the limit from `request.max_length` or `self.max_length`.
- Return an empty, zero-logprob result for an empty continuation.
- Raise a clear error when the continuation exceeds the limit.
- Truncate the context only after those checks.

Tests for empty, exactly-at-limit, over-limit, and per-request limits would cover the edge cases.

undfined (commented):
See inline comments.

src/olmo_eval/common/constants/infrastructure.py
```diff
@@ -193,6 +193,7 @@ def cluster_has_weka(cluster: str) -> bool:
     "vllm_server": "vllm",
     "hf": "hf",
     "olmo_core": "olmo_core",
+    "olmo_core_vlm": "olmo_core",
```
undfined, line 197: Since `olmo_core_vlm` uses the `olmo_core` optional dependency here, package overrides need the same normalization as `olmo_core`.

Currently, with `provider.kind=olmo_core_vlm`:

- `provider.package=2.3.0` remains the invalid raw requirement `2.3.0`.
- Source URLs do not receive the `ai2-olmo-core[torchao,transformers]` binding.
- The bundled `olmo_core` extra is retained alongside the override.

Please handle both provider kinds in `normalize_provider_package_for_kind`:

```python
if provider_kind in {"olmo_core", "olmo_core_vlm"}:
    return _normalize_olmo_core_package(package)
```

and add `"olmo_core_vlm": "olmo_core"` to `provider_extra_override`. The existing OLMo-core package tests can be parameterized over both kinds.

src/olmo_eval/inference/providers/olmo_core_vlm/provider.py
```diff
@@ -0,0 +1,873 @@
...
+        index out of bounds for those IDs. Cached because the weights are frozen for
+        the provider's lifetime and the embedding runs once per decode step.
+        """
+        import torch
+
+        if self._embedding_weight_cache is None:
+            embeddings = self.model.lm.embeddings
+            extra = getattr(embeddings, "extra_weight", None)
+            self._embedding_weight_cache = (
+                embeddings.weight if extra is None else torch.cat([embeddings.weight, extra], dim=0)
```
undfined: When extra vocabulary exists, this caches a concatenated copy of the entire embedding table. For a 4096-wide fp32 embedding at the 8B vocabulary size, that is roughly 2.5 GB of additional device memory.

A split lookup for base and extra IDs would avoid the copy. At minimum, `_embedding_weight_cache` should be cleared in `close()` before deleting the model.

src/olmo_eval/evals/tasks/common/base.py
```diff
@@ -172,6 +172,12 @@ class TaskConfig:
     #: beaker launcher mounts each as the user-scoped secret ``{user}_{NAME}``.
     required_secrets: tuple[str, ...] = ()
 
+    #: How a task that builds its prompt from a bare label should render it. The
+    #: pointing tasks read this; it follows the checkpoint, since instruction-tuned
+    #: and pretrain models were trained on different prompt forms.
+    prompt_templates: str | None = None
```
undfined: Nothing in this branch reads `prompt_templates` or `system_prompt_style`, and `TaskConfig.to_dict()` does not serialize them. I'd land these fields with the pointing-task consumer. If they need to remain in this split, they should be included in serialization and described in the PR as forward plumbing.

src/olmo_eval/inference/providers/config.py
```diff
@@ -61,7 +61,9 @@ class ProviderConfig:
     kwargs: Mapping[str, Any] = field(default_factory=dict)
 
     # Providers that require GPU resources for local inference
-    _GPU_PROVIDERS: ClassVar[frozenset[str]] = frozenset({"vllm", "vllm_server", "hf", "olmo_core"})
+    _GPU_PROVIDERS: ClassVar[frozenset[str]] = frozenset(
+        {"vllm", "vllm_server", "hf", "olmo_core", "olmo_core_vlm"}
```
undfined, line 65: There's another provider-kind allowlist in `runners/asynq/batching/config.py`: `olmo_core_vlm` should be included in `_SEQUENTIAL_ONLY`. Otherwise a streaming configuration passes validation but ultimately serializes behind the provider's `_model_lock`.

src/olmo_eval/inference/providers/olmo_core_vlm/cache.py
```diff
@@ -0,0 +1,318 @@
...
+                raise RuntimeError(f"KV cache overflow: {total} > allocated {k_cache.shape[1]}")
+            k_cache[:, pos:total] = k.to(k_cache.dtype)
+            v_cache[:, pos:total] = v.to(v_cache.dtype)
+            k_full = k_cache[:, :total].to(q.dtype)
+            v_full = v_cache[:, :total].to(q.dtype)
+
+            has_leftpad = getattr(kv_cache_manager, "_has_leftpad", False)
+            attn_mask: torch.Tensor | None = None
+            if seq_len > 1 or or_mask is not None or and_mask is not None or has_leftpad:
+                base = torch.ones(seq_len, total, device=q.device, dtype=torch.bool).tril(
```
undfined: This mask construction is one of the most correctness-sensitive parts of the provider, and an error here could produce plausible but incorrect generations. Could it be extracted into a module-level helper and tested directly for causal prefill, image `or_mask`, left padding, and decode? That would cover the logic without requiring an OLMo-core model.

undfined (changes requested):
Some suggestions before merging, but overall looks great!

src/olmo_eval/inference/providers/olmo_core_vlm/checkpoint.py
jason718: Fixed in c6d6a99 — sys.modules is consulted first and find_spec catches ValueError as you suggested. Regression test calls the helper twice (TestMmOlmoUnpickleShim).

src/olmo_eval/inference/providers/olmo_core_vlm/provider.py
jason718: Fixed in c6d6a99, mirroring OlmoCoreProvider: limit resolves from request.max_length or self.max_length (ValueError on <=0), empty continuations return an empty zero-logprob result, over-limit continuations raise before any tensor work, and truncation happens after the checks. Tests cover empty / at-limit / over-limit / per-request limit / non-positive (TestVLMLogprobsBoundaries).

src/olmo_eval/common/constants/infrastructure.py
jason718: Fixed in c6d6a99 — normalize_provider_package_for_kind handles both kinds, and olmo_core_vlm maps to the olmo_core extra in provider_extra_override, so version shorthand, git-URL binding, and bundled-extra replacement all behave identically. The shorthand test is parameterized over both kinds and now covers the git-URL form too.

src/olmo_eval/inference/providers/olmo_core_vlm/provider.py
jason718: Fixed in c6d6a99 with the split lookup: base IDs go to embeddings.weight and extra IDs to extra_weight in place, so no concatenated copy exists at all (the cache attribute is gone rather than cleared). Equivalence against a concatenated-table lookup is tested (TestVLMEmbedSplitVocab).

src/olmo_eval/evals/tasks/common/base.py
jason718: Agreed — the fields are out of this PR (c6d6a99) and land with their first consumer in the framework/captioning PR of the stack, where they are also serialized in TaskConfig.to_dict.

src/olmo_eval/inference/providers/config.py
jason718: Fixed in c6d6a99 — olmo_core_vlm joins _SEQUENTIAL_ONLY, so a streaming config fails validation up front. Good catch: it would indeed have serialized behind the provider's _model_lock.

src/olmo_eval/inference/providers/olmo_core_vlm/cache.py
jason718: Extracted in c6d6a99 as build_decode_attention_mask, a module-level helper taking (seq_len, pos, total, or_mask, and_mask, cache_leftpad). Direct tests cover causal prefill, the plain decode step (no mask), or_mask reopening and its left-alignment onto cached keys, and the left-padding semantics — real queries never see pad keys, pad-slot queries keep their causal rows (TestBuildDecodeAttentionMask).

jason718:
@abhishekraok @gregorybchris @AkshitaB Please help me review this.

abhishekraok (commented):
Two issues found in the multimodal inference paths. Validation: 74 targeted CPU tests passed; full GPU generation was not tested.

src/olmo_eval/inference/providers/olmo_core_vlm/checkpoint.py
```diff
@@ -0,0 +1,478 @@
...
+    :param image_patch_token_id: The ``<im_patch>`` vocab ID resolved from the
+        tokenizer; only used for the mm_olmo format (OLMo-core configs carry
+        their own).
+    :param attention_backend: Normalize every attention config to this backend
+        (``None`` keeps the checkpoint's). Training configs often pin ``flex``
+        or ``flash_2``, but eval needs the dense ``torch`` backend: it is the
+        one that supports the bidirectional image-token ``or_mask`` AND the
+        KV-cached decoding installed by :func:`enable_kv_cache`.
+    """
+    from olmo_core.nn.vision.multimodal import MultimodalLMConfig
```
abhishekraok, line 189: [P1] Pin an OLMo-core version that includes the vision modules

This imports `olmo_core.nn.vision`, but the `olmo_core` extra still pins `ai2-olmo-core[torchao,transformers]==2.4.0`. I inspected the published `ai2_olmo_core-2.4.0-py3-none-any.whl`: it contains no `olmo_core/nn/vision` package. Since `olmo_core_vlm` maps to this extra, standard installations and Beaker jobs fail during provider initialization unless users supply a package override. Please pin a compatible release or source revision so the declared installation supports the new provider.

src/olmo_eval/inference/providers/huggingface.py
```diff
@@ -149,6 +455,95 @@ def _truncate_at_stop(
...
 
+    def _build_chat_messages(self, request: LMRequest) -> list[dict[str, Any]]:
+        """Build processor chat messages, injecting images into the user turn.
+
+        Mirrors the released Molmo2 HF eval: a single user turn whose content is
+        the image(s) followed by the question text. A system message (if present)
+        is preserved as a text-only turn; images attach to the first user turn.
+        """
+        image_parts = [{"type": "image", "image": img} for img in (request.images or ())]
+        messages = request.messages or ({"role": "user", "content": request.prompt},)
```
abhishekraok: [P2] Preserve completion prompt text when the harness injects a system message

For a completion request, the question lives in `request.prompt` and `request.messages` is initially empty. When `HarnessConfig.system_prompt` is set, `Harness._apply_config()` populates `messages` with the system message. This expression then selects that tuple and discards the question. I reproduced this with `prompt="How many people?"` and one image: the resulting chat contains an image-only user turn followed by the system message, with no question text. Without an image it contains only the system message.

The same expression in `OlmoCoreVLMProvider._chat_text` (provider.py:256) has the same behavior. Please preserve the completion prompt as a user turn in both paths when the existing messages contain only the injected system message.

src/olmo_eval/inference/providers/olmo_core_vlm/checkpoint.py
jason718: Fixed in 5214e90, though the story turned out messier than a version bump: 2.6.0 does ship olmo_core.nn.vision, but it rejects vision-branch training configs ('flex' is not a valid AttentionBackendName; TransformerBlockConfig has no masked_dropout) and lacks nn.vision.molmo2_loader — I reproduced all three against the published wheel and a real checkpoint. So the provider now has its own olmo_core_vlm extra pinned to a vision-branch commit (ai2-olmo-core @ git+...OLMo-core.git@2dad929a), the text provider's olmo_core==2.4.0 pin is untouched, the two extras are declared conflicting, and BACKEND_OPTIONAL_GROUPS/beaker map the provider kind to the new extra — standard installations and Beaker jobs now work with no override. Also hardened: backend values are rewritten in the raw config before enum coercion, and the molmo2_loader imports raise an actionable message under an incompatible install. Verified with a GPU smoke through the declared extra: checkpoint load, image pointing, text generation. The pin collapses back into olmo_core once a release includes the multimodal classes.

src/olmo_eval/inference/providers/huggingface.py
jason718: Fixed in 5214e90 — both providers now build their chat turns via a shared chat_messages_for_request helper that appends request.prompt as a user turn whenever the existing messages contain no user turn, so a harness-injected system message no longer displaces the question (and an explicit user turn still takes precedence over prompt). Tests cover bare prompt, injected-system + prompt, existing user turn, and the empty request (TestChatMessagesForRequest); the GPU smoke exercised the completion path end to end.

gregorybchris (changes requested):

Looks great overall, and thank you for the thorough responses to the earlier round. I verified every one of Tyler's seven findings and Abhishek's two against the current branch: the `find_spec` guard, the logprob boundary handling, the split embedding lookup, both allowlists, the package normalization, the extracted mask helper, the `olmo_core_vlm` extra, and the `chat_messages_for_request` helper are all present with the tests described. Ruff, `ty`, and the unit suite (2193 passed) are clean on my machine.

I found two issues that I think should be fixed before merge, both in the "infrastructure failure becomes a score" class, plus a lockfile change worth confirming. The rest are non-blocking.

### Blocking

1. `src/olmo_eval/common/types/base.py:234`
   `LMRequest.images` is a shared contract, but only the two multimodal providers read it. `vllm`, `vllm_server`, `litellm`, `olmo_core`, and `mock` ignore the field, so a vision task run against a text provider produces text-only completions and scores them as a normal run. I reproduced this with `MockProvider`: a CHAT request carrying `images=("/nonexistent/cat.png",)` returns a canned answer with no error. Once the tasks in #317 onward land, `-m <text model> -t molmo2_imageqa` reports Success with meaningless numbers. Please make providers that do not support images raise a clear `ValueError` when `request.images` is set (a check in `InferenceProvider` keyed on a `supports_images` attribute would cover all of them), and add a test. Rejecting at planning time would be even better, since the run then fails before any model loads.

2. `src/olmo_eval/inference/providers/olmo_core_vlm/provider.py:692`
   A prompt that exceeds `max_length` is skipped with `metadata["skipped"] = "prompt_too_long"` and an empty output. The runner scores that empty output like any other, so a long-prompt instance becomes a zero rather than a recorded failure, and the hard-failure gate never sees it. The text `OlmoCoreProvider` has no equivalent silent path. Could this raise instead, so the runner records the instance in `scoring_errors` and the failure rate reflects it? If skipping is the intended contract, the runner needs to count `skipped` outputs as failures, and that belongs in the same change.

### Non-blocking

3. `uv.lock`
   The lock now carries two `torchao` entries, 0.15.0 and 0.17.0, where `main` has only 0.17.0. The 0.15.0 entry appears to have come in with the git-pinned OLMo-core resolution. Please confirm which extras resolve to which version, in particular that the text `olmo_core` extra still gets 0.17.0, and note it in the description. If the text path moved to 0.15.0, that is a core numeric dependency change and deserves the small parity canary we ask for on dependency bumps.

4. `src/olmo_eval/inference/__init__.py:135`
   `kind=olmo_core` now sniffs the checkpoint and silently constructs `OlmoCoreVLMProvider`. On Beaker, `BACKEND_OPTIONAL_GROUPS["olmo_core"]` installs the 2.4.0 extra, which has no `olmo_core.nn.vision`, and the two extras are declared conflicting, so this route can only fail at worker init there. Locally it works only when the user happened to install the VLM extra. It also makes the provider class depend on on-disk content rather than declared config. I would drop the sniffing and raise a clear error that names `kind=olmo_core_vlm`, or at minimum log at WARNING when the reroute happens.

5. `src/olmo_eval/inference/providers/config.py:242`
   `ProviderConfig.from_dict` now forwards every unknown key into `kwargs` for every provider kind, not just the multimodal ones. Previously a typo in a YAML provider block was dropped; now it reaches the constructor and either raises there (vLLM) or is absorbed by `**kwargs` (HF passes it to `from_pretrained`). That is arguably better, but it is a behavior change for all providers with no test. Please add a `from_dict` test and mention the change in the description.

6. `src/olmo_eval/runners/asynq/monitoring.py:20`
   `OLMO_EVAL_PROVIDER_INIT_TIMEOUT` is parsed at import time with a bare `float()`, so a bad value fails on import of the monitoring module far from the cause, and the variable is not documented anywhere in the stack, including the README section in #324. Could it be read inside `wait_for_provider_init` with a clear error, and documented?

7. `src/olmo_eval/inference/providers/olmo_core_vlm/provider.py:818`
   The logprob entries here carry only `token` and `logprob`. `HuggingFaceProvider` and the text `OlmoCoreProvider` also emit `token_id` and `bytes`, and `ngram_copying` reads `bytes`. Could this match the other providers?

8. `src/olmo_eval/inference/providers/huggingface.py:539` and `olmo_core_vlm/provider.py:652`
   Both multimodal paths `.strip()` the generated text; the HF text path does not. If the strip mirrors mm_olmo, a one-line comment saying so would help. If not, it changes what the scorers see.

9. `src/olmo_eval/inference/providers/huggingface.py:96`
   `_patch_processor_optional_attribute_kwargs` replaces `ProcessorMixin.__init__` process-wide, so it affects any other processor loaded in the same process. It is idempotent and guarded, which is good. Worth a sentence in the docstring about the blast radius.

10. Branch base
    The stack is based on `254e2c22`, four commits behind `main`, and the "Rebase on main; regenerate the lock" commit did not actually move the base. This PR merges cleanly, but the lock was regenerated against the old base, so please rebase before merging.

### Nits

- Nit: `checkpoint.py:42` has a dangling comment ("Default tokenizer for Molmo2-family models...") with no constant under it.
- Nit: `checkpoint.py:261` and `:425` tell users to pin `provider.package=...OLMo-core.git@vision`; that is superseded by the `olmo_core_vlm` extra this PR adds.
- Nit: `preprocessing.py:17` has an empty `if TYPE_CHECKING: pass`.
- Nit: `infrastructure.py:129` adds `ai2/holmes` to `WEKA_CLUSTERS`, which is unrelated to this PR.
- Nit: the description still lists `TaskConfig.prompt_templates / system_prompt_style` under core plumbing; those moved to #317.

### Before merge

Items 1 and 2, a confirmation on item 3, and a rebase onto current `main`. After that, feel free to merge once CI is green. Items 4 through 10 can be follow-ups if you prefer; if so, please file an issue for item 4 so the `olmo_core` auto-route does not get forgotten.

jason718:
All seven items addressed in f411948 (blocking) / description (notes):

**1 (blocking)** — providers now declare `supports_images` and the harness rejects image requests aimed at a provider that cannot render them, with an error naming `provider.kind=olmo_core_vlm`. HF sets it from its multimodal mode, the VLM provider sets it, and the mock opts in deliberately so vision suites stay dry-runnable. Tests: `TestImageRequestRejection` (reject / pass / text-unaffected).

**2 (blocking)** — the `prompt_too_long` silent skip is gone: a multimodal prompt exceeding `max_length` now raises with the lengths named, mirroring the text provider's budget error, so the runner records it and the failure gate sees it. (Multimodal prompts can't be truncated without corrupting the image-token layout, so failing is the only honest option.)

**3** — good catch, and it was actually worse than two entries: uv had unified both conflict forks on the git pin's `torchao==0.15.0`, silently downgrading the *text* extra from main's 0.17.0. The `olmo_core` extra now pins `torchao==0.17.0`; verified per extra via `uv export` (0.17.0 / 0.15.0) and noted in the description.

**4** — the reroute now logs a WARNING naming the explicit kind, and the README steers to `provider.kind=olmo_core_vlm` first, with the Beaker failure mode spelled out. Kept the detection itself: it is load-bearing for existing local workflows.

**5** — behavior kept (a typo surfacing in the provider beats silent dropping), now tested (`TestProviderConfigKwargsPassthrough`) and noted in the description.

**6** — `OLMO_EVAL_PROVIDER_INIT_TIMEOUT` is read at use time with an error that names the variable, and documented in the README's multimodal section (this stack's PR 5).

**7** — VLM logprob entries now carry `token_id` and `bytes` like the other providers.

Full suite at the stack tip: 2,419 passed; lint/format/`ty` clean.

gregorybchris (approved):

PR #317, opened 2026-08-25
Vision task framework + captioning benchmarks (2/5)

jason718:
## Description

Second of five PRs splitting #278 (after #316, now merged). **Milestone: the `evals/vision/` package exists and its first family — PixMo-Cap dense captioning, GPT-judge scored — runs end to end.** Based directly on `main`.

### The `evals/vision/` package

```
evals/vision/
├── tasks/base.py       # VisionTask — consolidated instance caching + image-attached requests
├── scoring/
│   ├── prompts.py           # mm_olmo prompt families: 43 seeded pointing templates, style prefixes
│   ├── prompt_templates.py  # vendored mm_olmo question templates (caption, count, ...)
│   └── judges.py            # dense-caption GPT judge (charxiv + math_vista join in #323)
├── benchmarks/dense_caption.py   # `dense_caption`, `dense_caption_captioner`
└── data/
    ├── images.py       # lazy image references (picklable across worker processes)
│   └── paths.py        # $MOLMO_DATA_DIR resolution, fail-fast when missing (path rebasing lands with its first consumer in #322)
```

Registration mirrors `evals/tasks`: one `from . import vision` line in `evals/__init__.py`, explicit benchmark imports inside the package.

**Two amendments to the reviewed layout sketch, for sign-off:**

1. `benchmarks/` holds **flat modules, not per-benchmark directories** — 17 of the 19 benchmarks are single files, and shared parsers (`mmmu_parsing` serves mmmu, mmmu_pro *and* multi-image) belong in `scoring/`.
2. Vision scorers are **not re-exported through `common/scorers/__init__.py`** — that module is imported by `evals/tasks/common/base.py` (every task in the repo), so an eager re-export from `evals.vision.scoring.*` creates a package cycle.

### `VisionTask`

`ImageQATask`, `PointingTask`, `MultiImageQATask` and `DenseCaptionEval` on the vision branch carried byte-identical instance/request mechanics; `VisionTask` consolidates them. Subclasses implement `_build_instances` and optionally `_attach_images`, which returns **lazy references** (paths / picklable callables) so the parent process never decodes pixels; providers resolve them at inference time. No runner code is touched. `config.limit` is left to the runner's seeded random sample, exactly as for text tasks (dense caption applies its own index-stable limit while building, because its seeded prompt depends on the raw line index).

### Task identity + config

- `TaskConfig.prompt_templates` / `system_prompt_style` land here with their first consumer, serialized in `to_dict` (with a test that changing either changes the task hash); the `-o` allowlist picks them up via main's derived mechanism (#367)
- `DenseCaptionJudgeScorer.to_dict` emits only `type`/`name`/`model` — cache location and mode are machine-local and no longer perturb the task hash (test: hash is identical with `DENSE_CAPTION_EVAL_DIR` redirected)
- `required_secrets=("OPENAI_API_KEY",)` on both captioning tasks — beaker never mounted the judge's key before. The launcher resolves it as the user-scoped secret `{beaker-username}_OPENAI_API_KEY` (existing convention; the launch error names the exact `beaker secret write` command when it is missing)
- Unknown `system_prompt_style` values raise; `num_samples > 1` is rejected (the judge protocol is defined for one sample)

### Dense-caption metrics + judge

- One parameterized `DenseCaptionMetric` whose `compute_instance` stores its own field (`None` when that judge side is invalid); the `avg` primary has no exact per-instance decomposition and stores nothing. Regression tests poison the scorer channel.
- Legacy 0–100 scale kept, display format pinned to `raw` (the viewer no longer renders 40.0 as 4000%)
- GPT/cache failures propagate to the runner's structured `scoring_errors` instead of silently shrinking the valid-example denominator; only unparseable judge replies are per-example invalid
- Two mm_olmo parity decisions are documented in place as deliberate: the end-anchored `Not Stated` regex and scoring only the judge lines returned — both are how every published Molmo2 dense-caption number was produced
- Judge cache follows `DENSE_CAPTION_EVAL_DIR` (`$DENSE_CAPTION_EVAL_DIR/gpt4-cache`)

### Ai2 data paths

The Weka defaults behind `MOLMO_DATA_DIR` / `DENSE_CAPTION_EVAL_DIR` remain for now so Ai2 launches need no extra flags, but a missing root now **fails fast with the variable named** rather than reading a nonexistent path. Their removal per the #153/#276 convention is tracked in #381 (linked from the module docstrings).

### Verification

- 2,730 instances, matching the mm_olmo-validated runs; the full stack later ran this task end to end on Beaker (2,730/2,730, zero failures)
- Judge unit tests; prompt-family tests include an mm_olmo parity guard (`MM_OLMO_ROOT`, skipped when absent)
- Full suite green; lint/format/`ty` clean

## Review history

Tyler's 10 findings (per-instance metric storage, exception swallowing, cache dir, task hash, lazy images, `Not Stated` parity, partial-response parity, display scale, multi-sample aggregation, style validation) and the review agent's 9 (scorer `to_dict` vs hash, Weka defaults, runner-side limit, `load_instance_image` delegation, `pixmo_cap` alias, allowlist rebase, the untested `--secret-env` launcher tweak — removed rather than tested, stale comment, public style-set names) are all addressed in the current head.

## Type of Change

## Checklist

- [x] Any dependent changes have been merged and published (#316)

Part of #278 (tracking). Next: #322 — pointing + counting.


src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+
+_DEFAULT_EVAL_DIR = "/weka/oe-training-default/mm-olmo/dense_caption_eval"
+
+# Shared scorer instance — all 5 metrics hold a reference so _get_scorers()
+# deduplicates to a single GPT-judge call per example (via Scorer.__call__).
+_JUDGE = DenseCaptionJudgeScorer()
+
+
+# ---------------------------------------------------------------------------
+# Metric definitions
```
undfined, line 48: These metric classes inherit Metric.compute_instance(), whose default falls back to the shared scorer channel. DenseCaptionJudgeScorer returns recall, so build_predictions() currently stores that recall value under every per-instance metric—including consistency, recall-at-10, number of statements, and avg. I reproduced all five fields being written as 0.4 when the actual values differed. This corrupts prediction exports and pairwise analysis even though the task-level aggregates are correct. Please implement compute_instance() for each metric from dense_caption_result, preserving the aggregate’s scaling, or disable scorer fallback where there is no exact per-instance equivalent.

src/olmo_eval/evals/vision/scoring/judges.py
```diff
@@ -0,0 +1,364 @@
...
+                num_covered, num_statements = parse_recall_output(raw)
+                recall_valid = num_statements > 0
+                result["recall"] = num_covered / num_statements if recall_valid else 0.0
+                result["recall_at_10"] = (
+                    min(num_covered, 10) / min(num_statements, 10) if recall_valid else 0.0
+                )
+                result["num_statements"] = num_statements
+                result["num_covered"] = num_covered
+                result["recall_valid"] = recall_valid
+            except Exception as exc:
```
undfined: Catching every exception here prevents the task runner’s ContextScorer wrapper from recording a structured scoring_errors entry. Authentication, rate-limit, cache-I/O, and programming failures instead become recall_valid=False, after which the metrics silently remove the affected examples from their denominators; a complete outage produces an ordinary all-zero result. Please let unexpected exceptions propagate and only handle explicitly recognized malformed judge responses locally. The same applies to the consistency block below.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+from olmo_eval.evals.vision.scoring.prompt_templates import dense_caption_question
+from olmo_eval.evals.vision.tasks.base import VisionTask
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_EVAL_DIR = "/weka/oe-training-default/mm-olmo/dense_caption_eval"
+
+# Shared scorer instance — all 5 metrics hold a reference so _get_scorers()
+# deduplicates to a single GPT-judge call per example (via Scorer.__call__).
+_JUDGE = DenseCaptionJudgeScorer()
```
undfined, line 44: DENSE_CAPTION_EVAL_DIR is documented as the root containing gpt4-cache, but it only controls the reference-data loader. The scorer instantiated here still uses /weka/oe-training-default/mm-olmo/dense_caption_eval/gpt4-cache; I verified that setting DENSE_CAPTION_EVAL_DIR leaves scorer.cache_dir unchanged. A non-Weka run can therefore load its data correctly but fail when writing the judge cache. Please derive the cache path from the same root, or add a separate documented cache-directory override.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+        a checkpoint a natural-language instruction instead measures it out of distribution:
+        on a 4B stage-1 run it cost 7.1 recall points (38.18 -> 45.28) and made 3.4% of
+        answers come back as `<points .../>` markup rather than prose.
+
+        So `-o system_prompt_style=style_and_length_v2` selects the captioner prompt here,
+        the same override that selects it for the pointing tasks.
+        """
+        if self.caption_prompt is not None:
+            return self.caption_prompt
+        style = self.config.system_prompt_style or self.default_system_prompt_style
```
undfined, line 203: I think I touched on this in the previous change set but just reiterating here; this makes system_prompt_style output-affecting, but TaskConfig.to_dict() does not serialize either system_prompt_style or prompt_templates. I verified that the natural-language and long_caption 65: configurations produce the same task hash and that the stored config contains neither field. Runs using different prompts can therefore collide in artifact naming and result storage. Please serialize both fields and add a test asserting that changing the prompt style changes the task hash.

src/olmo_eval/evals/vision/tasks/base.py
```diff
@@ -0,0 +1,53 @@
...
+        yield from self._instances_cache
+
+    @abstractmethod
+    def _build_instances(self) -> Iterator[Instance]:
+        """Yield all instances for ``self.config.split`` (before ``limit``)."""
+        ...
+
+    def _attach_images(self, instance: Instance) -> tuple[Any, ...] | None:
+        """The images to send with ``instance``; ``None`` sends a text-only request."""
+        image = load_instance_image(instance)
```
undfined: prepare_task_items() formats every instance up front, so this opens every image before the workers start. Those QueueItems remain retained by the parent, and pickling a PIL image for the multiprocessing queue forces its pixels to be decoded; the request is then sent back again in ResultItem. For the 2,730-image caption task this can retain multiple uncompressed copies of the dataset and keep every file descriptor open until serialization. Please keep a picklable path/lazy reference through the queue and resolve it in the worker/provider, and avoid returning the image payload with the completed request when it is no longer needed.

src/olmo_eval/evals/vision/scoring/judges.py
```diff
@@ -0,0 +1,364 @@
...
+def parse_recall_output(text: str) -> tuple[int, int]:
+    """Parse GPT stated/not-stated output.
+
+    Returns (num_covered, num_statements) counting only unambiguous lines.
+    Mirrors eval_recall() lines 323–346 in gpt_dense_caption_eval.py.
+    """
+    lines = [x.strip() for x in text.split("\n") if x.strip()]
+    valid_scores: list[bool] = []
+    for line in lines:
+        if re.fullmatch(r".*\bnot st[a-z]+$", line, flags=re.IGNORECASE):
```
undfined, line 216: The negative pattern only matches when the label ends the line, so common variants such as 1. Not Stated. and 1. Not Stated (missing) fall through to the broad " stated" check and are counted as covered. The current test explicitly preserves this misclassification. Unless incorrect legacy scores must be reproduced exactly, detect Not Stated before allowing trailing punctuation or explanation text, and update the test to expect it to be uncovered.

src/olmo_eval/evals/vision/scoring/judges.py
```diff
@@ -0,0 +1,364 @@
...
+            try:
+                raw = await _cached_gpt_call(
+                    _recall_prompt(mturk_statements, caption),
+                    model=self.model,
+                    cache_dir=self.cache_dir,
+                    cache_only=self.cache_only,
+                    recompute=self.recompute,
+                )
+                num_covered, num_statements = parse_recall_output(raw)
+                recall_valid = num_statements > 0
```
undfined: This considers any response with at least one parsed label valid, even when the judge omitted most of the requested labels. For example, a four-statement prompt followed by only `1. Stated` is recorded as valid recall `1.0`; the consistency path does the same thing. Please require the parsed label count to match the number of statements supplied to the judge, and treat partial responses as invalid or retryable.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+    def compute(self, responses: Sequence[Response]) -> float:
+        vals = [
+            o.metadata["dense_caption_result"]["recall"]
+            for r in responses
+            for o in r.outputs
+            if o.metadata
+            and "dense_caption_result" in o.metadata
+            and o.metadata["dense_caption_result"].get("recall_valid")
+        ]
+        return (sum(vals) / len(vals) * 100) if vals else 0.0
```
undfined: This returns recall on a 0–100 scale, but `Metric.pairwise_display_format()` classifies the name `recall` as a 0–1 percentage, and the results viewer multiplies it by 100. A score of `40.0` therefore renders as `4000.0%`. Please either keep the metric on the framework’s 0–1 scale or override its display format and unit for the legacy 0–100 representation.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+    """Mean recall (×100) over valid examples."""
+
+    name: str = "recall"
+    scorer: type[Scorer] | Scorer = _JUDGE
+
+    def compute(self, responses: Sequence[Response]) -> float:
+        vals = [
+            o.metadata["dense_caption_result"]["recall"]
+            for r in responses
+            for o in r.outputs
```
undfined: These comprehensions flatten every output, so a `num_samples > 1` run averages the samples even though `TaskConfig.output_score_aggregation` defaults to `MAX`. For outputs with recall `0.2` and `0.8`, task scoring selects `0.8` while this metric reports `50.0`. Please apply the configured per-response aggregation before averaging responses, or reject multi-sample runs for this task.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
```diff
@@ -0,0 +1,306 @@
...
+        on a 4B stage-1 run it cost 7.1 recall points (38.18 -> 45.28) and made 3.4% of
+        answers come back as `<points .../>` markup rather than prose.
+
+        So `-o system_prompt_style=style_and_length_v2` selects the captioner prompt here,
+        the same override that selects it for the pointing tasks.
+        """
+        if self.caption_prompt is not None:
+            return self.caption_prompt
+        style = self.config.system_prompt_style or self.default_system_prompt_style
+        if style in ("style_and_length", "style_and_length_v2"):
```
undfined: Anything other than the two `style_and_length` values falls through to the natural-language prompt, so a typo in `system_prompt_style` silently evaluates the wrong prompt family. Since this setting materially changes the score, please explicitly handle the supported demo-style values and raise `ValueError` for anything else, as the pointing prompt helper already does.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Fixed in 2f9a116 — the five metric classes are replaced by a parameterized DenseCaptionMetric whose compute_instance returns its own field on the aggregate's scale (None for examples that side of the judge marked invalid), and the aggregate is now literally the mean of the per-instance values. The avg primary has no exact per-instance decomposition (its two means run over different valid subsets), so it stores None and disables scorer fallback rather than approximating. Regression tests poison the scorer channel and assert every metric ignores it (TestDenseCaptionPerInstanceMetrics).

src/olmo_eval/evals/vision/scoring/judges.py
jason718: Fixed in 2f9a116 — both try/except blocks are gone entirely. It turned out the parse ladder never raises (garbage parses to zero counts → recall_valid/consistency_valid=False, matching mm_olmo), so the catches only ever swallowed GPT/cache failures — exactly the ones that should reach the runner's structured scoring_errors. Tests: a raising _cached_gpt_call now propagates (test_scorer_propagates_gpt_errors); an unparseable response is a per-example invalid result (test_unparseable_judge_response_marks_example_invalid).

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Fixed in 2f9a116 — the scorer's cache_dir default is now default_dense_caption_cache_dir(): $DENSE_CAPTION_EVAL_DIR/gpt4-cache when the reference-data root is overridden, else the shared weka cache. One env var moves both the data and the cache.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Both fields are serialized in TaskConfig.to_dict (this landed when the fields moved into this PR per your #316 comment — base.py:292-293), and 2f9a116 adds the test you asked for: TestPromptFamilyTaskIdentity asserts compute_task_hash changes when either system_prompt_style or prompt_templates changes.

src/olmo_eval/evals/vision/tasks/base.py
jason718: Fixed in 2f9a116 (single-image; d231b18 completes multi-image) — requests now carry lazy references: VisionTask._attach_images attaches the path string or the picklable module-level partial from instance.metadata, never a decoded PIL; providers resolve entries via the new olmo_eval.common.images.resolve_images right before preprocessing; and the runner strips the images tuple from the request echoed back in ResultItem (processing._result_request). The multi-image family attaches one lazy capped-list loader. Tests: TestLazyImageRequests / TestLazyMultiImageRequests (format_request attaches a nonexistent path without opening it; resolution opens paths, calls callables, flattens and caps lists; the echoed request is stripped).

src/olmo_eval/evals/vision/scoring/judges.py
jason718: We'd like to keep this one byte-exact to mm_olmo: the negative match in vixmo_caption_utils.py is the same end-anchored regex, so '1. Not Stated.' falls through to the positive branch upstream too. Every published Molmo2 dense-caption number was produced with this behavior, and the whole value of this port is comparability against those stored results. 2f9a116 documents the spot as vendored-verbatim (with the mm_olmo reference) so the choice is explicit rather than accidental; happy to revisit as a versioned parser if mm_olmo ever fixes it.

src/olmo_eval/evals/vision/scoring/judges.py
jason718: Same parity rationale as the 'Not Stated' thread: mm_olmo scores exactly the lines the judge returned (range(min(n, len(lines))) in vixmo_caption_utils.py), so requiring the label count to match would diverge from every stored dense-caption result. The fully-empty case is already recorded as invalid (recall_valid=False), and after the exception-handling fix a transport failure now propagates instead of masquerading as a partial response. Documented as vendored-verbatim in 2f9a116.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Fixed in 2f9a116 — the metrics keep mm_olmo's legacy 0-100 scale but pin pairwise_display_format() to 'raw' (and disable scorer fallback), so the viewer no longer multiplies by 100. Covered by test_legacy_scale_displays_as_raw.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Fixed in 2f9a116 by rejecting the configuration: dense_caption raises ValueError for num_samples > 1 before building instances. The judge protocol (and mm_olmo) is defined for one sample per example; averaging or maxing judged samples would both be inventions, so failing fast seemed more honest than picking one. Test: test_multi_sample_runs_rejected.

src/olmo_eval/evals/vision/benchmarks/dense_caption.py
jason718: Fixed in 2f9a116 — _question now validates against the same style sets the pointing helper uses and raises ValueError for anything else. Test: test_unknown_system_prompt_style_raises (with a deliberate typo).

gregorybchris (changes requested):

Looking good! I verified all ten of Tyler's earlier findings against the branch: the parameterized `DenseCaptionMetric` with its own `compute_instance` and disabled fallback, the removed `try/except` in the judge, the cache directory derived from `DENSE_CAPTION_EVAL_DIR`, both prompt fields in `TaskConfig.to_dict` with the hash test, lazy image references plus `_result_request`, the `ValueError` on unknown styles, the `num_samples > 1` rejection, and the `raw` display format. The two parity-preserving decisions on the "Not Stated" regex and partial judge responses are documented in place, which matches how this repo handles bug-compatible ports. Ruff, `ty`, and the unit suite (2234 passed) are clean locally.

Three things I think should be addressed before merge, then some smaller items.

### Blocking

1. `src/olmo_eval/evals/vision/scoring/judges.py:286`
   `DenseCaptionJudgeScorer.cache_dir` is a dataclass field, and `Metric.to_dict` serializes scorer instances with `asdict`, so the cache path is part of the task hash. I reproduced it: `dense_caption` hashes to `b55f93d3...` by default and `7eeccc5e...` with `DENSE_CAPTION_EVAL_DIR=/tmp/x`. The same eval on Weka and on a laptop therefore gets different artifact names and different stored configs, while `cache_only` and `recompute` also leak into the hash. Only `model` is output-affecting. Please give the scorer a `to_dict` that emits `type`, `name`, and `model`, and add a test asserting the hash is independent of the cache settings. The same fix is needed for the CharXiv and MathVista scorers in #323, where the effect is worse.

2. `src/olmo_eval/evals/vision/data/paths.py:12`, `benchmarks/dense_caption.py:40`, `scoring/judges.py:27`
   Three Weka paths ship as source defaults. Tyler asked in #153 and #276 that runner and task code not carry Ai2 paths; the two options he accepted were a generic default that Ai2 users override at launch, or naming the task as internal and filing a follow-up. The cheapest version here: keep the env vars, drop the Weka literals to `None`, and raise a clear error naming `MOLMO_DATA_DIR` / `DENSE_CAPTION_EVAL_DIR` when they are unset and the data is not found. The Weka values then live in the README and the Beaker launch docs. If you would rather defer, please file the issue and link it from the module docstrings.

3. `src/olmo_eval/evals/vision/tasks/base.py:33`
   `VisionTask.instances` applies `limit` by taking the first N. Every other task leaves `limit` to the runner, which draws a seeded random sample (`runners/asynq/preparation.py:60`), so `-o limit=100` means a random sample for text tasks and the first hundred lines for all 25 vision tasks. The dense-caption justification (the raw line index drives the seeded prompt) is specific to that task, and it already applies its own limit. Could the base class drop the slice and let the runner sample, unless prefix-limit is intended for the whole family, in which case the divergence should be documented on `VisionTask`?

### Non-blocking

4. `src/olmo_eval/evals/vision/data/images.py:45`
   `load_instance_image` duplicates `common.images.resolve_image` and has no caller in this PR; its only consumer arrives in #323. Either land it with that consumer or have it delegate to `resolve_image`.

5. `src/olmo_eval/evals/vision/benchmarks/dense_caption.py:279`
   `register_variant("dense_caption", "pixmo_cap")` registers an alias with no overrides. Variants in this repo describe a format or metric modifier, so `dense_caption:pixmo_cap` reads as if it changes something. Suggest dropping it or documenting why the alias is needed.

6. `src/olmo_eval/cli/utils.py:130`
   This hand-maintained allowlist conflicts with `main`, which added `strip_thinking` in #361 (the stack is based four commits behind). #367 proposes deriving the list from `TaskConfig`, which would remove the need for this edit entirely. A rebase is needed regardless.

7. `src/olmo_eval/cli/beaker/launch.py:579`
   Letting an explicit `--secret-env` mapping satisfy a required secret is a reasonable change, but it is untested and unrelated to vision. Please add a small test for it.

8. `src/olmo_eval/evals/vision/benchmarks/dense_caption.py:171`
   The comment "A missing scorer dependency scores every instance zero rather than failing" is no longer true after the exception-handling fix; an `ImportError` now propagates. Please update it so the next reader does not rely on it.

9. `src/olmo_eval/evals/vision/scoring/prompts.py`
   `_NO_PREFIX_STYLES` and `_STYLE_PREFIX_STYLES` are underscore-private but imported by `dense_caption.py`. Either export them or give them public names.

### Nits

- Nit: `tests/evals/vision/test_prompts.py:62` hard-codes `/root/mm_olmo`; an env var (`MM_OLMO_ROOT`) would let others run the parity guard.
- Nit: `VisionTask.dependencies = ["pillow"]` and `DenseCaptionEval.dependencies = ["pillow", "openai"]` restate the base entry; consider extending rather than replacing.

### Before merge

Items 1 through 3 (or, for item 2, a filed follow-up issue and the fail-fast error), plus the rebase from item 6. Then merge once CI is green. Thanks for the careful work on the earlier round.

jason718:
All nine items addressed in 8586d999 (now based directly on `main` after #316 merged):

**1 (blocking)** — `DenseCaptionJudgeScorer.to_dict` now emits only `type`/`name`/`model`, so cache location and mode no longer enter the task hash. Test: `TestJudgeCacheSettingsOutsideTaskHash` asserts the hash is identical with `DENSE_CAPTION_EVAL_DIR` redirected and that two scorers with different cache settings serialize identically. The same `to_dict` treatment goes to the CharXiv and MathVista scorers in #323.

**2 (blocking)** — took the deferral path you allowed: the Weka defaults stay (Ai2 launches keep working without three `-e` flags), but a missing `MOLMO_DATA_DIR` / `DENSE_CAPTION_EVAL_DIR` root now fails fast with the variable named, so a non-Ai2 run never silently reads a nonexistent path. Filed #381 for the removal per #153/#276, linked from both module docstrings.

**3 (blocking)** — `VisionTask.instances` no longer head-slices; `config.limit` is left to the runner's seeded sample exactly as for text tasks. Dense caption keeps its own index-stable limit while building (the seeded prompt depends on the raw line index), which is now the only place the family touches `limit`. Test: `TestLimitLeftToRunner`.

**4** — `load_instance_image` now delegates to `common.images.resolve_image`. **5** — the no-op `pixmo_cap` alias is gone (no references anywhere in the stack). **6** — rebased; with #367's derived allowlist our `cli/utils.py` edits disappeared entirely. **7** — the `--secret-env` rule is a small `unmapped_required_secrets` helper with a test. **8** — stale comment replaced. **9** — `NO_PREFIX_STYLES` / `STYLE_PREFIX_STYLES` are public names now.

Nits: the parity guard reads `MM_OLMO_ROOT` (default unchanged); `DenseCaptionEval.dependencies` extends `VisionTask.dependencies`.

Tyler's ten threads on this PR were all answered with fixes in the previous round and are now marked resolved. Full suite 2,404 passed on this branch; lint/`ty` clean.

jason718:
Follow-up on item 7: rather than adding a test for the `--secret-env` launcher tweak, it's removed from this PR entirely (07fa709f rebuilds the branch without it) — it was unrelated to vision, and the existing convention already covers the need: task-required secrets resolve to the user-scoped `{beaker-username}_OPENAI_API_KEY`, and the launch error prints the exact `beaker secret write` command when it's missing. `src/olmo_eval/cli/` is now untouched by this PR.

jason718:
Two more trims in a5714026: `_result_request` (and its test) is gone — with lazy image references nothing heavy was being echoed through the result queue, so `runners/asynq/processing.py` is untouched by this PR again — and `rebase_data_path` moves to #322 with its first consumers. The runner and CLI are now byte-identical to `main` in this PR.

src/olmo_eval/evals/tasks/common/base.py
```diff
@@ -296,6 +302,8 @@ def serialize_primary_metric(pm: Any) -> Any:
             "max_length": self.max_length,
             "answer_extractor": getattr(self.answer_extractor, "__name__", None),
             "dependencies": self.dependencies,
+            "prompt_templates": self.prompt_templates,
+            "system_prompt_style": self.system_prompt_style,
         }
         # Emitted only when set so that task hashes of runs without it are
         # unchanged from before the field existed.
```
gregorybchris: Could we emit these only when set, alongside `strip_thinking` and the judge fields just below? As written, every task in the repo gets a new hash — `arc_easy` `8738c566...` → `5c188729...`, `gsm8k` `4f68bf90...` → `1a55be89...` — because `None` still lands in the hashed JSON.

I verified this exact replacement locally: all five hashes I checked return to their `main` values, `tests/evals/vision/` stays 42/42 (`TestPromptFamilyTaskIdentity` included, since both fields default to `None` even on `dense_caption`), and ruff is clean.

```suggestion
        }
        # Emitted only when set so that task hashes of runs without it are
        # unchanged from before the field existed.
        if self.prompt_templates is not None or self.system_prompt_style is not None:
            serialized["prompt_templates"] = self.prompt_templates
            serialized["system_prompt_style"] = self.system_prompt_style
```

gregorybchris (approved):

Approving — this addresses everything from the last round, and the trims make it a much easier PR to read. One suggestion inline that I'd like to land with it.

I re-verified all nine items rather than taking the summary on trust:

- **Judge cache out of the task hash.** `dense_caption` hashes to `0cd777dc53b9407e` under the default root, `DENSE_CAPTION_EVAL_DIR=/tmp/x`, and `/tmp/y`. It was `b55f93d3...` / `7eeccc5e...` before.
- **`limit` left to the runner**, so the vision tasks now get the same seeded sample as the text tasks.
- **Weka paths**: the fail-fast error names the variable, and #381 records the literals, the accepted options, and why the interim state is what it is. That's the deferral I offered, done properly.
- Items 4, 5, 8, 9 and both nits all check out, and dropping the `--secret-env` tweak and `_result_request` rather than testing them was the right call — `cli/` and `runners/` are byte-identical to `main` now.

Locally: 2385 passed, 25 skipped; ruff, format and `ty` clean. CI is green.

### Before merge

`src/olmo_eval/evals/tasks/common/base.py:305` — `to_dict` emits `prompt_templates` and `system_prompt_style` unconditionally, so `None` enters the hashed JSON for every task in the repo and every existing task hash changes. Against `main` at `35ecf49e`:

| task | `main` | this PR |
|---|---|---|
| `arc_easy` | `8738c56690bd51aa` | `5c1887295371d1a3` |
| `gsm8k` | `4f68bf90a407205a` | `1a55be898c78d3f1` |
| `hellaswag` | `0d3fbae973978310` | `d01e4aba2e1f2523` |
| `humaneval` | `e90fac5e5d166beb` | `b6e5ee3150759390` |
| `drop` | `61d76aaf976b7df2` | `ebc7997df6421be7` |

Every stored result stops matching new runs of the same config, and pairwise tooling can't line the two up. This is the convention the comment four lines below already describes for `strip_thinking` and the judge fields — "emitted only when set so that task hashes of runs without it are unchanged from before the field existed" — so it just needs the same treatment.

I applied the inline suggestion locally to check the cost: all five hashes return to their exact `main` values, and `tests/evals/vision/` still passes 42/42, including `TestPromptFamilyTaskIdentity`. Both fields default to `None` even on `dense_caption`, so setting either one still changes the hash, and nothing about the vision tasks' identity changes. Worth a small regression test asserting a text task's hash is unaffected by the new fields, since nothing currently guards that.

Approving so this isn't gated on me — please push the suggestion (or an equivalent) and merge once CI is green.

jason718:
Thanks — and good catch on the hash regression; that one would have been expensive to notice after the fact.

Pushed your suggestion in 5dd87aa9. Confirmed independently against `main` at `35ecf49e`: all five hashes are back to their exact `main` values.

| task | this PR (before) | now |
|---|---|---|
| `arc_easy` | `5c1887295371d1a3` | `8738c56690bd51aa` |
| `gsm8k` | `1a55be898c78d3f1` | `4f68bf90a407205a` |
| `hellaswag` | `d01e4aba2e1f2523` | `0d3fbae973978310` |
| `humaneval` | `b6e5ee3150759390` | `e90fac5e5d166beb` |
| `drop` | `ebc7997df6421be7` | `61d76aaf976b7df2` |

`dense_caption` identity still responds to both fields (both default to `None`, so `TestPromptFamilyTaskIdentity` keeps working). Added the guard you asked for as `TestPromptFamilyFieldsAreOptIn`: three text tasks must not carry either key in their serialized config, plus a companion assertion that opting in *does* add the key back and change the hash — so the test can't be satisfied by simply dropping the fields.

Vision suite 48 passed; the rest of the stack is re-rolled on top and I'll merge once CI is green.

Issue #310, opened 2026-08-22
vllm_server: max_tokens beyond model context makes every request fail, run still reports Success with 0.0


abhishekraok:
Found during a GPU parity run of the new reasoning-regime tasks (popqa:chat, omega_500; PRs #307/#308) against oe-eval.

**Repro:** `olmo-eval beaker launch -m Qwen/Qwen3-8B -t omega_500 -G 1 ...` — omega_500/popqa:chat register `max_tokens=131072` (mirroring oe-eval's post-training configs), but Qwen3-8B's `max_model_len` is 40960.

**What happens (experiment `01M0MYH78J79E15EX61PXMC1KD`):**
1. Every single request 400s: `max_tokens=131072 cannot be greater than max_model_len=max_total_tokens=40960`.
2. The dispatch layer logs each as a WARNING (`process_fn raised BadRequestError`), scores the empty outputs, and the run completes with exit 0 and a green `Success` row showing 0.0000 — 14,767 instances "processed" in 263s.

**Two asks:**
1. **Clamp** `max_tokens` to the model context in the vllm_server provider, the way oe-eval does (`Running with effective max_gen_toks of 65192 instead of 65536!` — it uses `max_model_len - longest_prompt`). The provider already knows `max_length` (configured or discovered from `/models`) and has a remote-tokenizer path, so per-request `min(params.max_tokens, max_length - prompt_tokens)` is feasible; even a coarse `max_length - margin` clamp would prevent the failure mode.
2. **Fail loudly**: a run where 100% of requests raised `BadRequestError` should not finalize as Success/0.0. This looks like a sibling of #300 (fail fast on terminal provider failure) — a 400 that deterministically affects every request is terminal in practice.

Without a clamp, every task in the 131k-token post-training regime silently produces zeros on any model with a smaller context window. Workaround used for the parity runs: per-task `-o max_tokens=<context - 4096>`.

Happy to take a stab at the clamp if the approach sounds right.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01B9VUctQuzfAtrdpk4wDjEx

abhishekraok:
[...]

PR #319, opened 2026-08-25
olmo_core: accept uncapped max_tokens in the request trace

abhishekraok:
`OlmoCoreProvider._build_generation_kwargs` rejects `max_tokens <= 0`, which raises `TypeError: '<=' not supported between instances of 'NoneType' and 'int'` when `max_tokens` is `None`.

The generation path never hit this — `generate()` calls `_resolve_max_tokens()` first, which replaces `None` with the room left in the context. But `describe_request()` builds the same kwargs *without* resolving, and `runners/asynq/processing.py` traces every request before generation. So any task using `max_tokens=None` fails on this provider before inference starts.

That affects the three migrated post-training tasks in flight — #307 `popqa:chat`, #308 `omega_500`, #309 `gpqa_*:cot` — all of which use `max_tokens=None` (generate to the model's context limit) to avoid the fixed-131k footgun in #310. They ran fine on `vllm_server`, which is why this wasn't caught by the parity runs.

`None` now passes the guard; a concrete non-positive limit is still rejected.

Regression test added: `describe_request` with `SamplingParams(max_tokens=None)` on the existing fake-provider fixture. Verified it raises `TypeError` at `olmo_core.py:384` without the fix. Full suite 1958 passed, ruff/ty clean.

Reported by Codex review on #308.

### DIFF
--- a/src/olmo_eval/inference/providers/olmo_core.py
+++ b/src/olmo_eval/inference/providers/olmo_core.py
@@ -381,7 +381,10 @@ class OlmoCoreProvider(InferenceProvider):
         return input_ids
 
     def _build_generation_kwargs(self, params: SamplingParams) -> dict[str, object]:
-        if params.max_tokens <= 0:
+        # ``max_tokens=None`` means "generate to the context limit"; the
+        # generation path resolves it per prompt via ``_resolve_max_tokens``,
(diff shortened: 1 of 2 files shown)


undfined (approved):
lgtm

PR #321, opened 2026-08-25
Add MMLU chain-of-thought tasks (post-training regime)

abhishekraok:
Registers `mmlu_{subject}:cot` for all 57 subjects plus an `mmlu:cot` suite, mirroring oe-eval's `mmlu_{sub}:cot::hamish_zs_reasoning_deepseek` (the OLMO_3 suite entry): 0-shot chat, per-subject description requesting concise reasoning then `Therefore, the answer is: X`, temp 0.6 / top-p 0.95, extraction via the shared OLMO 3 regex-template cascade.

Why a separate module rather than a variant: the base `mmlu_{subject}` tasks score by continuation log-probability, which measures MCQ ranking rather than reasoning, so the CoT regime needs different doc processing entirely — same split as `gpqa:cot`.

- Prompts and gold letters byte-identical on **3,207 real docs across 7 subjects** (abstract_algebra, anatomy, high_school_mathematics, professional_law, world_religions, moral_scenarios, college_physics)
- Scorer verdicts identical on 56 synthetic responses run through oe-eval's `ExactMatch` metric
- 12 new tests; full suite 1987 passed; ruff/ty clean

**Stacked on #309** only for `OLMO_3_ANSWER_REGEX_TEMPLATES` in `common/answer_extraction.py`. If #309 is delayed, that constant could move to #308 and this retargets to `main` — retarget to `main` once #309 lands either way.

**Not yet done:** GPU parity run. One consideration for whoever schedules it — this is 57 tasks totalling ~14k items at temp 0.6 with unbounded generation, so it is the most expensive thing in the migration so far; worth pricing before running it on every checkpoint.

### DIFF
--- a/src/olmo_eval/evals/suites/mmlu.py
+++ b/src/olmo_eval/evals/suites/mmlu.py
@@ -42,3 +42,10 @@ MMLU_BPB = make_suite(
     "mmlu:bpb",
     tuple(f"mmlu_{s}:rc:bpb" for s in MMLU_SUBJECTS),
 )
+
+MMLU_COT = make_suite(
+    "mmlu:cot",
+    _task_names_variant(MMLU_SUBJECTS, "cot"),
+    aggregation=AggregationStrategy.AVERAGE,
+    description="0-shot chain-of-thought MMLU across all 57 subjects",
+)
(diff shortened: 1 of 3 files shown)


abhishekraok:
Superseded by #339 — same commits rebased onto `main`. GitHub closed this automatically when its base branch `task/add-gpqa-cot` was deleted as part of merging #309, and it can't be reopened once the base ref is gone.

PR #322, opened 2026-08-25
Pointing + counting benchmarks (3/5)

jason718:
## Description

Third of five PRs splitting #278 (stack: #316 → #317 → this). **Milestone: the complete pointing/counting family — 9 tasks, both suites, mm_olmo `_mp` parity.**

### Tasks

| Task | N | Prompt form |
|---|---|---|
| `pixmo_points_eval` / `_mp` | 1,215 | question / model-prompt |
| `sa_co_gold_subset` / `_mp` | 4,994 / 3,856 | question / model-prompt (different prepared subsets, as in mm_olmo) |
| `sa_co_gold_point_4k_mp` | 28,672 | model-prompt, 4096/subset |
| `sa_co_gold_point_mp` | 166,766 | model-prompt, unsampled (registered, deliberately outside the suite — ~6× the 4k cost for the same measurement) |
| `countbench_qa`, `pixmo_count` | 490 / 540 | counting (`point_count` style) |

Suites: `molmo2_pointing`, `molmo2_pointing_mp`.

### Consolidation (Tyler's #5, continued)

`PointingTask` and `ImageQATask` subclass `VisionTask` and lose their byte-identical `instances`/`format_request` blocks. `tasks/single_image.py` lands here because the counting benchmarks are `ImageQATask`s that share the pointing prompt family; the QA benchmarks reuse it in PR 4.

### Scoring

`scoring/pointing.py` — COCO-RLE mask decode + maximum-bipartite point matching (pycocotools/scipy, lazily imported), the independent max-over-annotators, and the empty-GT⇒1.0 convention, all mirroring mm_olmo's `SegmentationPointingScorer`. `scoring/count_parsing.py` — the vendored count parse ladder. `scoring/vqa.py` starts with the shared answer helpers + `PointCountScorer`; the QA scorers append in PR 4.

pycocotools + scipy join the `hf` extra — the judge-style silent-zero failure mode (every instance failing to score reports f1 = 0.0000 with status `Success`) is why these are declared task dependencies too.

### Verification

- Instance counts exact vs mm_olmo for all seven datasets; first-example prompts byte-identical, including the seeded `_mp` templates (`pointing: <label>` / 43-template forms)
- The #278 numbers these tasks reproduced: `pixmo_points_eval_mp` 0.8152 vs mm_olmo's stored 0.8083, `sa_co_gold_subset_mp` 0.6631 vs 0.6658 — with 3,856/3,856 and 1,215/1,215 stored-prompt parity
- Full suite: 2,016 passed; lint/format/`ty` clean

## Type of Change

## Checklist

- [x] Any dependent changes have been merged and published (stacked on #317)

Part of #278 (tracking). Next: PR 4 — image QA.


src/olmo_eval/evals/vision/tasks/pointing.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _pointing_results(responses: Sequence[Response]) -> Iterator[dict]:
+    for response in responses:
+        for output in response.outputs:
+            if output.metadata and "pointing_result" in output.metadata:
+                yield output.metadata["pointing_result"]
+
+
+@dataclass(frozen=True)
```
undfined, line 137: **These metrics will persist the wrong per-instance values.** `Metric.compute_instance()` falls back to the scorer channel, and `PointingScorer` returns only F1. As a result, exact storage and pairwise analysis record F1 for precision, recall, the bucketed metrics, and the presence metrics. The point-count metric family has the same problem: `close`, `valid`, per-count accuracy, and category-average all fall back to `correct`. The aggregate `compute()` results are still right, which makes this easy to miss. Please add metric-specific `compute_instance()` implementations for decomposable metrics, returning `None` when an instance is outside the metric's bucket, and disable scorer fallback for weighted or category-level metrics that do not have an exact per-instance representation. A regression test that inspects the persisted instance metrics would catch both families.

undfined (commented):
I found four issues in this layer. The first two affect persisted result correctness and task identity, so I think they should be fixed before merge. The other two affect pretrain prompt fidelity and the documented scoring failure behavior.

src/olmo_eval/evals/vision/tasks/pointing.py
```diff
@@ -0,0 +1,220 @@
...
+
+    default_prompt_templates = "uber_model_v2"
+    default_system_prompt_style = "demo_or_style_v2"
+
+    def format_query(self, label: str, index: int) -> str:
+        """Render the prompt for ``label`` at build-order position ``index``."""
+        return build_pointing_prompt(
+            label,
+            index,
+            prompt_templates=self.config.prompt_templates or self.default_prompt_templates,
```
undfined, line 100: **Can we include both prompt-family settings in `TaskConfig.to_dict()`?** The generated request now depends on `prompt_templates` and `system_prompt_style`, but neither field participates in the task hash. I confirmed that `uber_model_v2` / `demo_or_style_v2` and `none` / `style_and_length_v2` currently produce the same hash despite producing different prompts. That can make artifact names collide and can cause storage or pairwise tooling to treat incompatible runs as equivalent. Please serialize both fields and add a test asserting that changing either one changes the task hash.

src/olmo_eval/evals/vision/benchmarks/pixmo_count.py
```diff
@@ -0,0 +1,64 @@
...
+        import datasets
+
+        ds = datasets.load_from_disk(str(torch_datasets_dir() / "pixmo_datasets" / "count"))
+        ds = ds[self.config.split.value]
+        # Arrow order is load-bearing: the per-example question template is
+        # picked by an RNG seeded with the arrow index.
+        for idx in range(len(ds)):
+            ex = ds[idx]
+            yield Instance(
+                question=self.apply_family_prefix(pixmo_count_question(ex["label"], idx)),
```
undfined: **This bypasses `prompt_templates=none` for pretrain checkpoints.** The upstream formatter uses the bare lowercased label in `none` mode and only applies the `point_count:` family prefix, but this always expands the label through an instruction-style counting template. For example, the intended pretrain prompt is `point_count: cats`; this produces `point_count: How many cats are there ...`. Please branch on the effective `prompt_templates` value here: use `label.lower()` for `none`, otherwise use `pixmo_count_question(...)`, and then apply the family prefix. A test for both prompt families would prevent this from drifting.

src/olmo_eval/evals/vision/scoring/pointing.py
```diff
@@ -0,0 +1,164 @@
...
+
+            with Image.open(meta["image_path"]) as im:
+                image_w, image_h = im.size
+
+        # Decode the ground-truth masks for each annotator.
+        annotator_masks: list[list[np.ndarray]] = []
+        for rles in meta["pointing_annotators"]:
+            masks: list[np.ndarray] = []
+            for seg in rles:
+                m = decode_segmentation(seg, image_h, image_w)
```
undfined, line 160: **This does not implement the zero-score failure behavior documented by `PointingTask`.** Exceptions from mask decoding propagate through synchronous scoring, so a missing `pycocotools` installation or a malformed RLE aborts the entire run rather than scoring that instance as zero. Please either catch decoding/scoring failures per output and attach a complete zero-valued `pointing_result` (plus useful error metadata), or make the task fail fast during initialization and update the documentation accordingly. A test with a decoding exception would make the intended contract explicit.

src/olmo_eval/evals/vision/tasks/pointing.py
jason718: Fixed in 9bd43f3 — PointingMetric.compute_instance returns the example's own precision/recall/f1 (None outside the metric's bucket), PresenceMetric returns membership-scoped 0/1 (None for examples outside its gt class), and both disable scorer fallback since the scorer channel is f1. Weighted metrics store None — a weighted mean is not a mean of per-instance values. Same treatment for the count family: correct/close/valid store their own fields, per-count metrics store None off-count, and the category macro-average stores nothing; Ai2dMetric got the same fix while I was in the file. Regression tests poison the scorer channel and assert nothing falls back to it (TestPointingPerInstanceMetrics, TestPointCountPerInstanceMetrics).

src/olmo_eval/evals/vision/tasks/pointing.py
jason718: Both fields are serialized in TaskConfig.to_dict (landed with the fields themselves in the framework PR — base.py:292-293), and the stack now carries the test you asked for: TestPromptFamilyTaskIdentity asserts compute_task_hash changes when either field changes.

src/olmo_eval/evals/vision/benchmarks/pixmo_count.py
jason718: Fixed in 9bd43f3 — pixmo_count branches on the effective prompt_templates: 'none' sends the bare lowercased label (verified against mm_olmo's DataFormatter point_count branch: example['label'].lower()), the templated families keep the seeded counting templates, and the point_count: prefix applies in both. countbench_qa needs no branch: its dataset carries a pre-built question, and mm_olmo's none mode also prefers example['question'] when present. Test: test_pixmo_count_follows_the_checkpoint asserts the pretrain form is exactly 'point_count: cats'.

src/olmo_eval/evals/vision/scoring/pointing.py
jason718: Fixed in 9bd43f3 — the scorer now implements the documented contract: any per-example decode/scoring failure logs, attaches a complete zero-valued pointing_result with the ground-truth count (computable from the RLE list lengths without decoding), the abstention fields, and an error string, and returns 0.0 so the run continues. Test: a malformed RLE (TestPointingScorerFailureContract) scores zero with the error recorded.

gregorybchris (changes requested):

Nice work on the parity here; the prompt-family handling and the `_mp` seeding look right, and I verified Tyler's four findings are addressed: `PointingMetric` and `PresenceMetric` have their own `compute_instance` with fallback disabled (weighted metrics store `None`), both prompt fields are in the hash, `pixmo_count` branches on `prompt_templates`, and the pointing scorer implements the documented zero-score contract with a test. Ruff, `ty`, and the unit suite (2260 passed) pass locally.

Two things to address before merge.

### Blocking

1. `src/olmo_eval/evals/vision/scoring/pointing.py:119`
   `PointingScorer.score` catches every exception and returns a zero-valued result. That satisfies the documented contract for a malformed RLE, but it also catches `ImportError` when `pycocotools` or `scipy` is missing, and any programming error in `_score`. A worker without those packages therefore scores all 4,994 SACoGold instances as `f1=0.0` and the run reports Success, which is the failure mode the PR description itself calls out. Please narrow the catch to data errors from decoding and matching (`ValueError`, `KeyError`, `TypeError`, `IndexError`), let `ImportError` and everything else propagate so the runner records a structured scoring error, and check the imports once at task initialization so a missing dependency fails before inference starts. It would also help to surface the per-example `error` field as an aggregate, for example an `n_scoring_errors` metric like CharXiv's `n_invalid`, so a partial failure is visible in `metrics.json`.

2. `src/olmo_eval/evals/vision/tasks/single_image.py:33`
   `ChartQaSubsetMetric` (and `MeanScorerMetric` on line 19) are defined here with no consumer; they are first used in #323. `ChartQaSubsetMetric` also inherits the default `compute_instance`, so the `_human` and `_aug` metrics persist the overall scorer value for every instance, including instances outside their subset. I reproduced it in the #323 branch: an `is_human=False` instance under `relaxed_correctness_human` stores `1.0`. This is the same class of bug Tyler flagged on the pointing and dense-caption metrics. Please give it a `compute_instance` that returns `None` outside the subset and disable scorer fallback, with the same poison-the-scorer-channel regression test the other metric families have. Since the class has no caller here, it may be cleaner to move it and `MeanScorerMetric` to #323 with the tasks that use them.

### Non-blocking

3. `src/olmo_eval/evals/vision/tasks/pointing.py:50`
   The comment "a missing scorer dependency scores every instance zero rather than failing the run" describes the behavior item 1 asks to change; please update it together.

4. `src/olmo_eval/evals/vision/scoring/vqa.py:17` and `scoring/pointing.py:99`
   `_response_text` is duplicated here and a third copy arrives in #324. One helper in `scoring/__init__.py` or a small `scoring/common.py` would do.

5. `pyproject.toml:60`
   `scipy~=1.17.1` is a tight pin inside the `hf` extra while `pycocotools` is unpinned. If the pin is deliberate (a known scipy break), a comment would help; otherwise a `>=` lower bound is friendlier to other extras.

6. `src/olmo_eval/evals/vision/tasks/single_image.py:144`
   `Ai2dMetric` is also forward plumbing for #323. Same suggestion as item 2: land it with its consumer, or say so in the description.

### Nits

- Nit: `sa_co_gold.py` builds `_METRICS` and `_POINT_METRICS` with the same `pointing_metrics(...)` call; the first could be reused.
- Nit: `scoring/pointing.py` opens the image to read its size when `image_size` is missing from metadata; all three tasks populate it, so a hard `KeyError` with a clear message might be preferable to per-instance file I/O inside scoring.

### Before merge

Items 1 and 2, plus the rebase noted on #316. Then merge once CI is green.

jason718:
Both blocking items and the non-blocking ones are addressed; rebased onto `main` now that #317 merged.

**1 (blocking)** — `PointingScorer.score` catches only malformed-ground-truth errors (`ValueError`, `KeyError`, `TypeError`, `IndexError`) and scores those examples zero with the error recorded. A missing `pycocotools`/`scipy`, or a bug in the scorer, now propagates so the runner records a structured scoring error rather than reporting 4,994 zeros as Success. `PointingTask` also probes both backends while building instances, so that environment fails **before a model is loaded**. And `pointing_metrics` now reports `n_scoring_errors`, so a partially unscorable run is visible in `metrics.json` and not only in the logs. Tests cover all three: the import probe, a raising scorer path, and the error count.

**2 (blocking)** — took the cleaner option you offered: `MeanScorerMetric`, `ChartQaSubsetMetric` and `Ai2dMetric` **move to #323** with the tasks that use them (nothing here consumed them). `ChartQaSubsetMetric` gets `compute_instance` returning `None` outside its subset plus disabled fallback there, with the poison-the-scorer-channel test.

**3** — comment updated with the behavior change. **4** — one `response_text` helper in `scoring/common.py` replaces the copies. One detail worth flagging: the three copies were **not** identical — the pointing one stripped, the VQA/multi-image ones did not. Since the vendored parsers do their own normalization, the helper preserves the non-stripping semantics and the pointing call site strips explicitly, so no scorer's behavior changed. CharXiv's `_response_text` takes a `Response` and deliberately ignores `extracted_answer`, so it stays separate. **5** — `scipy~=1.17.1` matches the existing pin in the `analysis` extra; annotated as such. **6** — `Ai2dMetric` moved to #323 with its consumer.

Nits: `sa_co_gold` reuses `_METRICS`; a missing `image_size` is now validated before scoring and raises (every pointing task records it, so its absence is a task bug, not bad data — this also removes per-instance file I/O from the scoring path).

Full suite 2,663 passed; lint/format/`ty` clean; no task hash on `main` changes.

gregorybchris (approved):

Approving — both blocking items are properly fixed, and the extras you added on top are the right ones. Three small things below, none of them worth holding the merge for.

I re-verified each item rather than taking the summary on trust:

- **Scoring failures are honest now.** `PointingScorer.score` catches only `ValueError`/`KeyError`/`TypeError`/`IndexError`, so a missing `pycocotools`/`scipy` or a scorer bug reaches the runner. The `require_scoring_dependencies()` probe in `PointingTask.instances` is better than what I asked for — the environment fails before a model loads, not at the first scored example. And `n_scoring_errors` makes a partially unscorable run visible in `metrics.json`. `PointingErrorCountMetric` gets `compute_instance` as a 0/1 flag with `pairwise_display_format() -> "raw"`, which is the right shape for a count.
- **Metrics moved rather than forward-plumbed.** `MeanScorerMetric`, `ChartQaSubsetMetric` and `Ai2dMetric` land in #323 with their consumers. Every metric class remaining here has `compute_instance` plus a deliberate `supports_pairwise_scorer_fallback`.
- Items 3, 4, 5, 6 and both nits check out. Good catch that the three `_response_text` copies were not identical — preserving the non-stripping semantics and stripping at the pointing call site is the right call, and worth having written down.
- **No hash churn.** `arc_easy`, `gsm8k`, `hellaswag`, `humaneval`, `drop` and `dense_caption` hash identically to `main` at `e1e16885`, and the three new pointing tasks hash the same across two consecutive processes.

Locally: 2500 passed, 30 skipped; ruff, format and `ty` clean (the one `ty` diagnostic is the pre-existing `harness/sandbox/executor.py` suppression, untouched here and not in CI's resolution). CI is green.

### Non-blocking

1. `src/olmo_eval/evals/vision/scoring/pointing.py:111`
   The private `_response_text` is now dead — line 19 imports the shared `response_text` and line 182 is the only call site. Ruff doesn't flag unused module-level functions, so nothing caught it. Please delete it so the next reader doesn't wonder which one to use.

2. `src/olmo_eval/evals/vision/tasks/pointing.py:131` and `tasks/single_image.py:18`
   `_pointing_results` and `_point_count_results` flatten every output, so a `num_samples > 1` run averages the samples instead of applying `TaskConfig.output_score_aggregation` (default MAX); `_pointing_result` / `_point_count_result` take the first output for the per-instance value. I reproduced the equivalent on the CharXiv metric in #323: two instances sampled twice with scores `[0, 1]` and `[1, 0]` give `0.5` where MAX gives `1.0`. No task here sets `num_samples > 1`, so this only bites on `-o sampling_params.num_samples=N`. `dense_caption` solved it by rejecting the configuration outright — could the pointing and count families do the same, or apply the configured aggregation before averaging?

### Nits

- Nit: `src/olmo_eval/evals/suites/molmo2.py:3` — "Grows with the vision families as they land: captioning now, pointing/counting, image QA, and multi-image next" describes in-progress state, which `CLAUDE.md` asks us to avoid, and it gets staler with each PR in the stack. It came in with #317 so it's already on `main`; this is just the natural place to fix it.

### Before merge

Nothing. Item 1 is a one-line delete if you want it in this PR, item 2 and the nit are fine as follow-ups. Merge once CI is green.

jason718:
Pushed 90094b82 with item 1 and the nit: the dead private `_response_text` is gone (it was the non-stripping variant, which is why my earlier replace aimed at the stripping one never matched it), and the `molmo2.py` docstring no longer describes the stack's in-progress state. Item 2, the `num_samples > 1` aggregation, stays a follow-up alongside #392.

jason718:
Follow-up cleanup in 1d3110be, from a naming and packaging pass on this PR:

- `scoring/vqa.py` → `scoring/counting.py`. In this PR it held only `PointCountScorer`, plus an `_answers` helper with no caller here (dropped; it arrives in #323 with the VQA scorers that use it).
- `tasks/single_image.py` → `tasks/image_qa.py`, after the `ImageQATask` it defines. The task modules are split by how a task is scored, and pointing is single-image too, so the old name was on the wrong axis.
- `pycocotools` and `scipy` move from the `hf` extra to a new **`vision`** extra. They are the pointing scorer's dependencies, needed whichever provider runs the model; under `hf`, an `olmo_core_vlm` install lacked them. The dev group pulls `vision`, so CI still exercises the scorer, and beaker jobs are unaffected since they install each task's own `dependencies`. The lock still adds only `pycocotools 2.0.11` versus `main`, with no other version changes in any install profile.
- The SA-Co gold-set builder no longer calls `rebase_data_path` on paths it already built from the current data root. The function stays for the prepared datasets, whose manifests record absolute paths.

No behavior change: the moved definitions are AST-identical, all 377 `main` task hashes are unchanged, and the suite is green (2,517 here, 2,663 at the stack tip).

PR #323, opened 2026-08-25
Image-QA benchmarks (4/5)

jason718:
## Description

Fourth of five PRs splitting #278 (stack: #316 → #317 → #322 → this). **Milestone: the flagship `molmo2_imageqa` suite — 14 tasks — runs end to end.**

### Benchmarks (12 new; the suite's other 2 are PR 3's counting tasks)

ai2d, chart_qa, doc_qa, info_qa, math_vista, mmmu, mmmu_pro, real_world_qa, text_vqa, vqa2, charxiv_descriptive, charxiv_reasoning — with their variants (`math_vista:offline|gpt`, `ai2d:transparent|test`, `chart_qa:test`). `mmmu_pro` stays decoupled from `mmmu` (own prompts/parser/scorer, faithful to the official repo; primary = SOTA-reported overall).

Suites: `molmo2_imageqa` (AVERAGE, all 0-1 primaries) and `molmo2_imageqa_caption` (DISPLAY_ONLY — dense caption reports 0-100).

### Scoring layer

- `scoring/vqa_normalization.py`, `multiple_choice.py` (the MMMU parser, shared with mmmu_pro and the PR-5 multi-image family), `math_vista_offline.py`, `mmmu_pro.py`, `charxiv.py` — vendored parsers/prompts, behavior-preserving; CharXiv's trailing whitespace is part of the official prompt bytes, so lint exempts vendored modules file-by-file
- QA scorer classes join the counting scorer in `scoring/vqa.py`; the CharXiv judge and MathVista GPT extraction join the dense-caption judge in `scoring/judges.py`, sharing its cached-call machinery
- `scoring/__init__.py` re-exports the parser functions (internal to `evals.vision`, so no cycle with `common/scorers`)

### Fixes

- `required_secrets=("OPENAI_API_KEY",)` on charxiv_descriptive/reasoning and math_vista — same beaker live-bug as dense caption in #317

### Verification

- **Whole-stack AST audit**: all 167 top-level symbols from the reviewed #278 modules are present and definition-identical in the new layout; the only 7 changed definitions are the intentional VisionTask consolidation (3) and added secrets (3), plus one import-qualifier rewrite
- mm_olmo dump-parity tests (prompt + metric parity against released reference dumps) moved and passing; instance construction covered by `test_image_qa_requests.py`
- Full suite: **2,152 passed, 9 skipped**; lint/format/`ty` clean

```bash
uv run pytest tests/evals/vision/ -q          # 185 passed — the whole vision family
uv run pytest tests/ --ignore=tests/integration -q
```

## Type of Change

## Checklist

- [x] Any dependent changes have been merged and published (stacked on #322)

Part of #278 (tracking). Next: PR 5 — multi-image + docs.


gregorybchris (changes requested):

This is a big, careful port, and the vendored parsers with their byte-identical constant pins and dump-parity tests are exactly the right way to bring these benchmarks over. Ruff, `ty`, and the unit suite (2396 passed) pass locally.

I found four issues that affect task identity or persisted results, so I think they should be fixed before merge. The first two are the same bug classes Tyler flagged on #317 and #322 recurring in new code.

### Blocking

1. `src/olmo_eval/evals/vision/scoring/vqa.py:264` and `scoring/judges.py:561`
   `MathVistaGptScorer.cache_dir` and `CharxivJudgeScorer.cache_dir` default to a fresh `tempfile.mkdtemp(...)` per process, and `Metric.to_dict` serializes scorer instances with `asdict`, so the cache path enters the task hash. I reproduced it: two consecutive processes give `charxiv_descriptive` hashes `79f3d83e...` and `ea483ce4...`, `charxiv_reasoning` `dcb9a7b2...` and `6697c850...`, and `math_vista` `eaa0b125...` and `fe846043...`. Every run of these three tasks therefore gets a unique task hash and artifact name, and storage and pairwise tooling can never match two runs of the same config. `math_vista:offline` is stable. Please give these scorers (and `DenseCaptionJudgeScorer` in #317) a `to_dict` that serializes only `type`, `name`, and `model`, and add a test that constructs the config twice and asserts equal hashes.

2. `src/olmo_eval/evals/vision/tasks/single_image.py:33`, `benchmarks/mmmu_pro.py:63`, `benchmarks/charxiv.py:111` and `:135`
   `ChartQaSubsetMetric`, `MmmuProSettingMetric`, `CharxivScoreMetric`, and `CharxivInvalidCountMetric` inherit the default `compute_instance`, which falls back to the scorer channel. I reproduced all four: an `aug` instance under `relaxed_correctness_human` stores `1.0`; a `vision` instance under `standard_10_accuracy` stores `1.0`; a non-counting instance under `descriptive_counting` stores `1.0`; and a validly graded instance under `n_invalid` stores `1.0`. Aggregates are right, so this is invisible in `metrics.json`, but prediction exports and pairwise analysis are wrong for every subset metric in the suite. Please add `compute_instance` overrides that return `None` outside the metric's scope (and `0.0`/`1.0` invalid-flag semantics for `n_invalid`), disable scorer fallback where there is no per-instance meaning, and extend the poison-the-scorer-channel regression test that the dense-caption and pointing families already have.

3. `src/olmo_eval/evals/vision/scoring/vqa.py:198`, `:292`, `:311`
   `MathVistaOfflineScorer.score` and both branches of `MathVistaGptScorer.ascore_with_context` catch every exception and return `0.0`. An expired `OPENAI_API_KEY`, a rate limit, or a bug in `normalize_extracted_answer` therefore becomes a zero for every affected instance and the run reports Success. This is the pattern Tyler asked to remove from the dense-caption judge in #317, for the same reason: the runner's `ContextScorer` wrapper only records a structured scoring error if the exception reaches it. Please let transport and programming errors propagate and handle only recognized extraction failures locally, recording a parse-status flag in `output.metadata` so a genuine zero is distinguishable from a failed extraction afterwards.

4. `src/olmo_eval/evals/vision/benchmarks/mmmu_pro.py:175`
   `MmmuProTask.format_request` overrides the base and resolves every image eagerly (`c() if callable(c)`) in the parent process, then puts the decoded PIL images on the request. That reintroduces the memory and pickling problem #317 fixed with lazy references, for up to seven images per instance across 3,460 standard-setting instances. `common.images.resolve_images` already flattens list-valued and callable entries as of #324, so this can return the lazy partials from `_attach_images` and let the provider resolve them.

### Non-blocking

5. `src/olmo_eval/evals/vision/scoring/judges.py:475`
   The CharXiv grading call retries on any exception and returns the official dummy (`score -1`) after ten attempts. That is parity-faithful, and `n_invalid` makes it visible, which is good. But an auth failure or outage still yields a complete run of zeros with Success. Could the task fail when every grading call returned the dummy, and could a 401 skip the retry ladder?

6. `src/olmo_eval/evals/vision/benchmarks/math_vista.py:40` and `benchmarks/charxiv.py:189`, `:262`
   The comments "without the client every instance scores zero instead of failing the run" describe the behavior item 3 removes. Please update them together.

7. `src/olmo_eval/evals/vision/benchmarks/real_world_qa.py:53` and `mmiu.py` in #324
   `assert` is used for dataset validation. Asserts are stripped under `-O` and read as invariants rather than data checks; please raise `ValueError` with the offending row.

8. `src/olmo_eval/evals/vision/scoring/mmmu_pro.py:150`, `scoring/judges.py:376`, and `scoring/multi_image.py` in #324
   Mid-module imports with `# noqa: E402` in three files. Moving them to the top costs nothing and removes the suppressions.

9. `src/olmo_eval/evals/suites/molmo2.py`
   `molmo2_imageqa` now includes `math_vista` (GPT extraction by default) and both CharXiv tasks, so the suite needs `OPENAI_API_KEY`. The README in #324 says two tasks call OpenAI; it is four, and the suite-level requirement is worth stating.

### Nits

- Nit: `AnlsScorer.name = "ansl"` is presumably vendored from mm_olmo's key; a comment saying the misspelling is deliberate would stop someone "fixing" it and breaking parity.
- Nit: `scoring/__init__.py` re-exports about thirty parser functions; if the goal is a public surface for tests, a shorter list would be easier to keep stable.

### Before merge

Items 1 through 4. Item 1 in particular should land together with the equivalent change in #317. Then merge once CI is green.

jason718:
All four blocking items and the non-blocking ones are addressed; rebased onto `main` now that #317 merged.

**1 (blocking)** — `CharxivJudgeScorer` and `MathVistaGptScorer` serialize only `type`/`name`/`model`, so the per-process cache directory no longer enters the task hash. I reproduced your finding first (two consecutive processes: `charxiv_descriptive` `7e772c3c` → `f055e085`, `math_vista` `b53dbf65` → `8d579c2d`, with `arc_easy` stable as the control); after the fix both are identical across processes. Tests assert hash stability across constructions and that the serialized form carries no temp path. The equivalent `DenseCaptionJudgeScorer` change landed in #317.

**2 (blocking)** — `ChartQaSubsetMetric`, `MmmuProSettingMetric`, `MmmuProOverallMetric`, `CharxivScoreMetric` and `CharxivInvalidCountMetric` implement `compute_instance` and disable scorer fallback; `n_invalid` uses the 0.0/1.0 invalid-flag semantics you specified, and the overall MMMU-Pro metric stores nothing since an average of two setting means has no per-instance decomposition. Extended the poison-the-scorer-channel test to all of them.

**3 (blocking)** — `MathVistaOfflineScorer.score` and both branches of `MathVistaGptScorer.ascore_with_context` catch only extraction failures now; an expired key, a rate limit or a bug in `normalize_extracted_answer` propagates to the runner's scoring errors. A recorded zero is distinguishable from a wrong answer via `output.metadata["extraction_failed"]`.

**4 (blocking)** — already fixed: `MmmuProTask` uses `_attach_images` and returns the lazy partials. That surfaced a related gap, since `resolve_images` at this point in the stack kept `None` entries — the unused slots of MMMU-Pro's variable image count would have reached the provider as holes. It now drops them (and flattens list entries, which #324 needs).

**5** — the CharXiv grader stops retrying on an auth failure instead of burning ten attempts to produce dummy scores (401/403 or an `AuthenticationError`). Left the all-dummy run check out for now: `n_invalid` makes it visible, and failing the task is a policy call better made once #392 (`ResponseBatchScorer`) reworks this path. **6** — comments updated. **7** — `real_world_qa` raises `ValueError` with the offending row. **8** — all mid-module imports hoisted; no `# noqa: E402` remains anywhere in `evals/vision/`. **9** — the README now says four tasks call OpenAI and that the suite needs the key.

Nits: the `ansl` misspelling is annotated as deliberate mm_olmo key parity.

Full suite 2,663 passed; lint/format/`ty` clean; no task hash on `main` changes.

gregorybchris (approved):

Approving — all four blocking items are fixed, and the `resolve_images` gap you found while fixing item 4 was a good catch. Two non-blocking notes below.

Reproduced rather than read:

- **Task hashes are stable.** `charxiv_descriptive` `3aba511be7d9025f`, `charxiv_reasoning` `b362f8d52bd06cee`, `math_vista` `cc81822ca775050b` — identical across two consecutive processes and unchanged under `MATHVISTA_GPT_CACHE_DIR=/tmp/zz`. They were changing every process before (`79ca4430` → `73089a4c` and so on). Existing tasks are untouched: `arc_easy`, `gsm8k`, `hellaswag`, `humaneval`, `drop` all match `main` at `e1e16885`.
- **Per-instance metrics are scoped.** `ChartQaSubsetMetric(human)` stores `1.0` for a human example and `None` for an augmented one; `MmmuProSettingMetric(standard10)` stores `1.0` in-setting and `None` for a vision example; `MmmuProOverallMetric` stores `None`; `CharxivScoreMetric` stores `1.0` graded / `0.0` on the dummy; `n_invalid` stores `0.0` graded and `1.0` invalid. All five report `supports_pairwise_scorer_fallback() == False`. Leaving `MeanScorerMetric` on the inherited fallback is right — it is literally the mean of the scorer channel.
- **Honest zeros.** Both MathVista scorers catch only extraction errors, and `output.metadata["extraction_failed"]` distinguishes a recorded zero from a wrong answer. The remaining `except Exception` in `scoring/math_vista_offline.py` are inside the vendored offline parser and are parse-level, which is the right place for them. The CharXiv auth short-circuit raises instead of burning the ladder into dummy scores.
- **Lazy images.** `MmmuProTask._attach_images` returns the partials, and `resolve_images` dropping `None` entries is the right fix — the unused slots of MMMU-Pro's variable image count would otherwise have reached the provider as holes.
- Items 5 through 8 and the `ansl` nit all check out; no `# noqa: E402` remains anywhere in `evals/vision/`.

Locally: 2643 passed, 32 skipped; ruff, format and `ty` clean (the single `ty` diagnostic is the pre-existing `harness/sandbox/executor.py` suppression, untouched here). CI is green.

### Non-blocking

1. `src/olmo_eval/evals/vision/benchmarks/charxiv.py:99`
   `_results` flattens every output, so a `num_samples > 1` run averages the samples rather than applying `TaskConfig.output_score_aggregation` (default MAX), and `_result_for` (`:106`) takes the first output for the per-instance value. I reproduced it: two instances sampled twice with scores `[0, 1]` and `[1, 0]` give `compute() == 0.5` and `compute_instance() == 0.0`, where MAX aggregation gives `1.0`. The same shape is in `tasks/pointing.py:131` and `tasks/single_image.py:18`, noted on #322. Nothing here sets `num_samples > 1`, so it only bites on an explicit `-o sampling_params.num_samples=N` — but `dense_caption` rejects that configuration outright, and it would be good for the judged benchmarks to be consistent with it one way or the other.

2. Item 9 is handled, but not in this PR
   The README lives in #324, and #324 does now say four tasks call OpenAI and that `molmo2_imageqa` needs the key as a suite — so this is genuinely covered. Just flagging that the note above describing it as landed here doesn't match the diff; this PR touches no `README.md`. Worth keeping the two straight so the item doesn't look done twice or not at all.

### Before merge

Nothing. Both items are fine as follow-ups, and item 1 would sit naturally with #392. Merge once CI is green.

jason718:
Correction to my earlier reply: I listed item 9 (the OpenAI-key README note) as addressed here, but this PR doesn't touch `README.md`. The fix is in #324, whose README now says four tasks call OpenAI and that `molmo2_imageqa` needs the key as a suite. Thanks for catching the mismatch.

PR #327, opened 2026-08-26
Fail loudly on partial instance failures

yilunzhao:
Fixes the reporting half of #310: a task can lose most of its instances and still report a clean, scored, exit-0 run.

We hit this on a real run: a healthy vLLM server rejected 47 of 63 instances one by one (agentic loops outgrew the context window, each request 400'd), and the run still logged `Processed 63/63`, printed a green Success row scored on the surviving 16, and exited 0. The only trace of the 47 dead instances was one WARNING line deep in the log — `compute_task_metrics` builds an error summary from failed instances and then never attaches it, nothing compares saved against processed, and nothing fails a run on its failure rate.

## Changes

- **Attach the error summary that is already built.** `TaskResult.error_summary` is now set (a separate field — setting `error` would drop the task's metrics entirely).
- **Record saved-vs-processed accounting.** `instances_processed` and `instances_failed` join the existing saved count, and show up in the completion line, a new `Instances` column in the results table, `metrics.json`, and the stored task row.
- **Gate on the hard-failure rate.** Above `max_hard_failure_rate` (default 0.05) the task is marked failed and the run exits 1 — after results are written, so `metrics.json` and predictions stay on disk. Separately, a task that saves zero instances fails at any threshold: it produced no metrics, so there is nothing to publish. Hard failure = error with no output; soft failures (e.g. `MaxTurnsExceeded` with a fallback answer) still score, as before.

```json
"num_instances": 16,
"instances_processed": 63,
"instances_failed": 47,
"error_summary": "47 instances failed (first: VLLMValidationError | status_code: 400 | ...)"
```

## Notes for reviewers

- A run that used to pass with heavy instance loss now exits 1. Set `max_hard_failure_rate=1.0` (YAML or `-o`) to stop failing runs on partial instance loss; a task that saves zero instances still fails, since it produced no metrics. The accounting is recorded either way.
- A gated task publishes no metrics, so it drops out of its suite average (deliberate — a subset score shouldn't pollute the average); aggregation logs each exclusion.
- The `task_results` table gains three columns via a migration that does **not** auto-run: apply `make db-upgrade` before rolling out this code, or inserts will fail.

## Out of scope

- The `max_tokens` clamp (#310's other half) — this PR only makes those failures impossible to miss.
- Soft-failure visibility: an instance that gives up but returns a fallback answer still reads as saved while scoring near zero. Bigger remaining gap; needs its own counters.
- Tasks that fail before dispatching any instance, and the Beaker UI's processed-count line.

## Tests

New suite drives a healthy provider stub that 400s a fixed subset of instances through the real request path: the incident shape (47/63) and the 100% case both turn loud, clean runs stay green, threshold boundaries and the CLI override are covered, and a mutation-style test pins the write-then-gate ordering. Full unit suite, lint, and type checks pass; the migration round-trips against a real Postgres with columns matching the ORM.


src/olmo_eval/runners/asynq/preparation.py
```diff
@@ -202,21 +211,35 @@ def compute_task_metrics(
...
             config=task.config.to_dict(),
             num_instances=0,
             metrics={},
-            error=error_summary,
+            error=gate_error or summary,
             duration_seconds=duration_seconds,
+            error_summary=summary,
+            instances_processed=total_instances,
+            instances_failed=instances_failed,
+            hard_failure_rate_exceeded=gate_error is not None,
```
undfined, line 242: An all-failed task can still exit successfully when `max_hard_failure_rate=1.0`. In that case, `gate_error` is `None`, so the result has an error but `hard_failure_rate_exceeded=False`. Since `check_hard_failure_gate()` only checks that flag, the CLI exits 0 even though the task produced no metrics.

We should either require `max_hard_failure_rate < 1.0` or always fail the run when no responses were scored. It would also be worth validating the configured range so values like `5` don’t accidentally disable the gate.

src/olmo_eval/storage/base.py
```diff
@@ -163,6 +163,9 @@ def convert_runner_results(
                 task_hash=task_hash,
                 task_config=task_config,
                 num_instances=task_data.get("num_instances"),
+                instances_processed=task_data.get("instances_processed"),
+                instances_failed=task_data.get("instances_failed"),
+                error_summary=task_data.get("error_summary") or task_data.get("error"),
```
undfined: This drops the gate error whenever `error_summary` is present. For example, Postgres stores `47 instances failed (first: boom)` but loses `Hard failure rate 74.6% exceeds the maximum of 5.0%`.

There isn’t another error or status field on the stored task row, so database consumers can’t tell that the threshold was exceeded. Could we persist the task error separately, or combine it with the instance summary here?

src/olmo_eval/runners/asynq/preparation.py
yilunzhao: Good catch — fixed by splitting the gate into two conditions: the rate budget still governs partial loss (so 1.0 keeps tolerating it), and a task with instances processed but zero saved now fails on its own condition at any threshold, since there is no rate at which an empty result is worth publishing. Docs and README updated; tests cover all-failed at 1.0 (loud), partial at 1.0 (quiet), all-failed at default (loud).

src/olmo_eval/storage/base.py
yilunzhao: Fixed — the stored row now combines both into the single field, classification first: "Hard failure rate 74.6% exceeds the maximum of 5.0% | 47 instances failed (first: ...)". Identical halves aren't duplicated. Storage test pins the combined form.

undfined (approved):
lgtm! note that you'll need to run the migrations to get the updated schema. Docs should cover this but ping me if you have trouble.

yilunzhao:
Thanks — we'll run the migration before pointing the updated code at the shared database.

PR #328, opened 2026-08-26
Stop harness overrides from mutating cached presets

yilunzhao:
An `-o scaffold_kwargs...` override writes into the preset's own dict.

`_apply_harness_overrides` passes whatever `HarnessConfig.to_dict()` returns straight to `_apply_dotlist_overrides`, which sets its leaves in place. `to_dict` handed back the config's own `scaffold_kwargs`, so those writes landed on the config itself — and presets are cached for the life of the process, so one launch's override survived into every config built from that preset afterwards.

Nothing has caught this because no preset currently ships `scaffold_kwargs`; the first one to do so would silently inherit the previous launch's settings. Fixed with a deep copy in `to_dict`.

Tests: three regression tests in `tests/core/harness/test_config.py` covering the returned dict, a single override, and two overrides accumulating. All three fail without the one-line change.


gregorybchris (approved):
Verified locally. The aliasing is real: with the original `to_dict`, two different `-o scaffold_kwargs...` overrides both stuck to a cached preset. With this change the preset stays clean. All three new tests fail against the original code and against a shallow `dict()` copy, so they discriminate correctly.

Non-blocking nit: `test_two_overrides_of_one_config_do_not_accumulate` could also assert `second.scaffold_kwargs["model_settings"]["top_p"] == 0.9`.

The same leak exists one level over. `ProviderConfig.to_dict()` makes a shallow copy of `kwargs`, so `-o provider.kwargs.extra_body.b=2` still mutates the cached preset. I'll open a follow-up that deep-copies once in the override path.

PR #329, opened 2026-08-26
Default enable_thinking off in the agents scaffold for vLLM-served runs

yilunzhao:
The `openai_agents` scaffold builds its own OpenAI client via `provider.get_openai_client()` and never goes through the provider's generate path, so a provider-level `chat_template_kwargs` was inert here. The request body this scaffold builds is the only place the value reaches the server.

That matters because no agentic preset configures a vLLM reasoning parser. A backbone whose chat template defaults thinking on would silently run in thinking mode and leak `<think>` blocks straight into the scored output. The scaffold now sends `chat_template_kwargs={"enable_thinking": false}` by default, gated to the vLLM-server provider class — the one that exposes a `chat_template_kwargs` attribute — because managed APIs reject unrecognized body fields with a 400. (A `vllm_server` provider pointed at a managed `base_url` would now get the field by default; before, that only happened when the kwarg was set explicitly.) An explicitly configured value still wins; the default only fills a gap.

For a model whose chat template already defaults thinking off, the kwarg is a no-op and nothing changes. This pins an implicit template default explicitly.

Tests: nine cases in `tests/core/harness/test_openai_agents_scaffold.py` covering the default, an explicit `enable_thinking: true` winning, unrelated configured kwargs keeping the default, a managed-API provider getting no injection at all, a metrics-wrapped provider still being recognized, the built dict aliasing neither the provider config nor the module-level default, and the logged outcomes. One of them drives the SDK with a stub client to confirm the value lands in the chat completions request body.


src/olmo_eval/harness/scaffolds/openai_agents.py
```diff
@@ -29,6 +29,45 @@
...
+    Args:
+        provider: The inference provider backing the agent.
+
+    Returns:
+        The kwargs to send in the request body, or None if the provider does not
+        support the field.
+    """
+    configured = getattr(provider, "chat_template_kwargs", _UNSUPPORTED)
+    if configured is _UNSUPPORTED:
+        return None
```
undfined, line 72: We probably want to log something here about this outcome. wdyt?

undfined (approved):
Looks good, one suggestion

src/olmo_eval/harness/scaffolds/openai_agents.py
yilunzhao: Added — the default now logs once at INFO per worker (the resolver sits behind the agent cache, so it runs once per worker init, not per instance), with the no-op outcomes at DEBUG. Two caplog tests pin the behavior.

undfined (approved):

PR #331, opened 2026-08-27
Self-correct unknown tool calls via a hidden fallback tool

yilunzhao:
Revives a change that was reviewed and approved earlier but never landed — the branch was closed during a cleanup. Re-cut onto current `main`, which has since gained the forced-final-answer path for `max_turns`; that path and the `[Tool error: …]` backstop are unchanged.

## What it does

When a weak model calls a tool that does not exist — Qwen3.5-9B calling a tool named `thought`, treating its own reasoning as a tool call — the SDK raises from its own dispatch and the run's entire answer becomes the string `[Tool error: Tool thought not found in agent openai_agents]`. The scaffold now rewrites an unknown call onto a hidden internal fallback tool whose result names the tools that do exist, so the model gets an ordinary tool result, corrects itself, and finishes. Retries still consume the `max_turns` budget, and the fumble stays visible in the trajectory.

## Validation

The original unit tests (rewrite, hidden-tool wiring, reserved-name guard, kwargs forwarding) carry over. Added tests drive the real SDK dispatch path against a scripted client: one builds an agent without the fallback tool and asserts the exact failure string, pinning the behavior being worked around; the others show that with the fix an unknown call returns as a tool result, the loop continues, and the answer survives — including repeated unknown calls, and endless ones still terminating on the max-turns path. A normal run is unaffected.

Newer `openai-agents` releases handle this in the SDK; upgrading is tracked separately, since our pinned version cannot move without also resolving the `openai` client constraint.


yilunzhao:
On hold by design: this lands after the SDK upgrade tracked in #332, as the backstop layer beneath the SDK's native unknown-tool handling (which doesn't cover custom/freeform tools). Will be rebased once the upgrade merges.

yilunzhao:
Closing this in favour of the SDK upgrade in #333. Stacking these commits on that branch and running them through the scaffold showed that the native tool-not-found path covers everything this PR does: an unknown call comes back as a tool result naming the real tools, repeated bad names still terminate on the max-turns path, and give-up semantics and normal runs are unchanged. On one point the native path is better: the trajectory keeps the name the model actually called, where the hidden tool rewrote it to `tool_not_found` and buried the real name in the arguments.

The one gap left is the forced final answer, which runs with no tools but inherited the main loop's formatter and so would name tools it cannot call. That is a four-line fix with the end-to-end tests from here; it follows in #336.

PR #333, opened 2026-08-27
Upgrade openai-agents to 0.20.x with native unknown-tool recovery

yilunzhao:
`openai-agents` moves from 0.7.0 to the 0.20.x series — the last line before 0.21 crosses to `openai` 3.x.

The `clients` extra pinned `openai~=2.21.0`, a patch pin left over from when 2.21 was current; 0.20.0 declares `openai>=2.45,<3` itself, so that pin stopped being protective and became simply unsatisfiable, and it now tracks the same floor and the same major boundary. The `openhands` conflict entry stays, because its `openai==2.8` constraint is independent of any of this.

Every SDK surface this scaffold touches survives the jump: `Agent` is still a stdlib dataclass, so the forced-final-answer path's `dataclasses.replace` still works; `ModelSettings.extra_body` still reaches `chat.completions.create`; `function_tool(strict_mode=False)` still returns a mutable `FunctionTool`; and the converter the vLLM `strict` patch targets is unchanged. Two error paths that were matched by exception class *name* are now real imports from `agents.exceptions`.

The SDK grew an opt-in for unresolved tool calls in 0.17.4, so a hallucinated tool name no longer aborts a whole instance with `ModelBehaviorError`. The SDK default message is `Tool 'x' not found.`, which tells the model nothing about what it should have called, and `ToolErrorFormatterArgs` carries the failed name but no inventory of valid ones — so the formatter closes over the real tool names and tells the model exactly which tools it can call. This covers function tools only: the SDK has no equivalent branch for custom/freeform calls, which still raise. Refusals changed shape too — they used to arrive as empty final output and now raise `ModelRefusalError` — and are recorded as a scored instance carrying an error rather than taking down the run.

A paired A/B on DeepScholar-Bench at limit 20 with Qwen3.5-9B, two replicates per arm, shows nothing attributable to the upgrade: exportable rate moves less between the arms than it does between two replicates of the *same* arm, and the unknown-tool path never fires here because the preset exposes a single tool.

Tests: six cases in `tests/core/harness/test_openai_agents_scaffold.py` — three on the formatter itself (the message it builds, deferring to the SDK default for approval rejections, and an empty tool inventory), one that the scaffold really does opt in and passes the resolved tool names, one that drives the SDK's own dispatch with a scripted client so an unknown call comes back as our message instead of raising, and one on the refusal path.

Two open scaffold PRs touch this file and will need rebasing: the `enable_thinking` chat-template default and the fallback-tool unknown-tool correction.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ litellm = [
     "litellm~=1.80.11",
 ]
 agents = [
-    "openai-agents~=0.7.0",
+    "openai-agents~=0.20.0",
     "httpx~=0.28.1",
 ]
 analysis = [
@@ -95,7 +95,9 @@ gpu = [
     "nvidia-ml-py>=12.560 ; platform_system != 'Darwin'",
 ]
 clients = [
-    "openai~=2.21.0",
+    # openai-agents pulls openai>=2.45 into the same resolution, so a patch pin here is
+    # unsatisfiable rather than protective. Track the same floor and the 3.x major boundary.
+    "openai>=2.45,<3",
 ]
 storage = [
     "olmo-eval[s3,postgres]",
--- a/src/olmo_eval/harness/scaffolds/openai_agents.py
+++ b/src/olmo_eval/harness/scaffolds/openai_agents.py
@@ -88,6 +88,35 @@ def _resolve_chat_template_kwargs(provider: InferenceProvider) -> dict[str, Any]
     return resolved
 
 
+def _make_tool_error_formatter(valid_tool_names: Sequence[str]) -> Any:
+    """Build a ``RunConfig.tool_error_formatter`` that names the tools the model may call.
+
+    The SDK default for a missing tool is ``Tool 'x' not found.``, which tells the model
+    nothing about what it should have called. Weak models only recover when the error
+    spells out the real names, so the inventory is captured here: ``ToolErrorFormatterArgs``
+    carries the *failed* name but no list of valid ones.
+
+    Args:
+        valid_tool_names: Names the model is actually allowed to call.
+
+    Returns:
+        A formatter suitable for ``RunConfig.tool_error_formatter``.
+    """
+    if valid_tool_names:
+        guidance = f"Available tools: {', '.join(valid_tool_names)}. Call one of these exact names."
+    else:
+        guidance = "No tools are available."
+
+    def format_tool_error(args: Any) -> str | None:
+        # Approval rejections are routed through this same hook; returning None leaves
+        # the SDK default in place for every kind we have nothing better to say about.
+        if args.kind != "tool_not_found":
+            return None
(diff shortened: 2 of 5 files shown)


undfined (approved):
Looks good. I'm iffy on test coverage for the openai client and it's used in a number of places. I would confirm that the client version that gets pulled in transitively is not breaking anything at the provider layer.

yilunzhao:
Rebased onto main. The overlap with the enable_thinking change was additive on both sides, so both survive; the scaffold's full test set passes with that default and this PR's exception handling together.

On the client caveat: the surface this repo consumes from `openai` is small. Two `create` methods and about ten response fields, no `openai.types` imports in production code, no Responses API, no streaming. Diffing every type we touch between the old and new client, the only new required field is `InputTokensDetails.cache_write_tokens`, in the Responses namespace no provider uses (that is the field that broke agents 0.7.0). Its chat-side counterpart arrived optional, and the exception classes [retry.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/retry.py) resolves by name are unchanged.

Coverage was the real gap: the existing provider mocks set `usage=None`, so a shape change would have zeroed token accounting rather than failed. [test_openai_client_surface.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/tests/providers/test_openai_client_surface.py) now drives each consuming site with real SDK response objects: [vllm_server.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/providers/vllm_server.py), the client factory in [litellm.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/providers/litellm.py), [retry.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/retry.py), the [metrics collector](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/metrics/core/collector.py), and the converter patch in [utils.py](https://github.com/allenai/olmo-eval/blob/b278b94c72fde6d5d5fbb1b445db090403acd459/src/olmo_eval/inference/utils.py). It needs the `openai` extra, so like the other scaffold tests it runs locally and skips in CI.

Live: the vLLM-server path is covered by the earlier paired A/B, and the managed-API path by a litellm smoke through sol on three instances (all scored, tool calls and usage intact). The two launch shapes pull different extras, so both client-constraint routes are exercised. The smoke needs the paper-search task that is still only on the feature branch, so it ran from a graft of this branch onto that one rather than from this branch alone.

Lint, types, and the unit suite match main.

PR #334, opened 2026-08-27
Fix ZebraLogic chat request formatting

### LINKED ISSUE #313: zebralogic:chat never applies its ChatFormatter — scored in completion mode
`register_variant("zebralogic", "chat", formatter=ChatFormatter(), ...)` sets a formatter, but the `ZebraLogic` task class pins `request_type` to `RequestType.COMPLETION` and its `format_request` never consults `config.formatter` (`src/olmo_eval/evals/tasks/zebralogic.py`). So the `:chat` variant silently runs as a raw completion — no chat template applied.

Confirmed empirically by a cross-architecture SFT eval session across three independent runs: `zebralogic:chat`'s logged requests carry a raw string context (starts `"\n# Example Puzzle"`), while genuinely chat-formatted tasks in the same jobs (harmbench, ifeval_mt_wildchat) carry role/content message lists.

Impact is largest on small instruct-tuned models that depend on their chat template: a 1.3B-active MoE scored 0.007 vs 0.182/0.171 for 7B dense/hybrid on this task, and that gap is likely an artifact.

Fix shape: follow the idiom used elsewhere (e.g. gpqa.py, code_fresh.py) — `request_type` and `format_request` should defer to `config.formatter` when set. Note this will shift existing `zebralogic:chat` numbers, which are currently completion-mode numbers.

Excelius-Wang:
## Summary

- honor the configured formatter when ZebraLogic builds an LM request
- keep the base ZebraLogic task on its existing completion path
- add CPU-only regression coverage for both task variants

Fixes #313

## Testing

- `uv run --no-group vllm pytest tests/evals/tasks/test_zebralogic.py`
- `uv run --no-group vllm ruff check src/ tests/`
- `uv run --no-group vllm ruff format --check src/ tests/`
- `uv run --no-group vllm ty check src/ alembic/`
- full non-integration suite: `1997 passed, 7 skipped`


abhishekraok:
Verified against #313:

- Base task unchanged: no formatter → `COMPLETION`, raw prompt (confirmed by test).
- `:chat` now produces a real chat request — `request_type` correctly falls back to the base property that derives from `config.formatter`, same idiom as gpqa/code_fresh.
- Fewshot path is safe: `num_fewshot=0` short-circuits `get_fewshot()` to `[]` with no dataset access.
- Full non-integration suite on the merge with current main: **2166 passed**.

Approved the CI workflow runs (fork PRs need maintainer approval) and updated the branch. Note for anyone comparing runs: pre-fix `zebralogic:chat` numbers were completion-mode and are not comparable.

abhishekraok (approved):

abhishekraok:
Thanks for the fix.

PR #335, opened 2026-08-27
fix(gsm8k): honor zero-shot override

no0p:
## Summary

- honor  for GSM8K instead of silently retaining all eight fixed demonstrations
- add prompt-level regression coverage for 0-, 3-, and 8-shot configurations

## Validation

- ============================= test session starts ==============================
platform linux -- Python 3.12.13, pytest-9.0.3, pluggy-1.6.0
rootdir: /home/robert/proj/olmo-eval
configfile: pyproject.toml
plugins: anyio-4.9.0, cov-7.1.0
collected 42 items

tests/evals/tasks/test_tasks.py ........................................ [ 95%]
..                                                                       [100%]

============================== 42 passed in 0.39s ============================== (42 passed)
- ruff format..............................................................Passed
ruff check...............................................................Passed

### DIFF
--- a/src/olmo_eval/evals/tasks/gsm8k.py
+++ b/src/olmo_eval/evals/tasks/gsm8k.py
@@ -74,10 +74,7 @@ class GSM8K(Task):
                     },
                 )
             )
-        num = self.config.num_fewshot
-        if num and num < len(instances):
-            instances = instances[:num]
-        return instances
+        return instances[: self.config.num_fewshot]
 
     def format_request(self, instance: Instance) -> LMRequest:
         fewshot = self.get_fewshot()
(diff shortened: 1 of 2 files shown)


gregorybchris (approved):

gregorybchris:
Nice explanation from Claude for any future readers of this PR: 

> What the bug was. In _build_fewshot the old guard was if num and num < len(instances). When num_fewshot is 0 that condition is false, so the slice never ran and all 8 fixed examples stayed in. Asking for GSM8K 0-shot quietly gave you 8-shot. The new line, instances[: self.config.num_fewshot], gives the same result as the old code for every other value

PR #339, opened 2026-08-31
Add MMLU chain-of-thought tasks (post-training regime)

abhishekraok:
Registers `mmlu_{subject}:cot` for all 57 subjects plus an `mmlu:cot` suite, mirroring oe-eval's `mmlu_{sub}:cot::hamish_zs_reasoning_deepseek` (the OLMO_3 suite entry): 0-shot chat, per-subject description requesting concise reasoning then `Therefore, the answer is: X`, temp 0.6 / top-p 0.95, extraction via the shared OLMO 3 regex-template cascade.

Why a separate module rather than a variant: the base `mmlu_{subject}` tasks score by continuation log-probability, which measures MCQ ranking rather than reasoning, so the CoT regime needs different doc processing entirely — same split as `gpqa:cot`.

- Prompts and gold letters byte-identical on **3,207 real docs across 7 subjects** (abstract_algebra, anatomy, high_school_mathematics, professional_law, world_religions, moral_scenarios, college_physics), re-run against the current head after the extraction module moved to `evals/extract/`
- Scorer verdicts identical on 56 synthetic responses run through oe-eval's `ExactMatch` metric
- 17 tests (pytest-style); ruff/ty clean

**Parity:** scored the paper's Olmo-3-7B-Think oe-eval generations (14,042 instances, 57 subjects) through this task: prompts identical to the logged requests, 14,042/14,042 verdicts agree, macro suite score 0.764283 reproduced. Reasoning inside `<think>` is stripped before extraction, matching oe-eval's model-level `r1_style` processing (#348).

Replaces #321, which GitHub closed automatically when its base branch (`task/add-gpqa-cot`, #309) was deleted on merge. Same commits, rebased onto `main`.

### DIFF
--- a/src/olmo_eval/evals/tasks/mmlu_cot.py
+++ b/src/olmo_eval/evals/tasks/mmlu_cot.py
@@ -29,6 +29,7 @@ from olmo_eval.data import DataSource
 from olmo_eval.evals.extract import (
     OLMO_3_ANSWER_REGEX_TEMPLATES,
     extract_answer_with_format,
+    extract_think_answer,
 )
 from olmo_eval.evals.tasks.common import Task, register
 from olmo_eval.evals.tasks.mmlu import DEFAULT_MMLU_PATH, MMLU_SUBJECTS
@@ -65,9 +66,14 @@ def _mcq_query(question: str, choices: list[str]) -> str:
 
 
 def _extract_letter(text: str) -> str:
-    """Answer letter via the shared regex cascade; empty when nothing matches."""
+    """Answer letter via the shared regex cascade; empty when nothing matches.
+
+    Reasoning enclosed in ``<think>`` tags is dropped first, as the reference
+    harness does for reasoning models, so letters mentioned while thinking
+    cannot be mistaken for the final answer.
+    """
     answer = extract_answer_with_format(
-        text,
+        extract_think_answer(text) or "",
         answer_regexes=_ANSWER_REGEXES,
         answer_regexes_templates=OLMO_3_ANSWER_REGEX_TEMPLATES,
     ).answer
--- a/src/olmo_eval/evals/tasks/mmlu_cot.py
+++ b/src/olmo_eval/evals/tasks/mmlu_cot.py
@@ -17,10 +17,6 @@ from collections.abc import Iterator
 from dataclasses import dataclass
 from typing import Any
 
-from olmo_eval.common.answer_extraction import (
-    OLMO_3_ANSWER_REGEX_TEMPLATES,
-    extract_answer_with_format,
-)
 from olmo_eval.common.metrics import AccuracyMetric
 from olmo_eval.common.scorers.base import Scorer
 from olmo_eval.common.types import (
@@ -32,6 +28,10 @@ from olmo_eval.common.types import (
     Split,
 )
 from olmo_eval.data import DataSource
+from olmo_eval.evals.extract import (
+    OLMO_3_ANSWER_REGEX_TEMPLATES,
+    extract_answer_with_format,
+)
 from olmo_eval.evals.tasks.common import Task, register
 from olmo_eval.evals.tasks.mmlu import DEFAULT_MMLU_PATH, MMLU_SUBJECTS
 
--- a/src/olmo_eval/evals/tasks/mmlu_cot.py
+++ b/src/olmo_eval/evals/tasks/mmlu_cot.py
@@ -1,0 +1,155 @@
+"""MMLU chain-of-thought tasks for the post-training regime.
+
+The base ``mmlu_{subject}`` tasks score by continuation log-probability,
+which measures multiple-choice ranking rather than reasoning. These tasks
+ask the model to reason in chat format and state a letter, then extract it
+with the shared regex cascade — matching the configuration used to report
+chain-of-thought MMLU for post-trained models.
+
+One task per subject (``mmlu_abstract_algebra:cot``, ...); the
+``mmlu:cot`` suite aggregates all 57.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Iterator
+from dataclasses import dataclass
+from typing import Any
+
+from olmo_eval.common.answer_extraction import (
+    OLMO_3_ANSWER_REGEX_TEMPLATES,
+    extract_answer_with_format,
+)
+from olmo_eval.common.metrics import AccuracyMetric
+from olmo_eval.common.scorers.base import Scorer
+from olmo_eval.common.types import (
+    Instance,
+    LMOutput,
+    LMRequest,
+    RequestType,
+    SamplingParams,
+    Split,
+)
+from olmo_eval.data import DataSource
+from olmo_eval.evals.tasks.common import Task, register
+from olmo_eval.evals.tasks.mmlu import DEFAULT_MMLU_PATH, MMLU_SUBJECTS
+
+_CHOICE_LABELS = ("A", "B", "C", "D", "E")
+
+_ANSWER_REGEXES = (r"\(?([A-D])\)?",)
+
+_COT_SAMPLING = SamplingParams(
+    max_tokens=None,
+    temperature=0.6,
+    top_p=0.95,
+)
+
+
+def _subject_text(subject: str) -> str:
+    return subject.replace("_", " ")
+
+
+def _description(subject: str) -> str:
+    return (
+        f"The following are multiple choice questions about {_subject_text(subject)}. "
+        "Summarize your reasoning concisely, then conclude with "
+        "'Therefore, the answer is: X' where X is one of A, B, C, or D.\n\n"
+    )
+
+
+def _mcq_query(question: str, choices: list[str]) -> str:
+    """Question with lettered choices, no trailing answer prefix."""
+    choices_text = "\n".join(
+        f" {label}. {text}" for label, text in zip(_CHOICE_LABELS, choices, strict=False)
+    )
+    return f"Question: {question}\n{choices_text}\n"
+
+
+def _extract_letter(text: str) -> str:
+    answer = extract_answer_with_format(
+        text,
+        answer_regexes=_ANSWER_REGEXES,
+        answer_regexes_templates=OLMO_3_ANSWER_REGEX_TEMPLATES,
+    ).answer
+    return re.sub(r"\(|\)", "", answer)
+
+
+@dataclass(frozen=True, slots=True)
+class MMLUCoTExactMatchScorer(Scorer):
+    """Case-insensitive exact match on the extracted answer letter."""
+
+    name: str = "exact_match"
+
+    def score(self, instance: Instance, output: LMOutput) -> float:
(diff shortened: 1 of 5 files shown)


tests/evals/tasks/test_mmlu_cot.py
```diff
@@ -0,0 +1,108 @@
...
+        self.assertEqual(request.messages[0]["role"], "user")
+
+    def test_skips_malformed_docs(self) -> None:
+        task = get_task("mmlu_anatomy:cot")
+        self.assertIsNone(task.process_doc({"question": "", "choices": ["a"], "answer": 0}))
+        self.assertIsNone(task.process_doc({"question": "q", "choices": [], "answer": 0}))
+        self.assertIsNone(task.process_doc({"question": "q", "choices": ["a"], "answer": "B"}))
+
+
+class TestMMLUCoTScorer(unittest.TestCase):
```
gregorybchris: Looks like a lot of existing tests in this repo use pytest rather than unittest. Perhaps we should stay consistent

gregorybchris (approved):

src/olmo_eval/evals/tasks/mmlu_cot.py
```diff
@@ -0,0 +1,155 @@
...
+from olmo_eval.common.types import (
+    Instance,
+    LMOutput,
+    LMRequest,
+    RequestType,
+    SamplingParams,
+    Split,
+)
+from olmo_eval.data import DataSource
+from olmo_eval.evals.extract import (
```
mgmorgan23, line 29: Not blocking of this PR, but it looks like `extract/answer_format.py` and `extract/mcq.py` may be now doing essentially the same thing -- we might want to compare and either standardize mcq formats or more thoroughly document which is which

src/olmo_eval/evals/tasks/mmlu_cot.py
```diff
@@ -0,0 +1,155 @@
...
+    extract_answer_with_format,
+)
+from olmo_eval.evals.tasks.common import Task, register
+from olmo_eval.evals.tasks.mmlu import DEFAULT_MMLU_PATH, MMLU_SUBJECTS
+
+_CHOICE_LABELS = ("A", "B", "C", "D", "E")
+
+_ANSWER_REGEXES = (r"\(?([A-D])\)?",)
+
+_COT_SAMPLING = SamplingParams(
```
mgmorgan23, line 41: I believe in oe-eval all post-training evals were actually run with an override of 32768 max_tokens – Saumya would know for sure (not blocking, we can always run with an override)

src/olmo_eval/evals/tasks/mmlu_cot.py
```diff
@@ -0,0 +1,155 @@
...
+
+def _description(subject: str) -> str:
+    return (
+        f"The following are multiple choice questions about {_subject_text(subject)}. "
+        "Summarize your reasoning concisely, then conclude with "
+        "'Therefore, the answer is: X' where X is one of A, B, C, or D.\n\n"
+    )
+
+
+def _mcq_query(question: str, choices: list[str]) -> str:
```
mgmorgan23, line 60: I think MCQAChatFormatter from the formatters has the same functionality if we want to stay centralized

src/olmo_eval/evals/tasks/mmlu_cot.py
```diff
@@ -0,0 +1,155 @@
...
+    answer = extract_answer_with_format(
+        text,
+        answer_regexes=_ANSWER_REGEXES,
+        answer_regexes_templates=OLMO_3_ANSWER_REGEX_TEMPLATES,
+    ).answer
+    return re.sub(r"\(|\)", "", answer)
+
+
+@dataclass(frozen=True, slots=True)
+class MMLUCoTExactMatchScorer(Scorer):
```
mgmorgan23: To be consistent with other tasks I believe `MultipleChoiceScorer(Scorer)` is already a case insensitive scorer – I think to ensure the extraction follows MMLU rules you can just set the class variable answer_extractor to be the correct function.

src/olmo_eval/evals/tasks/mmlu_cot.py
```diff
@@ -0,0 +1,155 @@
...
+            },
+        )
+
+    def format_request(self, instance: Instance) -> LMRequest:
+        return LMRequest(
+            request_type=RequestType.CHAT,
+            messages=({"role": "user", "content": instance.question},),
+        )
+
+    def extract_answer(self, output: LMOutput) -> str | None:
```
mgmorgan23: I think if you set the `answer_extractor` variable this is not necessary

mgmorgan23:
I left a few comments but this overall looks good to me! I think running a couple of full parity tests is definitely worth it – we could make it cheaper by running some of the baselines (both for reasoning and non-reasoning models) from the Olmo3 paper and just comparing against those numbers – I can help find exact beaker jobs and their metrics if useful!

tests/evals/tasks/test_mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Converted to plain pytest functions (with `parametrize`) in 3cfc1d1.

src/olmo_eval/evals/tasks/mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Agreed they overlap, but they aren't interchangeable: `mcq.py` is olmo-eval's native cascade (BBQ/WMDP/MedQA; `ANSWER: X` first, 4 patterns), while `answer_format.py` mirrors oe-eval's `extract_answer` (12 templates + a format score) and is what the parity ports need — the same text can extract differently under the two orders. Added cross-referencing module docstrings to both in 3cfc1d1. Standardizing would move numbers on one side, so leaving that as a follow-up.

src/olmo_eval/evals/tasks/mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

You're right about the number: the Olmo 3 release scripts pass `--oe_eval_max_length 32768` (open-instruct `scripts/train/olmo3/7b_instruct_rl.sh`, `olmo-hybrid/7b_release_evals.sh`), and `oe-eval.sh` feeds that one value to both the model `max_length` and `max_gen_toks` with `truncate_context: false` — so the effective budget was "until the context is full". `max_tokens=None` is olmo-eval's expression of exactly that (vLLM natively; `OlmoCoreProvider._resolve_max_tokens`), so serving with `max_model_len=32768` reproduces the published budget with no per-task override. Same convention as `gpqa:cot`, `popqa`, `omega_500`.

src/olmo_eval/evals/tasks/mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Close but not byte-identical: `MCQAChatFormatter` renders `(A) text` blocks separated by blank lines, while oe-eval's `hamish_zs_reasoning` MMLU CoT (`make_mcq_prompt`) renders `Question: …\n A. text\n B. text\n` under the per-subject description — which is what this task reproduces; prompts were verified byte-identical to oe-eval on 3,207 docs across 7 subjects. Switching would break that, so kept the task-local builder. A centralized option for oe-eval's letter style would be a reasonable follow-up.

src/olmo_eval/evals/tasks/mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Done in 3cfc1d1: `answer_extractor = _extract_letter` and `AccuracyMetric(name="exact_match", scorer=MultipleChoiceScorer)`; the custom scorer is gone. One asymmetry to flag: `gpqa:cot` records `answer_format_correct` in output metadata via its scorer (asked for in #309), and the extractor contract only sees text, so `mmlu:cot` won't — happy to align the two in a follow-up if we want that signal here.

src/olmo_eval/evals/tasks/mmlu_cot.py
abhishekraok: Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

Removed in 3cfc1d1.

abhishekraok:
> I left a few comments but this overall looks good to me! I think running a couple of full parity tests is definitely worth it – we could make it cheaper by running some of the baselines (both for reasoning and non-reasoning models) from the Olmo3 paper and just comparing against those numbers – I can help find exact beaker jobs and their metrics if useful!

Thanks for the review. I found the paper's oe-eval mmlu:cot runs in ai2/tulu-3-results (…-PAPER-run{1,2,3}-on-mmlu-cot-hamish_zs_reasoning_deepseek): Olmo-3-7B-Think (fixed-systemprompt runs), Olmo-3-32B-Instruct, and Olmo-3.1-32B-Think, 3 seeds each. For Olmo-3-7B-Instruct I only see the SFT and DPO stage runs — was the final RL model's mmlu:cot run under a different name, or should I use the DPO ones? If the Think and 32B jobs look right to you, I'll run the paired check against those.

abhishekraok:

**Paired parity against the paper's oe-eval run** — Olmo-3-7B-Think, `mmlu:cot::hamish_zs_reasoning_deepseek`, PAPER run1 (Beaker results `01KAD4QTVWCR9RZSS48YZ52040`), all 57 subjects, 14,042 instances:

| check | result |
|---|---|
| prompts vs. oe-eval's logged requests | 0 mismatches |
| gold letters | 0 mismatches |
| scorer verdicts on oe-eval's stored generations | **14,042 / 14,042** |
| suite score (macro over subjects) | 0.764283 both |

One real gap surfaced: the paper runs use oe-eval's model-level `process_output: r1_style`, which strips everything through `</think>` before scoring. Scoring the raw outputs instead agreed on only 13,910/14,042 (132 flips). `3f777aa` applies the identical `extract_think_answer` inside this task's extractor (no-op without `</think>`), restoring 14,042/14,042 from raw outputs. The sibling ports (gpqa:cot, popqa, omega_500) have the same gap — filed #348.

Also: the PAPER runs' model config has `max_length: 32768` with `max_gen_toks: 131072`, i.e. generation bounded by context — what `max_tokens=None` does here.

src/olmo_eval/evals/tasks/mmlu_cot.py
mgmorgan23: Makes sense! We probably should discuss unified task formatting as a follow up at some point

mgmorgan23:
Those parity checks look right to me! LGTM to merge

mgmorgan23 (approved):

PR #345, opened 2026-09-02
Upgrade torch to 2.13

gregorybchris:
## Description

Upgrades the evaluation stack to **PyTorch 2.13** and **vLLM 0.28.0**, and moves the default Docker image to **CUDA 13.0**.

The torch version is not really a free choice here: vLLM pins torch exactly, and **no vLLM release has ever pinned torch 2.12** (0.19.x → 2.10, 0.20.0–0.26.0 → 2.11, 0.27.0+ → 2.13). So 2.13 is the next reachable version after our current 2.10, and it comes with vLLM 0.28.

**Dependency changes**

- `vllm[runai]` 0.19.1 → 0.28.0
- `ai2-olmo-core[torchao,transformers]` 2.4.0 → 2.6.0
- `click~=8.3.2` → `>=8.4.2,<9` — vLLM 0.28 pulls `huggingface-hub>=1.27`, which requires `click>=8.4.2`; the old pin made the graph unsatisfiable
- Relocked: torch 2.10.0 → 2.13.0, torchvision 0.25.0 → 0.28.0, torchaudio 2.10.0 → 2.11.0, triton 3.6.0 → 3.7.1, transformers 5.7.0 → 5.16.1, flashinfer 0.6.6 → 0.6.16.post3, xgrammar 0.1.34 → 0.2.3

`transformers` was floated deliberately rather than left at the lock's 5.7.0. vLLM 0.26 moved Olmo/Olmo2 onto the Transformers modeling backend and vLLM 0.28 is tested against transformers 5.15, so staying eight minors behind what vLLM tests would be off the supported path for exactly the models we evaluate.

**Image / CI changes**

- Default image is now CUDA 13.0.3 + torch 2.13.0 (was 12.8.1 + 2.10.0)
- Dropped 12.8.0 / 12.8.1 from `SUPPORTED_CUDA_VERSIONS` and added 13.0.3. **PyTorch 2.12 removed the cu128 wheel build entirely** — torch 2.13 publishes cu126, cu129, cu130 and cu132 only, so a 12.8.x image can no longer install a matching torch
- CI's GPU smoke check moved off the stale `nvidia/cuda:12.0-base` tag to `nvidia/cuda:13.0.3-base-ubuntu24.04`
- README build docs updated to match the new supported set

**No source changes were needed.** olmo-eval touches vLLM only through its public API (`LLM`, `SamplingParams`, `RequestOutput`) and the `vllm serve` CLI flags, all of which are unchanged across 0.19 → 0.28. I checked the vLLM release notes for 0.20 through 0.28 against our usage; the breaking changes in that range (PagedAttention removal, the unified `Parser.parse()` interface, `LLM.reward` deprecation, PoolerConfig `logit_bias`/`logit_scale` renames, `calculate_kv_scales` and `override_attention_dtype` removal, the `CUDA_VISIBLE_DEVICES`/`device_ids` change, bitsandbytes moving out-of-tree) land on surfaces this repo does not use.

## Type of Change

- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update

Breaking in two ways: CUDA 12.8 images can no longer be built, and the default image now requires a CUDA 13 capable driver.

## Testing

- [x] Unit tests pass locally (`pytest tests/ --ignore=tests/integration/`) — **2149 passed, 8 skipped**
- [ ] Integration tests pass (if applicable) — **not run.** `tests/integration/test_vllm_provider.py` needs a GPU, and the new image needs a CUDA 13 driver, so this is gated on the driver rollout. Should be run before merge.
- [ ] New tests added for new functionality — none; this is a dependency and image upgrade with no source changes.

Also verified locally: `ruff check src/ tests/` and `ruff format --check src/ tests/` clean, `ty check src/ alembic/` unchanged from main, and `uv lock --check` clean.

## Checklist

Notes on the last two: the only test warning is the pre-existing `pkg_resources` deprecation from `syllapy` (transitive via `ifbench`), and `ty` still reports the same single pre-existing unused-`type: ignore` in `harness/sandbox/executor.py` that main reports — no new diagnostics. All upgraded dependencies are published releases on PyPI.

## Risks for reviewers

1. **Integration tests are unrun.** The vLLM provider path is exercised only by mocks in the unit suite. A real `vllm serve` startup and generation pass on CUDA 13 is the thing that would actually catch a 0.19 → 0.28 behavioural regression.
2. **torchao is downgraded, 0.17.0 → 0.15.0.** Not our choice: `ai2-olmo-core` 2.6.0 hard-pins `torchao==0.15.0`, which is a December 2025 release now paired with torch 2.13. Worth relaxing upstream in OLMo-core.
3. **OLMo-core has no torch 2.13 CI coverage.** Its CI and Docker builds currently extend through torch 2.12 / CUDA 13.0.
4. **Driver dependency.** Anything still on a CUDA 12 driver cannot run the new default image. There is no fallback image: 12.8.x can no longer produce a matching torch build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,7 +281,7 @@ jobs:
           docker --version
           docker compose version
           nvidia-smi
-          docker run --rm --gpus all nvidia/cuda:12.0-base nvidia-smi
+          docker run --rm --gpus all nvidia/cuda:13.0.3-base-ubuntu24.04 nvidia-smi
 
       - name: Run vLLM integration tests with Docker
         run: |
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,14 +5,14 @@
 #
 # Build:
 #   ./scripts/build_image.sh
-#   ./scripts/build_image.sh --cuda-version 12.8.0
+#   ./scripts/build_image.sh --cuda-version 12.9.1
 #   ./scripts/build_image.sh --platform linux/amd64
 #
 # Tags: cu{cuda}-trc{torch}-{arch}
-# Example: cu1281-trc2100-amd64
+# Example: cu1303-trc2130-amd64
 
-ARG CUDA_VERSION=12.8.1
-ARG TORCH_VERSION=2.10.0
+ARG CUDA_VERSION=13.0.3
+ARG TORCH_VERSION=2.13.0
 ARG PYTHON_VERSION=3.12
 ARG INSTALL_CHANNEL=whl
 ARG GIT_COMMIT=""
--- a/README.md
+++ b/README.md
@@ -1470,7 +1470,7 @@ Images are tagged with CUDA and PyTorch versions: `cu{version}-trc{version}-{arc
 ./scripts/build_image.sh
 
 # Specific CUDA + PyTorch version
-./scripts/build_image.sh --cuda-version 12.8.1 --torch-version 2.10.0
+./scripts/build_image.sh --cuda-version 12.9.1 --torch-version 2.13.0
 
 # Production build
 ./scripts/build_image.sh --platform linux/amd64
@@ -1479,7 +1479,7 @@ Images are tagged with CUDA and PyTorch versions: `cu{version}-trc{version}-{arc
 ./scripts/build_image.sh --help
 ```
 
-**Supported CUDA versions**: 12.6.1, 12.8.0, 12.8.1, 12.9.1
+**Supported CUDA versions**: 12.6.1, 12.9.1, 13.0.2, 13.0.3 (PyTorch 2.12+ no longer ships a cu128 build, so 12.8.x is unsupported)
 **PyTorch version**: Configurable via `--torch-version`
 **Configuration**: See `scripts/build_config.sh`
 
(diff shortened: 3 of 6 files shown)


abhishekraok:
I assume these will be merged when the machines are upgraded to Cuda 13.

Claude had one major finding:
"No source changes were needed" is wrong: BEAKER_DEFAULT_IMAGE and BEAKER_SANDBOX_IMAGE still point at ai2-tylerm/olmo-eval-cu1281-trc2100-amd64[-sandbox] (common/constants/infrastructure.py:22,25). Those are the images every olmo-eval beaker launch uses unless overridden (cli/beaker/launch.py:535-537), and they're torch 2.10 / CUDA 12.8.1 — the exact combination this PR declares unbuildable. A new image has to be built, pushed under a namespace that will outlive its owner, and these constants updated in the same PR. Without that, the default launch path hits finding 1's first bullet on day one.

Another warning: the image rebuild has to happen after the driver rollout is confirmed and from this branch's lock, then the constants change goes into this PR before merge — otherwise the sequencing lands the lock ahead of the image and the default launch path breaks on day one.

PR #347, opened 2026-09-02
Add vllm image support

donovanclay:
## Description

Teaches the inline `VLLMProvider` to send images, adds an exporter so it can read our own
OLMo-core checkpoints, and adds prompt-convention knobs (`prompt_style`, `image_mode`)
needed to evaluate non-Molmo models and to attribute error between perception and
knowledge.

Measured throughput on one 80 GB card:

| workload | HF | olmo_core | vLLM | vLLM vs olmo_core |
|---|---:|---:|---:|---:|
| short answers (`chart_qa`, cap 64 tok) | 2.0 it/s | 5.5 it/s | 11.9 it/s | **2.2×** |
| long decode (`charxiv_reasoning`, 1024 tok) | — | 0.8 it/s | 3.8 it/s | **4.8×** |

Those rows are Qwen3-VL-4B. On **Molmo2-4B** the HF provider is materially slower —
@jason718 measured ~0.64 req/s over full `chart_qa` (n=1920), where end-to-end wall clock
was vLLM 6m50s, `olmo_core` 12m37s, HF 47m43s, i.e. vLLM 7.0× HF and 1.85× `olmo_core`.
Read the HF column as model-dependent.

The gain tracks how much the task actually *decodes*; `max_tokens` is a ceiling, not a
target, so short-answer suites show much less of it. Note `olmo_core` is **not** the slow
path — it batches, and beats the HF provider, whose `_generate_multimodal` is structurally
batch-size 1.

<details>
<summary><b>Full batch-size / memory sweep (15 cells) and methodology</b></summary>

**Setup.** One GPU, Molmo2-4B architecture throughout, 200 `chart_qa:test` instances,
identical crop budget, greedy. `olmo_core` reads the v9 DCP checkpoint and the other two
read `allenai/Molmo2-4B` — same architecture, different weights, which is immaterial for
throughput since speed depends on shape rather than weight values. Peak GPU memory sampled
every 2 s per cell.

Two venvs are required (`ai2-olmo-core` and vLLM fight over the torch pin), so a 16th cell
re-runs the identical HF config in the *other* venv as a control: 2.0 vs 2.1 it/s, i.e. the
environment is not a confound.

| provider | knob | it/s (cap 64) | it/s (cap 512) | peak MiB |
|---|---|---:|---:|---:|
| vLLM | `max_num_seqs=256` | **11.9** | **12.2** | 65,591 |
| vLLM | `max_num_seqs=64` | 11.1 | 12.1 | 65,705 |
| vLLM | `max_num_seqs=8` | 9.0 | 10.6 | 62,279 |
| olmo_core | `batch_size=32` | 5.5 | 5.6 | 47,637 / 51,589 |
| olmo_core | `batch_size=8` | 5.3 | 5.4 | 27,045 |
| olmo_core | `batch_size=1` | 3.5 | 3.6 | 22,109 |
| HF | none — bs=1 structural | 2.0 | 2.1 | **11,461** |

**Findings.**
- **HF is the most memory-efficient by a wide margin** (11.5 GB, constant) and the slowest.
  It is the only option when VRAM is tight; it has no batch-size knob to trade.
- **`olmo_core` scaling stops paying at `batch_size=8`.** 1→8 buys +51% throughput for
  +5 GB; 8→32 buys **+4% for +21 GB**. Use 8–16.
- **vLLM's memory is flat across `max_num_seqs`** because it pre-allocates the KV cache from
  `gpu_memory_utilization` (0.8). The footprint is a *choice*, not a consequence of the
  batch setting — and it is why vLLM is OOM-killed at engine init on 48 GB cards (L40S,
  A6000) with no traceback, surfacing only as
  `Expected 1 inference worker(s) to initialize ... got 0`.
- **The cap-64 and cap-512 columns are nearly identical** because `chart_qa` answers are
  short, so generation stops well before either cap. This matrix therefore measures the
  *prefill-bound* regime only; the long-decode row in the summary table comes from a
  separate `charxiv_reasoning` run at 1024 tokens, where per-instance throughput is 0.8 vs
  3.8 it/s (output-token rate ~32 vs ~220 tok/s, but vLLM generated more tokens per batch,
  so 4.8× items/sec is the honest wall-clock figure).
- Load time also differs: `olmo_core` spends 243–264 s reconstituting 100+ `.distcp` shards
  versus 106–130 s for vLLM reading one safetensors file. At eval scale this amortizes, so
  the inference ratio is what matters.

</details>

### 1. Images in `VLLMProvider`

`_format_prompt` returns vLLM's dict form — `{"prompt": ..., "multi_modal_data": {"image":
[...]}}` — when a request carries images, rendering placeholders through the tokenizer chat
template. No `AutoProcessor` needed: the tokenizer-level template emits the same
placeholders as the processor-level one (verified on Qwen3-VL) and vLLM expands them. The
message shape matches `HuggingFaceProvider._build_chat_messages` and
`OlmoCoreVlmProvider._chat_text` so scores stay comparable across providers.

New kwargs: `max_images` (raises `limit_mm_per_prompt` off vLLM's default of 1),
`strip_reasoning`, `chat_template_kwargs`.

Three paths now fail loudly rather than silently dropping images or traces:
- `add_bos_token=False` + images — the pre-tokenized path cannot carry `multi_modal_data`.
- `logprobs()` + images — scores raw token sequences with nowhere to attach them.
- more images than `max_images` — vLLM raises for the *entire* `generate` call, so one
  oversized prompt fails its whole chunk. Two MMMU-Pro instances carry 35 images and cost
  128 of 5190 before this check.

### 2. Running our own checkpoints through vLLM

`tools/olmo_core_to_hf/export.py` converts an OLMo-core DCP step directory into a
released-Molmo2 HF directory; vLLM 0.19.1 registers `Molmo2ForConditionalGeneration`, so
the export loads directly. The weight mapping is not reimplemented — it calls OLMo-core's
`multimodal_lm_state_dict_to_hf`, already documented as the inverse of the loader used to
*read* released Molmo2 weights — and model assembly reuses the provider's own path.

Verified on 200 `chart_qa:test` instances:

| leg | score | text match vs DCP | load | infer |
|---|---:|---:|---:|---:|
| `olmo_core` on DCP (reference) | 72.50 | — | 243.2 s | 38.8 s |
| `hf` on export | 72.50 | 98.0% | 23.9 s | 99.0 s |
| `vllm` on export | 72.00 | 95.5% | 106.4 s | 16.8 s |

Independently reproduced by @jason718 on the released `Molmo2-4B/step2000` at full
`chart_qa` (n=1920), with a stronger proof than a score comparison: the export is **bitwise
equal to `allenai/Molmo2-4B` cast to bf16, 706/706 tensors**. `--verify` now does that check
(no GPU, no inference). Matching eval scores are weaker than they look — 72.50 vs 72.50 at
98% text agreement means some outputs differed and netted to zero on a lossy metric.

His n=1920 run also shows **vLLM is the *more* faithful of the two export paths**, not the
riskier one: raw-text agreement with the `olmo_core` DCP leg is 97.14% for vLLM against
96.30% for HF, despite HF and vLLM reading byte-identical weights, and HF lands furthest
from mm_olmo's own reference dump on all three metrics.

One caveat on his numbers: both export legs sit *below* the DCP leg on every metric — a
consistent direction, not scatter. A likely cause is precision rather than implementation:
the `olmo_core` leg runs `dtype=float32` with `autocast_dtype=bfloat16` (fp32 master weights,
bf16 compute) while both export legs are pure bf16. His fp32-vs-bf16 control cannot see this,
since vLLM casts to bf16 at load either way. Testable with one `olmo_core` run at
`dtype=bfloat16`.

### 3. `prompt_style`: our defaults are mm_olmo's SFT convention

Qwen3-VL-4B-Instruct through this harness versus its published numbers:

| task | `molmo` (default) | `:neutral` | published | Δ |
|---|---:|---:|---:|---:|
| ChartQA:test | 13.88 | **84.24** | 84.6 | −0.36 |
| DocVQA (val) | 0.00 | **94.52** | 95.3 | −0.78 |
| VQAv2 | 6.45 | **82.25** | 81.7 | +0.55 |
| TextVQA | 11.15 | **82.89** | 81.0 | +1.89 |
| InfoQA (val) | 0.00 | **78.81** | 80.3 | −1.49 |
| RealWorldQA | 43.79 | **72.29** | 70.9 | +1.39 |
| AI2D:test | 83.23 | 83.16 | 84.1 | −0.94 |
| MMMU | 49.67 | 49.56 | 67.4 | −17.84 |

Asked `chart_qa: How many food item is shown in the bar graph?` with no answer-format
instruction, the model replies "There are three bars shown in the chart.", which exact-match
against "3" scores 0. AI2D was always fine because it carries no style tag, has an explicit
instruction, and scores letters. **Not a budget problem:** sweeping MMMU's cap with the
prompt untouched moved 49.67 → 50.11 → 50.89 → 52.22 at 12/64/512/2048.

`prompt_style="neutral"` drops the tag, appends the standard short-answer instruction, and
strips MMMU's `<image N>` markers (stray tokens here, since images arrive via the chat
template; `mmmu_pro` already resolves them, `mmmu` never did). **The default stays `molmo`**
— our checkpoints were SFT'd on those tags and need them.

`prompt_style="cot"` + `mmmu:cot` handles MMMU, which is a reasoning benchmark: it permits
reasoning and requires a final `Answer:` line, paired with `extract_cot_answer`, which reads
the **last** marker (shared `clean_prediction` takes the first, which on a trace scores a
discarded intermediate guess). Wired via `TaskConfig.answer_extractor`, so
`clean_prediction` is untouched and no existing Molmo number shifts. MMMU goes
49.56 → 57.89 (2048 tok) → 60.11 (8192 tok).

### 4. `image_mode`: perception vs knowledge attribution

`TaskConfig.image_mode` = `real` | `none` | `caption` (+ `caption_source`), read by
`ImageQATask.format_request` and `MmmuProTask.format_request`. `tools/oracle_captions/`
generates GPT-4o captions, deduped by image content (`charxiv_descriptive` is 4000
instances over 1000 figures; `charxiv_reasoning` reuses all 1000 from cache for zero API
calls).

Three ways this could report a meaningless number instead of failing, all guarded: a
missing caption must not fall back to text-only; an unknown mode must not be treated as
`real`; and `mmmu_pro`'s **vision** setting refuses imageless modes, because its question is
rendered *into* the screenshot and the official parser falls back to a seeded random choice
— it would report ~25% rather than error.

### 5. Bugs found along the way

- **`HuggingFaceProvider` never applied the chat template on its text-only path.** It read
  `request.prompt`, but a CHAT request carries `messages` and leaves `prompt` empty. Zero
  tokens → `cannot reshape tensor of 0 elements into shape [1, 0, -1, 128]`, and the job
  reported **SUCCEEDED with n=0**.
- **`TASK_CONFIG_FIELDS` was a hardcoded allowlist drifted from the dataclass**, rejecting
  valid `-o` overrides with a message about flag placement. Four fields were already
  unreachable: `max_length`, `answer_extractor`, `output_score_aggregation`, `sandbox_env`.
  Now derived from `dataclasses.fields(TaskConfig)`.
- `strip_reasoning` moved to `olmo_eval.inference.reasoning` so both providers share it, and
  `runners/io/builders.py` persists `metadata["reasoning"]` so traces survive for error
  analysis.

## Type of Change

- [ ] Breaking change
- [ ] Documentation update

## Testing

152 tests pass across `tests/providers/`, `tests/cli/test_utils.py` and the new
`tests/evals/tasks/test_image_mode_ablations.py`. The vLLM image tests use the existing
`FakeLLM`/`FakeTokenizer` monkeypatch pattern, so they need no vLLM install.

**HF-vs-vLLM score parity**, the gate that matters. Full sweep, 500 instances per task,
both providers in the same venv, identical prompt style and token budget, differing only in
`provider.kind` (Qwen3-VL-4B-Instruct):

| task | HF | vLLM | Δ | text match | extracted match |
|---|---:|---:|---:|---:|---:|
| AI2D | 86.39 | 86.13 | −0.26 | 496/500 | 496/500 |
| ChartQA | 73.80 | 73.60 | −0.20 | 482/500 | 482/500 |
| DocVQA | 94.95 | 94.95 | +0.00 | 496/500 | 490/500 |
| InfoQA | 77.34 | 78.03 | +0.69 | 483/500 | 483/500 |
| TextVQA | 84.07 | 83.80 | −0.27 | 486/500 | 486/500 |
| VQAv2 | 80.00 | 80.27 | +0.27 | 491/500 | 491/500 |
| RealWorldQA | 73.60 | 73.20 | −0.40 | 496/500 | 496/500 |
| MMMU | 49.40 | 49.20 | −0.20 | 437/500 | 435/500 |

**mean |Δ| = 0.29, max 0.69, signs mixed** (4 negative, 3 positive, 1 zero; mean signed Δ
−0.05) — the signature of kernel-level numerical noise rather than a systematic difference.
The magnitudes match what the disagreement rate predicts: ~96% text agreement means ~20
instances differ, and coin-flip resolution of those moves the score 0–0.7 points. MMMU has
the lowest agreement (87%), as expected for the task with the longest generations.

Earlier point checks agree: MMMU on 900 instances gave 49.67 for both providers exactly, and
MMMU-Pro on 64 gave 27.92 for both. The Molmo2 export's ChartQA −0.50 (72.50 vs 72.00) sits
inside this distribution.

On the 64-instance MMMU-Pro subset raw *text* agreement is 45/64, which is benign: **zero**
mismatches occur where both providers produced a clean letter, and both ramble on exactly
27/64. Every mismatch has at least one side truncated by that task's 64-token cap, after
which the official parser scrapes a letter out of partial reasoning.

## Notes for reviewers

- `uv sync` needs **`--extra vllm` explicitly**. The `vllm` extra is in a `conflicts` set
  with `openhands`, so uv resolves a fork that drops it even though the `vllm` group is in
  `default-groups` — which is why `.venv` reports `vLLM NOT INSTALLED`.
- Qwen3-VL needs `-o provider.max_model_len=8192` (or more): it declares
  `max_position_embeddings=262144`, and vLLM sizes its KV cache to that, demanding 36 GiB.
  High-resolution document images also exceed 8192 *visual* tokens, so document tasks want
  32768 — which then requires an 80 GB card.
- **Molmo2 needs two flags that are easy to miss.** `-o provider.trust_remote_code=true`
  (it ships remote code; the default `False` crashes `hf` and `vllm` at worker init) and, for
  vLLM, `-o provider.max_num_batched_tokens=32768` — Molmo2's crops emit 31,872 mm tokens per
  item against vLLM's 8192 default, giving
  `Chunked MM input disabled but max_tokens_per_mm_item (31872) is larger than
  max_num_batched_tokens (8192)`.
- `--reference` is **required** when exporting a released mm_olmo checkpoint: their
  `config.yaml` carries no `model_id` for the exporter to default from.
- Do **not** install `ai2-olmo-core` into a vLLM env; they fight over the torch pin. Note the
  pinned `ai2-olmo-core==2.4.0` has no `olmo_core.nn.vision`, so multimodal work needs the
  vision branch.
- `doc_qa:test` and `info_qa:test` are **unlabelled** (labels live on the RRC server) and
  score 0.00 for any model while reporting success. Use the `val` defaults.
- This branch also carries two earlier commits that were on local `vision` but not
  `origin/vision` (`91f9f78` SplitVocabEmbedding `full_weight()`, `aa6e370` OPENAI_API_KEY
  strip). Happy to rebase them out if they should land separately.

## Checklist


jason718:
## Independent verification on a real Molmo2-4B checkpoint

I reproduced the export and parity claims against `Molmo2-4B/step2000` (the released mm_olmo DCP on weka) rather than the v9 checkpoint used in the description. Short version: **the science holds up, and I got a stronger proof of the export than the PR provides — but the documented workflow does not run as written.**

Everything below ran locally on one A100-80GB. **No Beaker jobs were involved, so there are no job IDs to cite.** Full repro at the bottom.

---

### 1. The export is provably correct (stronger than the PR's own evidence)

The description argues correctness from `hf` reproducing `72.50` exactly. That's weaker than it reads: 72.50 vs 72.50 at 98.0% text match means 4 of 200 outputs differed and happened to net to zero on a lossy metric.

There's a decisive test that needs **no GPU and no inference** — compare tensors directly:

```
export vs released allenai/Molmo2-4B
  tensors: 706 / 706 present, 0 missing, 0 extra, 0 shape mismatches
  export == released.to(bfloat16), bitwise: 706/706 tensors, 0 differing elements
```

`multimodal_lm_state_dict_to_hf` is exactly right, and so is the DCP read. Worth adding as a `--verify` mode — it's ~10 min on CPU and strictly better than an eval-score comparison.

Two notes from this:
- The released model ships **fp32**; the export defaults to `--dtype bfloat16`. **This turns out to be harmless** (see §2), but the copied `config.json` carries `torch_dtype: null`, so nothing in the output directory records the dtype it was actually written in.
- This checkpoint is **`mm_olmo_dcp`** (`config.yaml` + DCP), not the `olmo_core_dcp` the docstring documents. The export handles it fine, but `model_id` is absent from mm_olmo configs, so **`--reference` is mandatory for every released checkpoint** — and the usage example omits it. It fails loudly, which is right, but the example should show it.

### 2. Three-way parity: it passes

chart_qa validation, **full split n=1920**, greedy, identical task config, only the provider varies:

| metric | mm_olmo ref | olmo_core (DCP) | hf (export) | vllm (export) | vllm (released fp32) |
|---|---:|---:|---:|---:|---:|
| **relaxed_correctness** (primary) | 0.8380 | 0.8375 | 0.8333 | 0.8354 | 0.8354 |
| scifi_relaxed_correctness | 0.8516 | 0.8495 | 0.8458 | 0.8474 | 0.8474 |
| em | 0.7490 | 0.7510 | 0.7448 | 0.7474 | 0.7474 |

Max spread **0.42 pts** (8 instances of 1920); all three land within 0.5 pts of mm_olmo's own reference dump. This is the leg the description only covers once (chart_qa, n=200) — it holds at 9.6× that sample size on a real released checkpoint.

**The bf16 default is a non-issue.** vLLM on the bf16 export vs vLLM on the released fp32 weights: **identical scores** (0.000 on all three metrics), 100% per-instance correctness agreement, 99.79% raw text. vLLM casts to bf16 at load either way, so fp32 storage buys nothing.

### 3. Two things worth correcting in the write-up

**The deltas are one-directional.** Both export legs sit *below* the DCP leg on every metric — not the mixed-sign scatter the HF↔vLLM table reports. Small, but a bias rather than noise. Since dtype is ruled out, the residual is implementation, and the description's "consistent with kernel-level numerics" framing doesn't quite cover it.

**HF is the least faithful of the three, not vLLM.** Raw-text agreement:

```
  olmo_core vs vllm   97.14%     <-- different weights, different impl
  olmo_core vs hf     96.30%
  hf        vs vllm   96.35%     <-- identical weights!
```

vLLM agrees with olmo_core *more* than HF does, despite HF and vLLM reading byte-identical weights, and HF is furthest from the mm_olmo reference on all three metrics. If anything the path this PR adds is **more** faithful than the existing one — which is a better argument for the PR than the one currently being made.

**Throughput** (same 1920 instances, wall clock end-to-end incl. load): `vllm` **6m50s**, `olmo_core` **12m37s**, `hf` **47m43s**. vLLM = **7.0× HF**, **1.85× olmo_core**. HF is materially worse than the description implies — I measured ~0.64 req/s on Molmo2 (64-request batches at ~100s) against the quoted 2.0 it/s on Qwen3-VL.

---

### 4. Blockers: the documented verify flow doesn't run

`export.py` signs off with:

> VERIFY before use: run the same task via `provider.kind=olmo_core` on the DCP and `provider.kind=vllm` on this export, and require the scores to agree.

Following that verbatim fails three times. Two are doc gaps; one is a real design gap.

**(a) `-o provider.trust_remote_code=true`** — Molmo2 ships remote code; the default `False` crashes both `hf` and `vllm` at worker init.

**(b) `-o provider.max_num_batched_tokens=32768`** — vLLM dies with:
```
Chunked MM input disabled but max_tokens_per_mm_item (31872) is
larger than max_num_batched_tokens (8192).
```
Molmo2's crops produce 31,872 mm tokens per item. The "Notes for reviewers" section documents a `max_model_len` gotcha for Qwen3-VL but not this one — and this is the one that hits **our own checkpoints**, i.e. the headline use case.

**(c) `-o provider.multimodal=true`** — this one should be fixed in code, not documented around.

`HuggingFaceProvider.__init__` auto-detects multimodal via `is_olmo_core_hf_export()`, which returns `detect_checkpoint_format(...).format == "olmo_core_unsharded"` — i.e. it tests for the **pre-existing** export format (`olmo_core_config.json` + OLMo-core-named weights). This PR's `tools/olmo_core_to_hf/export.py` produces a **different artifact**: a real HF `config.json` with HF-named weights. So detection returns `False`, the provider falls through to `_init_text`, and dies with:

```
Unrecognized configuration class Molmo2Config for this kind of AutoModel: AutoModelForCausalLM
```

The repo now has two incompatible things both called an "OLMo-core export," and the provider only knows about the older one. Suggest either teaching `is_olmo_core_hf_export` (or a sibling predicate) to recognize the new layout, or having the export write a marker the provider can key on. The failure mode is the provider silently selecting the wrong code path, which is exactly the class of bug this PR is otherwise good about closing.

---

### 5. Also worth a look (from reading the diff, not measured)

Ranked; happy to open these separately if useful.

1. **`base.py` `__post_init__`** — the new dict→`SamplingParams` coercion does `SamplingParams(**overrides)`, building a fresh object from the override keys alone. `-o sampling_params.max_tokens=2048` on humaneval drops all 10 `stop_sequences`. Doesn't affect any vision number (every image-QA task sets `temperature=0.0` and no `stop_sequences`), but it turns a loud `unhashable type: 'dict'` into a silent wrong score for text tasks. Real fix is in `runners/asynq/runner.py:_build_task_overrides`, which should route dotted `sampling_params.*` into `sampling_overrides`.
2. **`mmmu:cot` inherits `max_tokens=12`** — the registration comment says "Needs a large max_tokens (~2048)" but doesn't set it, so the default invocation of the feature truncates before the required `Answer:` line.
3. **`limit_mm_per_prompt` is injected unconditionally** (`max_images` defaults to 8), including for text-only models where vLLM rejects it outright.
4. **The `max_images` guard reads `self._max_images`, not the resolved engine limit** — an explicit `limit_mm_per_prompt` makes the guard permissive, which is the case it exists to catch.
5. **Only 1 of 4 `format_request` implementations honors the new knobs.** `render_question` is called from `image_qa_base.py` only; `mmmu_pro.py` calls `resolve_image_mode` but not `render_question`, and `pointing_base.py` / `multi_image_base.py` call neither — so `-o image_mode=none -t blink` runs the full image benchmark and reports it as a text-only ablation.
6. **`TaskConfig.to_dict()` omits `image_mode` / `caption_source` / `prompt_style`**, so an ablation run serializes byte-identically to the real benchmark run.
7. `ruff format --check` fails on `tests/evals/tasks/test_image_mode_ablations.py` (the hand-wrapped `parametrize` list). Note CI is gated to `pull_request: branches: [main]`, so nothing but GitGuardian runs on vision-targeted PRs — I ran the full suite locally and it's otherwise green (2105 passed, 9 skipped; `ruff check` and `ty check` clean).

---

<details>
<summary><b>Full repro</b></summary>

**Environment.** One A100-SXM4-80GB (driver 570.124.06), weka mounted locally. PR head `06b6a96`. Two venvs are required — `ai2-olmo-core` and vllm fight over the torch pin:

```bash
# venv-oc: export + olmo_core_vlm + hf legs
uv sync --frozen --no-group vllm --extra olmo_core --extra hf
uv pip install "ai2-olmo-core[torchao,transformers] @ git+https://github.com/allenai/OLMo-core.git@vision"
#   -> ai2-olmo-core 2.5.0 (commit 8018f472), torch 2.10.0+cu128, torchao 0.15.0, transformers 5.7.0
#   NB: the pinned release ai2-olmo-core==2.4.0 has NO olmo_core.nn.vision, so the vision
#       branch override is required for any multimodal checkpoint.

# venv-vllm: vllm leg
uv sync --frozen --extra vllm --extra hf
#   -> vllm 0.19.1, torch 2.10.0+cu128, transformers 5.7.0
```

**Export** (~10 min, CPU-bound, ~27 GB RSS):
```bash
python tools/olmo_core_to_hf/export.py \
  --checkpoint /weka/oe-training-default/mm-olmo/released-models-molmo2-1225/Molmo2-4B/step2000 \
  --out <out>/export-molmo2-4b \
  --reference allenai/Molmo2-4B     # required: mm_olmo config.yaml has no model_id
```

**Tensor check** — load both state dicts, compare `export[k]` against `released[k].to(torch.bfloat16)` for all 706 keys. Result: bitwise equal, 706/706.

**The four eval legs.** Note `-o` must follow `-t`/`--harness`:
```bash
DCP=/weka/oe-training-default/mm-olmo/released-models-molmo2-1225/Molmo2-4B/step2000
EXP=<out>/export-molmo2-4b
REF=~/.cache/huggingface/hub/models--allenai--Molmo2-4B/snapshots/042abfa7a38879a376cec03d949eff0aefaa0600

# 1. olmo_core on DCP                                          12m37s
olmo-eval run --harness default -o provider.kind=olmo_core_vlm \
  -m "$DCP" -t chart_qa --no-save-requests

# 2. hf on export                                              47m43s
olmo-eval run --harness default -o provider.kind=hf \
  -o provider.trust_remote_code=true -o provider.multimodal=true \
  -m "$EXP" -t chart_qa --no-save-requests

# 3. vllm on export                                             6m50s
olmo-eval run --harness default -o provider.kind=vllm \
  -o provider.trust_remote_code=true -o provider.max_num_batched_tokens=32768 \
  -m "$EXP" -t chart_qa --no-save-requests

# 4. vllm on released fp32 (dtype control)                      ~7m
olmo-eval run --harness default -o provider.kind=vllm \
  -o provider.trust_remote_code=true -o provider.max_num_batched_tokens=32768 \
  -m "$REF" -t chart_qa --no-save-requests
```

`chart_qa` validation is 1920 instances (960 human + 960 augmented); no `limit`, so 1 instance = 0.052 pts. Reference numbers are mm_olmo's own dumps at `<released-models>/Molmo2-4B/predictions-ck2000-chart_qa-validation/metrics.json` (em 0.74896 = exactly 1438/1920), which also match the values in the `chart_qa.py` docstring.

Agreement was computed from the per-instance `*-predictions.jsonl` (`native_id` → `model_output[0].text` / `.extracted_answer`), 1920 common ids across all four legs.

</details>

<sub>Reviewed and reproduced with Claude Code. Raw metrics/predictions for all four legs are on the box I ran this on — say the word and I'll attach the JSON.</sub>

donovanclay:
Thanks — this is a much better verification than the PR had, and two of the diff findings are
exactly the failure mode this PR is otherwise trying to close. All pushed in `8049c6e`; one
push-back and one hypothesis at the end.

## The bitwise check is now `--verify`

You're right that arguing correctness from `72.50 == 72.50` is weak — 98% text agreement
means outputs differed and netted to zero on a lossy metric. `--verify` does the tensor
comparison, no GPU and no inference, and prints the shape you used. On our fine-tuned v9
against `allenai/Molmo2-4B`:

```
export vs allenai/Molmo2-4B
  tensors: 706 / 706 present, 0 missing, 0 extra, 0 shape mismatches
  export == reference.to(torch.bfloat16), bitwise: 0/706 tensors, 4648972032 differing elements
```

Same structural result as yours; values differ because v9 is fine-tuned, and the tool says
which of those two situations you're in rather than just reporting a number.

## Blockers

**(c) fixed in code**, as you suggested. The root cause is that the two artifacts want
*different loaders*: `is_olmo_core_hf_export` correctly identifies the legacy layout, which
needs `_init_multimodal_from_olmo_core_export`. This PR's output is a genuine HF directory
and needs no special case at all — so `looks_multimodal_hf()` now detects any
image-text-to-text checkpoint from its config (a `vision_config`, or an architecture in
transformers' image-text-to-text mapping). Side benefit: `-o provider.kwargs.multimodal=true`
is no longer needed for `allenai/Molmo2-4B` or Qwen3-VL either. `multimodal` became
tri-state, because with `bool = False` an explicit "off" was indistinguishable from unset.

**(a) and (b)** are in "Notes for reviewers", with the `max_tokens_per_mm_item (31872)` error
quoted so it's greppable. Agreed the vLLM one is the more embarrassing omission given it hits
our own checkpoints.

## Diff findings

All six accepted. Ranking them by blast radius rather than your order:

- **#5 (`format_request` coverage)** — the worst of them. `-o image_mode=none -t blink`
  silently ran the full image benchmark and labelled it a text-only ablation. `mmmu_pro` now
  calls `render_question`; `pointing_base`/`multi_image_base` **raise**, since wiring them up
  properly is a separate feature and pretending is worse than refusing.
- **#1 (`sampling_params`)** — fixed at the layer you identified. `_build_task_overrides`
  routes dotted `sampling_params.*` into `sampling_overrides`, which `preparation.py` applies
  with `replace()`, so `stop_sequences` survive. The `base.py` coercion merges too, as
  defence for other callers.
- **#2** `mmmu:cot` sets `max_tokens=2048`. **#4** the guard reads the effective
  `limit_mm_per_prompt` rather than the constructor default. **#6** `to_dict()` now serializes
  the ablation fields — and `prompt_templates`/`system_prompt_style`, which were already
  missing before this PR. **#7** formatted.

## Push-back: `limit_mm_per_prompt` on text-only models (#3)

I can't reproduce this. Six text-only vLLM runs — Qwen3-1.7B/4B/8B/14B/32B and
Qwen3-4B-Instruct-2507, all `Qwen3ForCausalLM`, no vision tower — ran with the default
`max_images=8`, i.e. `limit_mm_per_prompt={"image": 8}`, and all six completed normally on
vLLM 0.19.1. Which model/version produced the rejection? Happy to add the conditional with a
repro; I'd rather not guard against a failure six runs contradict.

There is also an existing opt-out I should have documented: the injection is behind
`if max_images > 0`, so `-o provider.kwargs.max_images=0` disables it.

## Hypothesis: the one-directional deltas

You're right that mixed-sign noise doesn't describe your table, and I've corrected the
write-up. But I don't think your dtype control rules dtype out — by your own point, vLLM
casts to bf16 at load either way, so bf16-export vs fp32-released can't see the effect.

What differs between the legs isn't storage, it's *compute precision*: the `olmo_core` leg
runs `dtype=float32` with `autocast_dtype=bfloat16` — fp32 master weights, bf16 compute —
while both export legs are pure bf16. That asymmetry predicts a small, consistent drop for
both export legs, which is the shape you measured.

One run settles it: `olmo_core` on the DCP with `-o provider.dtype=bfloat16`. If the gap
closes it's precision, not implementation. Worth doing before we describe it either way.

Also taking the framing point in §3: vLLM agreeing with `olmo_core` more closely than HF does
(97.14% vs 96.30%) while reading byte-identical weights is a better argument for this PR than
the one I was making, and the description now leads with it.

Yes please on the raw JSON — mainly the per-instance predictions for the four legs, so I can
check whether the residual disagreements concentrate anywhere.

donovanclay:
Ran the two experiments I owed you. One confirms your finding's cause; the other caught that
my first fix didn't actually work.

## Your one-directional deltas are precision, not implementation

Full `chart_qa` validation (n=1920) on `Molmo2-4B/step2000`, varying only `dtype` on the
`olmo_core` leg:

| leg | relaxed_correctness |
|---|---:|
| `olmo_core` `dtype=float32` (fp32 master + bf16 autocast) | **0.8375** |
| `olmo_core` `dtype=bfloat16` | **0.8339** |
| your `hf` export (bf16) | 0.8333 |
| your `vllm` export (bf16) | 0.8354 |

The fp32 leg reproduces your 0.8375 exactly. Dropping it to bf16 moves it **into the export
band** — 0.8339 lands between your two export legs. That 0.36-point shift accounts for
essentially all of your 0.42-point spread.

So the export legs aren't systematically worse: the DCP leg was systematically *better*,
because it keeps fp32 master weights and computes in bf16 while both export legs are bf16
throughout. Your fp32-vs-bf16 control couldn't see this, since vLLM casts at load either way
— the difference is compute precision, not storage. Worth knowing generally: any
`olmo_core` number at the default `dtype=float32` is ~0.35 pts above the same weights
evaluated in pure bf16.

## My fix for (c) was wrong, and your repro is what would have caught it

The end-to-end run failed with the same error as before the fix. `looks_multimodal_hf` looked
for a `vision_config` attribute or an architecture in transformers' image-text-to-text
mapping. Molmo2 has **neither**: it names its vision tower `vit_config`, and
`Molmo2ForConditionalGeneration` is remote code so it's absent from the built-in mapping.

My unit tests passed the whole time, because they asserted against a hand-built config
carrying `vision_config` — they encoded my assumption rather than the artifact. Exactly the
class of thing your independent repro exists to catch.

The authoritative signal was in the config all along:

```json
"auto_map": {
  "AutoConfig": "configuration_molmo2.Molmo2Config",
  "AutoModelForImageTextToText": "modeling_molmo2.Molmo2ForConditionalGeneration"
}
```

A remote-code repo declares the head it loads with. Detection now checks `auto_map` first,
then `vit_config`/`vision_tower_config`/`vision_config`, then the built-in mapping. Against
the real configs on disk: v9 export **True**, Qwen3-VL-4B-Instruct **True**, Qwen3-4B
**False**.

End to end, with **no** `multimodal` flag:

```
Detected an image-text-to-text HF config for .../hf-exports/v9-step30000; loading as multimodal
exp1-hf-autodetect   relaxed_correctness=0.7250   n=200
```

0.7250 — identical to the `olmo_core` DCP reference on the same 200 instances.

Both in `6188442`.

jason718:
Re-verified at `f5daedf`. **You're right about `limit_mm_per_prompt` and right about the dtype hypothesis — I was wrong on both.** Details below, plus the JSON you asked for, plus one new blocker: the tip doesn't build against any published OLMo-core.

## I withdraw #3 — and I can tell you why your six runs are correct

Your six runs are right and my finding was wrong. In vLLM 0.19.1, `limit_mm_per_prompt` is consumed only inside a multimodality guard (`vllm/config/model.py:615`):

```python
# Init multimodal config if needed
if self._model_info.supports_multimodal:
    ...
    mm_config_kwargs = dict(..., limit_per_prompt=limit_mm_per_prompt, ...)
```

For `Qwen3ForCausalLM` that branch never runs and the kwarg is **silently discarded** — no validation, no raise. The `ValueError("The model is not multimodal.")` at `model.py:1354` that I keyed on is inside `get_multimodal_config()`, a getter only reached for multimodal models. It has nothing to do with passing the kwarg.

So there's nothing to guard against on the text-only side, and no conditional is needed. Sorry for the noise — I verified the injection was unconditional and then took the consequence on trust instead of checking it.

The memory-profiling half is real but you've already handled it: for a *multimodal* model on a text-only workload the cap does feed worst-case profiling, and `-o provider.kwargs.max_images=0` opts out. Worth putting that opt-out in the docstring next to the error message, since the error names `max_images=N` but not `0`.

## Your dtype experiment settles the one-directional deltas — my control was blind to it

Agreed, and the experiment is convincing. Your `dtype=bfloat16` leg at 0.8339 landing between my two export legs (0.8333 / 0.8354) accounts for essentially the whole 0.42-pt spread.

You're also right about *why* my control couldn't see it: I varied **storage** precision and vLLM normalizes that at load, so both arms of my control were bf16 compute. The axis that mattered was fp32 master weights + bf16 autocast vs pure bf16, which only the `olmo_core` leg can vary. So: not implementation, and the export legs aren't worse — the DCP leg was better. I'd fold your "any `olmo_core` number at default `dtype=float32` sits ~0.35 pts above the same weights in pure bf16" line into the docs somewhere durable; that's a cross-cutting gotcha for anyone comparing an `olmo_core` number against a vLLM one.

## Verified fixed at `f5daedf`

Re-ran against the real artifacts rather than fixtures:

| # | check | result |
|---|---|---|
| (c) | `looks_multimodal_hf` on real configs | export **True**, `allenai/Molmo2-4B` **True**, Qwen3-0.6B / SmolLM2-135M / tiny-Llama **False** |
| #1 | `-o sampling_params.max_tokens=64 -t gsm8k`, mock, end-to-end | `max_tokens=64` **and** `stop_sequences=['Question:', '\n\n']` survive |
| #2 | `mmmu:cot` | `max_tokens=2048` |
| #5 | `render_question` call sites | `mmmu_pro.py:191` + `image_qa_base.py:315`; pointing/multi-image `reject_unsupported_prompt_knobs` |
| #6 | `to_dict()` | all five fields serialized |
| #7 | `ruff format --check` / `ruff check` | clean, 523 files |

Using `auto_map` as the authoritative signal is the right call — it's the model declaring its own head rather than us pattern-matching config shapes.

## New: the tip doesn't build against any published OLMo-core

`f5daedf` imports two symbols that **do not exist in the OLMo-core `vision` branch**:

```
tests/providers/test_olmo_core_vision_key_compat.py  3 failed, 2123 passed, 9 skipped

E  ImportError: cannot import name 'canonicalize_vision_keys'
   from 'olmo_core.nn.vision.molmo2_loader'
   src/olmo_eval/inference/providers/olmo_core_vlm_utils.py:498
```

I checked two commits — `8018f472` (branch tip on Sep 6) and `2dad929a` (tip today) — and `canonicalize_vision_keys` and `legacy_vision_key_mapping` are absent from both. `grep -rn` over the installed package finds neither. So this isn't a stale venv; the OLMo-core side looks unpushed.

The two call sites fail differently, which is worth fixing even after the dependency lands:

- **`olmo_core_unsharded`** (line 498) — bare `from ... import canonicalize_vision_keys`, so **hard `ImportError`**. That path is currently unusable.
- **`olmo_core_dcp`** (line 488) — guarded by `hasattr(model, "legacy_vision_key_mapping")`, so it degrades to `key_mapping=None`. No crash, but the legacy-vision remap the commit advertises is **silently inert** on every published OLMo-core — a legacy `vision.*` DCP checkpoint still fails to load, just without saying why.
- **`mm_olmo_dcp`** — unaffected (falls through to `_load_mm_olmo_state_dict`), which is why my four legs still reproduce.

Suggest making both sites consistent: either both hard-fail with a message naming the required OLMo-core commit, or both degrade with a warning. Right now the same commit does one of each. And since `pyproject.toml` pins `ai2-olmo-core==2.4.0` — which has no `olmo_core.nn.vision` at all — there's nowhere recording which OLMo-core a multimodal run actually needs. A pin or a runtime version check with a real message would save the next person the hour I just spent.

## Still open (small): `base.py` merge doesn't merge

The runner fix is correct and is the path that matters. The `base.py` defence layer, though, merges onto the wrong target:

```python
base_params = type(self).__dict__.get("sampling_params")
```

`type(self)` is `TaskConfig` itself, not the task class, and `TaskConfig.__dict__["sampling_params"]` is **`None`** (the dataclass field default). So `base_params` falls back to bare `SamplingParams()` and task-specific fields are still dropped:

```
gsm8k: cfg.sampling_params.stop_sequences = ('Question:', '\n\n')
dataclasses.replace(cfg, sampling_params={'max_tokens': 64})
  -> max_tokens=64, stop_sequences=None
```

That's the same bug in a new shape, and `dataclasses.replace` on a built config is exactly how variants and `prepare_task_items` go. The original `SamplingParams` is already gone by the time `__post_init__` runs, so a merge can't be done there at all — I'd drop the merge and keep `__post_init__` as pure validation (reject unknown keys, raise on a dict that reached it), leaving the runner as the single place that merges.

## Where the residual disagreements concentrate

Ran the analysis you wanted before shipping the files — short answer, **they don't concentrate anywhere interesting**.

93 of 1920 instances (4.8%) have at least one leg disagreeing on raw text.

| cut | disagreements | baseline |
|---|---|---|
| human subset | 45/960 = 4.7% | — |
| augmented subset | 48/960 = 5.0% | — |
| numeric gold | 78/93 = 84% | 75% of all instances |
| mean output length | 5.2 chars | 5.1 chars overall |

- **Not** subset-dependent — human and augmented are within 0.3 pts of each other.
- Mildly enriched in numeric golds (84% vs 75%), consistent with `relaxed_correctness`'s 5% numeric tolerance flipping near-boundary values.
- **Not** a truncation artifact: mean output length on disagreements is identical to overall (5.2 vs 5.1 chars). Different from the MMMU-Pro pattern in the description, where every mismatch had a truncated side — chart_qa answers are far too short to hit the 12-token cap.
- Only **30 of the 93** text disagreements (32%) changed the score; normalization absorbs the other two-thirds.

So the residual is diffuse per-instance numeric jitter, which is what the dtype story predicts.

## Files

Per-instance predictions and metrics for all four legs, plus the two tensor-comparison scripts:

```
/weka/oe-training-default/jasonr/pr347-verification/
  chart_qa_val_{oc,hf,vllm,vllm_released}.jsonl   # 1920 rows each, native_id keyed
  metrics_{oc,hf,vllm,vllm_released}.json
  cmp_tensors.py        # name/shape/value diff vs a reference repo
  cmp_bf16.py           # the bitwise `export == reference.to(bfloat16)` check
```

Left on weka rather than attached here since this repo is public. Say the word if you'd rather have them as a gist.

<sub>Re-verified with Claude Code on the same A100; `mm_olmo_dcp` legs unchanged from the previous comment, so those numbers stand as posted.</sub>

src/olmo_eval/inference/providers/olmo_core_vlm_utils.py
```diff
@@ -476,13 +477,34 @@ def load_checkpoint_weights(
...
+            if hasattr(model, "legacy_vision_key_mapping")
+            else None
+        )
+        load_model_and_optim_state(
+            str(Path(checkpoint_dir) / "model_and_optim"), model, key_mapping=key_mapping
+        )
         return
 
     if info.format == "olmo_core_unsharded":
+        from olmo_core.nn.vision.molmo2_loader import canonicalize_vision_keys
```
jason718, line 498: **Blocker: this symbol does not exist in any published OLMo-core.**

`canonicalize_vision_keys` is absent from the `vision` branch at both `8018f472` (tip on Sep 6) and `2dad929a` (tip today); `grep -rn` over the installed package finds neither it nor `legacy_vision_key_mapping`. I reinstalled from the current tip specifically to rule out a stale venv. The OLMo-core side looks unpushed.

Effect on a clean checkout:

```
tests/providers/test_olmo_core_vision_key_compat.py   3 failed, 2123 passed, 9 skipped

E  ImportError: cannot import name 'canonicalize_vision_keys'
   from 'olmo_core.nn.vision.molmo2_loader'
   src/olmo_eval/inference/providers/olmo_core_vlm_utils.py:498
```

Because this is a bare module-level-style import inside the branch, the whole `olmo_core_unsharded` path is unusable rather than degraded — contrast the `olmo_core_dcp` path a few lines up, which is `hasattr`-guarded.

Two asks:
1. Push the OLMo-core change (or point at the branch/commit that has it).
2. Record the requirement somewhere. `pyproject.toml` pins `ai2-olmo-core==2.4.0`, which has no `olmo_core.nn.vision` **at all**, so nothing in the repo says which OLMo-core a multimodal run needs. A pin, an extra, or a runtime check that names the required commit would have saved me an hour of bisecting my own environment.

src/olmo_eval/inference/providers/olmo_core_vlm_utils.py
```diff
@@ -476,13 +477,34 @@ def load_checkpoint_weights(
...
+        # OLMo-core moved the ViT and connector under a `vision_backbone.` submodule, so a
+        # freshly built model reports `vision_backbone.vision.*` while every checkpoint
+        # written before that rename has bare `vision.*`. Remap at the checkpoint layer.
+        #
+        # Passing this unconditionally is safe and needs no format sniffing:
+        # `swap_param_keys` skips any entry whose checkpoint-side key is absent from the
+        # checkpoint metadata, so a post-rename checkpoint is untouched.
+        key_mapping = (
+            model.legacy_vision_key_mapping()
+            if hasattr(model, "legacy_vision_key_mapping")
```
jason718, line 489: **This guard makes the remap silently inert on every published OLMo-core.**

`legacy_vision_key_mapping` doesn't exist in OLMo-core `vision` at `8018f472` or `2dad929a` (same missing-dependency problem as line 498). So `hasattr` is always `False` today, `key_mapping` is always `None`, and the legacy-vision remap this commit advertises never runs. A legacy `vision.*` DCP checkpoint still fails to load — just without any indication why.

The inconsistency is the part worth fixing: the same commit hard-fails on the unsharded path and silently no-ops on the DCP path. Pick one. Given the commit message is "Accept both vision key layouts", I'd make both fail loudly and name the OLMo-core commit required — a silent no-op on a compatibility shim is the failure mode this PR is otherwise good at closing.

(For what it's worth, `mm_olmo_dcp` takes neither branch, which is why my four verification legs still reproduce on the released Molmo2-4B.)

src/olmo_eval/evals/tasks/common/base.py
```diff
@@ -185,6 +203,28 @@ def __post_init__(self) -> None:
...
+            if unknown:
+                raise ValueError(
+                    f"unknown sampling_params field(s): {', '.join(sorted(unknown))}; "
+                    f"valid: {', '.join(sorted(valid))}"
+                )
+            # Merge onto whatever the class already declares rather than rebuilding from the
+            # override keys: `SamplingParams(**overrides)` would drop every field the caller
+            # did not mention, e.g. humaneval's stop_sequences. The runner now routes dotted
+            # overrides through sampling_overrides, so this is defence for other callers.
+            base_params = type(self).__dict__.get("sampling_params")
```
jason718, line 223: **This merges onto the wrong target, so it still drops task-specific fields.**

`type(self)` is `TaskConfig`, not the task class, and `TaskConfig.__dict__["sampling_params"]` is `None` — so `base_params` always falls through to a bare `SamplingParams()` and the merge is a no-op against anything the task declared. Verified at `f5daedf`:

```
gsm8k: cfg.sampling_params.stop_sequences = ('Question:', '\n\n')

dataclasses.replace(cfg, sampling_params={'max_tokens': 64})
  -> max_tokens=64, stop_sequences=None
```

Same bug as before, new shape — and `dataclasses.replace` on a built config is exactly how variants and `prepare_task_items` go.

I don't think this is fixable here: by the time `__post_init__` runs, the original `SamplingParams` has already been replaced by the dict, so there's nothing to merge against. Suggest deleting the merge and leaving `__post_init__` as pure validation (reject unknown keys, raise if a dict reaches it at all), with the runner as the single place that merges. **The runner fix itself is correct** — I confirmed end-to-end that `-o sampling_params.max_tokens=64 -t gsm8k` keeps `stop_sequences`, so the actual CLI path is fine.

src/olmo_eval/inference/providers/huggingface.py
```diff
@@ -200,6 +201,68 @@ def prepare_inputs_for_generation(
...
+        return False
+
+    config_kwargs = {
+        key: value
+        for key, value in model_kwargs.items()
+        if key in HuggingFaceProvider._TOKENIZER_KWARGS
+    }
+    try:
+        config = AutoConfig.from_pretrained(model_name, **config_kwargs)
+    except Exception:
```
jason718, line 243: **This swallows the one failure mode most likely to hit Molmo2.**

`AutoConfig.from_pretrained` on a remote-code repo raises unless `trust_remote_code=True` is passed. When a caller hasn't passed it, this `except` returns `False`, the provider takes the text path, and you get back exactly the error this function was written to eliminate:

```
Unrecognized configuration class Molmo2Config ... AutoModelForCausalLM
```

Worse, on the way there transformers writes an **interactive prompt** to stdout:

```
Do you wish to run the custom code? [y/N]
```

In a Beaker worker with no stdin that's at best noise in the log and at worst a hang.

Measured at `f5daedf` against real configs on disk:

| model | `trust_remote_code` | result |
|---|---|---|
| PR-347 export (Molmo2) | `True` | **True** |
| PR-347 export (Molmo2) | *(unset)* | **False** + interactive prompt |
| `allenai/Molmo2-4B` | `True` | **True** |
| Qwen3-0.6B / SmolLM2-135M / tiny-Llama | either | **False** |

So the detection logic is right — `auto_map` is the correct authoritative signal and I'd keep it. It's only the no-flag path that regresses to the old confusing error. Suggest catching the remote-code case specifically and raising something that names `trust_remote_code`, rather than folding it into a blanket `return False`.

jason718 (changes requested):
Re-reviewed at `f5daedf`. Most of the round-1 findings are genuinely fixed and I've verified them against real artifacts rather than fixtures — `looks_multimodal_hf` on real configs, the sampling_params runner fix end-to-end on gsm8k, `mmmu:cot` at 2048, `to_dict()` serializing all five fields, lint clean. The `auto_map` approach is the right call.

Requesting changes for one reason: **the branch tip doesn't build against any published OLMo-core.** `canonicalize_vision_keys` and `legacy_vision_key_mapping` exist in neither `8018f472` nor today's tip `2dad929a`, so `pytest` is 3 red on a clean checkout and the `olmo_core_unsharded` path raises `ImportError`. That's a hard blocker until the OLMo-core side is pushed and the dependency is recorded somewhere — `pyproject.toml` currently pins `ai2-olmo-core==2.4.0`, which has no `olmo_core.nn.vision` at all.

Three smaller inline comments alongside it: the `hasattr` guard that makes the DCP remap silently inert, the `base.py` merge that doesn't merge, and the `except Exception` that swallows the missing-`trust_remote_code` case back into the old confusing error.

Also, to restate from the thread so it isn't lost: **I withdraw the `limit_mm_per_prompt` finding entirely** — your six runs are right, vLLM 0.19.1 discards the kwarg for non-multimodal models (`vllm/config/model.py:615`), and no guard is needed. And your dtype experiment settles the one-directional deltas; my storage-dtype control couldn't see compute precision. Both corrections are in the thread with detail.

Nothing here touches the parity result — the four verification legs run through `mm_olmo_dcp`, which takes neither of the broken paths, so those numbers stand.

donovanclay:
Thanks — and thank you for the two withdrawals. The `vllm/config/model.py:615` pointer is the mechanism I couldn't find, and you're right that the storage-dtype control couldn't see compute precision; I should have said explicitly that the `olmo_core` leg runs fp32 master weights under bf16 autocast rather than leaving you to infer it.

## The blocker: right diagnosis, wrong ref

The symbols aren't unpushed — they're on **[allenai/OLMo-core#834](https://github.com/allenai/OLMo-core/pull/834)**, head `donovan/training-speed-and-image-v10` @ `208d0be`, base `vision`. You checked `vision` at `8018f472` and `2dad929a`, which is why they weren't there: #834 is still open, so it hasn't landed on the base branch. Everything is fetchable now:

| symbol | at `208d0be` |
|---|---|
| `canonicalize_vision_keys` | `nn/vision/molmo2_loader.py:219`, exported from `nn/vision/__init__.py` |
| `legacy_vision_key_mapping` | `nn/vision/multimodal.py:393` |
| `strip_vision_backbone_prefix`, `VISION_BACKBONE_PREFIX` | present, exported |

```
pip install 'ai2-olmo-core @ git+https://github.com/allenai/OLMo-core.git@donovan/training-speed-and-image-v10'
```

That's entirely my fault for not naming the branch in the PR description — "an hour of bisecting my own environment" is exactly the cost of that omission.

**But the substance of the finding stands, so I've fixed it rather than just pointing at the branch.** A reviewer shouldn't have to know which of my open PRs to correlate, and `pyproject.toml` pinning `ai2-olmo-core==2.4.0` — which has no `olmo_core.nn.vision` at all — really does leave nothing in the repo stating the requirement.

## Three fixes pushed (`aa423a9`)

**1. Both vision-key paths now fail loudly and name #834** (`olmo_core_vlm_utils.py`)

You were right that the inconsistency was the worst part. The `hasattr` guard is gone; `_legacy_vision_key_mapping()` raises when the helper is absent, and the unsharded import is wrapped so an `ImportError` becomes the same message instead of naming only a symbol. Both now say which OLMo-core to install and how.

I took your "pick one" as "fail loudly" rather than "no-op quietly," on the reasoning that the helper and the rename it compensates for ship in the *same* OLMo-core change — so its absence means the install is simply too old, and the alternative is a legacy checkpoint loading with an empty vision tower and scoring at chance with nothing in the log.

**2. `base.py.__post_init__` is now validation-only**

Your diagnosis was exact: `type(self)` is `TaskConfig`, whose `sampling_params` default is `None`, so `base_params` always fell through to a bare `SamplingParams()` and the merge was a no-op against anything the task declared. I took your suggestion and deleted the merge entirely — unknown keys still raise, and a dict reaching `__post_init__` now raises `TypeError` naming `sampling_overrides` as the supported route. The runner stays the single merge point.

**3. The `trust_remote_code` case is raised, not swallowed** (`huggingface.py`)

Also fixed the interactive-prompt hazard you flagged, which I hadn't considered: `config_kwargs.setdefault("trust_remote_code", False)` turns the `[y/N]` stdin prompt into a deterministic raise, so a Beaker worker can't hang on it. `_is_remote_code_error()` then matches the refusal and re-raises naming the flag. Everything else falls through to `return False` as before, so an unreadable or unrecognised repo still takes the text path.

## Your "3 red on a clean checkout" is fixed too

The OLMo-core-dependent cases are now `skipif`-guarded in two tiers — `_needs_olmo_core` for the DCP cases, which only need the module importable, and `_needs_olmo_core_vision_keys` for the unsharded ones — each with #834 named in the skip reason. So a clean checkout reports skips that say what's missing, not failures that look like broken code.

```
with #834 installed:  8 passed
with no OLMo-core:    2 passed, 6 skipped
full suite:           2125 passed, 11 skipped
ruff check / format:  clean    ty: clean
```

## Still open: recording the dependency

I haven't touched the `pyproject.toml` pin, because I don't think olmo-eval should pin a git branch on `vision`. The ordering that actually resolves it is OLMo-core#834 merging first, then this PR pinning a released `ai2-olmo-core` that contains the vision module. Until then the runtime errors above are the stopgap — they name the branch at the point of failure, which is where it's useful.

Happy to add an optional `[multimodal]` extra pinning the git ref in the meantime if you'd rather have it machine-readable than only in an error message. Your call — it's the last thing between this and green from my side.

PR #349, opened 2026-09-03
Let sandboxes receive files too large for a command

gregorybchris:
## Problem

A scorer reaches a sandbox only through a command string. swe-rex runs that through `execve`, which caps a **single argument at 128KB** (`MAX_ARG_STRLEN`). Fixtures larger than that cannot reach the container at all — the request succeeds and the failure happens inside, at the exec boundary.

That is fine for the code evals we have today, where the test payload is a few hundred bytes of assertions. It rules out any task whose fixtures are megabytes rather than kilobytes.

## Approach

swe-rex already serves a `POST /write_file` endpoint whose payload travels in the request body rather than in argv. It was simply never surfaced: `SandboxExecutor` keeps its runtime private, and `CapabilityExecutionEnvironment` published only `execute` / `execute_command` / `execute_code`.

- `SandboxExecutor.write_files()` exposes it, going through the same transport retries as every other runtime call.
- `SandboxManager.execute_with_files()` stages files and runs a command against them **while holding a single lease**.

The lease is the crux. `execute_command` leases an executor from the pool per call, so a separate write and run can land on different containers, and the command would execute somewhere its files were never written. Both steps now happen inside one `_execute_with_lease`.

`StagingExecutionEnvironment` is declared separately from `ExecutionEnvironment` rather than added to it, so environments that cannot stage files still satisfy the base protocol, and a scorer needing the capability fails a type check instead of an attribute lookup at runtime.

## Tests

`tests/core/harness/test_sandbox_file_staging.py` asserts the properties that matter:

- files are present before the command runs
- with several executors and concurrent calls, each command runs on the executor that received *its* files
- staging respects per-executor concurrency limits

## Notes

No existing task changes behavior; this only adds a capability. First caller is the LiveCodeBench task, whose test payloads are 10–20MB at p90 — see #3 in this stack.

### DIFF
--- a/src/olmo_eval/common/execution/__init__.py
+++ b/src/olmo_eval/common/execution/__init__.py
@@ -1,6 +1,11 @@
 """Execution helpers for scoring and sandboxed code execution."""
 
-from .environment import ExecutionEnvironment, ExecutionResult, ScoringContext
+from .environment import (
+    ExecutionEnvironment,
+    ExecutionResult,
+    ScoringContext,
+    StagingExecutionEnvironment,
+)
 from .process_pool import (
     ProcessOutputScore,
     ProcessPoolManager,
@@ -20,6 +25,7 @@ __all__ = [
     "ProcessScoringPoolConfig",
     "ScoringContext",
     "SerializedProcessScorer",
+    "StagingExecutionEnvironment",
     "default_process_pool_workers",
     "serialize_process_scorer",
 ]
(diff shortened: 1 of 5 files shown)


gregorybchris:
Superseded by #352 — the branch was renamed to `chrisg/sandbox-file-staging` and renaming a PR's head branch closes it rather than retargeting it. Same commit, same description.

PR #350, opened 2026-09-03
Load the data files a caller asked for from script-based repos

gregorybchris:
## The bug

`datasets` 4.x refuses to run a repository's legacy loading script, so `HuggingFaceBackend` falls back to reading the repository's data files directly. That fallback chose files by testing whether the **subset name appears in the filename**:

```python
subset = source.subset or ""
candidates = [f for f in repo_files if subset in f and ...]
```

With no subset configured, `subset` is `""` — and `"" in filename` is true for every file in the repository. Any explicitly declared `data_files` was ignored, and a request for one file quietly returned all of them.

Measured on `livecodebench/code_generation_lite`, whose six files are separate benchmark releases: asking for the 400 problems in `test.jsonl` returned **all 1055** across all six, with nothing to indicate the request had been widened. A caller gets a plausible-looking dataset that is not the one they asked for.

## Fix

The fallback now loads exactly the files a caller declared, and falls back to subset matching only when none were declared.

`DataSource.data_files` also accepts a `tuple[str, ...]` as well as a `str`, since one split can span several files — something the underlying `load_dataset` has always supported (`str | Sequence[str] | Mapping[...]`) and that `DataSource` had narrowed since the initial commit.

## Why the union type, and not tuple everywhere

`tuple` alone would be a quiet breaking change. `DataSource.to_dict()` feeds `TaskConfig.to_dict()`, which feeds `compute_task_hash()`, and results are stored under that hash. Serializing `"olympiad/test.jsonl"` as `["olympiad/test.jsonl"]` moves the hash:

```
frontierscience_olympiad, data_files as str   : 5901a8f78e0afbee
frontierscience_olympiad, data_files as tuple : 04f9e11d001651d9
```

**11 registered tasks** declare `data_files` today (`frontierscience_*`, `astabench_*`, eight `basic_skills_*`). Moving their hashes would detach every stored result from future runs, and the analysis tooling joins on `task_hash`.

So existing single-file declarations stay strings and serialize as strings. A separate consideration: `data_source` is reachable from CLI overrides, where values arrive as strings, and an always-tuple field would silently `list()` a string into seven single-character filenames.

## Tests

- `tests/data/test_hub_file_selection.py` — declared files are the only ones loaded; a single declared file is not widened; subset matching still works when nothing is declared; a missing subset still raises.
- `tests/data/test_sources.py` — tuple `data_files` survives `with_split`/`with_subset`, serializes as a list, leaves single-file declarations untouched, and keeps `DataSource` hashable.

Second in a three-PR stack; contains #349.

### DIFF
--- a/src/olmo_eval/common/execution/__init__.py
+++ b/src/olmo_eval/common/execution/__init__.py
@@ -1,6 +1,11 @@
 """Execution helpers for scoring and sandboxed code execution."""
 
-from .environment import ExecutionEnvironment, ExecutionResult, ScoringContext
+from .environment import (
+    ExecutionEnvironment,
+    ExecutionResult,
+    ScoringContext,
+    StagingExecutionEnvironment,
+)
 from .process_pool import (
     ProcessOutputScore,
     ProcessPoolManager,
@@ -20,6 +25,7 @@ __all__ = [
     "ProcessScoringPoolConfig",
     "ScoringContext",
     "SerializedProcessScorer",
+    "StagingExecutionEnvironment",
     "default_process_pool_workers",
     "serialize_process_scorer",
 ]
--- a/src/olmo_eval/common/execution/environment.py
+++ b/src/olmo_eval/common/execution/environment.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass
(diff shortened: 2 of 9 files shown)


gregorybchris:
Superseded by #353 — branch renamed to `chrisg/hf-data-files-selection`. Same commits, same description.

PR #351, opened 2026-09-03
Add LiveCodeBench code generation task

gregorybchris:
Ports LiveCodeBench from [oe-eval](https://github.com/allenai/oe-eval). Contest problems are tagged by contest date, so a model can be scored on problems published after its training data was collected, and solutions are graded by running them against the contest's own test cases.

Third in a stack — contains #349 (sandbox file staging) and #350 (data file selection), both of which this depends on. **Review those first;** the LiveCodeBench-specific change is the last commit.

## Tasks and variants

| spec | mirrors oe-eval |
|---|---|
| `livecodebench` | `::olmo3:adapt` — think-tag prompt, temp 0.6, top_p 0.95, context-bounded generation, 10 samples, pass@1/5/10 |
| `livecodebench:tulu` | `::tulu` — no reasoning instruction, temp 0.2, 2048 tokens |
| `livecodebench:lite` | `..._lite` — single sample, pass@1, for cheap iteration |
| `livecodebench:grpo` | `:grpo::tulu-thinker` — temp 1.0, top_p 1.0, 16384 tokens, primary pass@10 |

`livecodebench` is release_v3 (612 problems); `livecodebench_hidden` is the v4–v6 window after the v3 cutoff (443 problems), registered separately so problems postdating a model's training data can be scored on their own. Both carry all four variants.

Left out: `::tulu-thinker` (temp 0.8, superseded by the deepseek regime) and `::olmo3:midtrain` (the base regime plus stop sequences). Easy to add if wanted. Nothing is taken from oe-eval-internal, which is deprecated and whose `no_think_tags` configs drop the system prompt in a way oe-eval's do not.

## What makes this task different

Its fixtures. HumanEval ships a few hundred bytes of assertions per problem; LiveCodeBench ships the contest's real input/output pairs. Measured across release_v3:

- median problem: tens of KB
- p90: **10–20MB**
- largest: **92MB** compressed, 199MB decoded
- release total: 2.6GB compressed

Neither oe-eval nor upstream caps them, and capping is not viable: even a 4MB budget truncates 26% of release_v3 problems, keeping a median of 43% of their tests, which would break comparability with published numbers.

Two consequences:

**They reach the container as staged files** rather than in the command string — the capability added in #349.

**They are kept off the instances.** `build_requests` splats `instance.metadata` into the run's request records, so carrying payloads there would write gigabytes per run. Instances carry a dataset row reference and the scorer reads the payload when it grades.

That row reference is an implicit coupling, so the scorer refuses to grade a row whose `question_id` does not match the instance. Grading against another problem's tests would score every solution wrong while still producing a plausible-looking number. Alignment was checked across both releases: 612 and 443 instances, all unique ids, zero mismatches.

## Grading

A port of the reference harness, staged into the container as a script, following the existing `harness/sandbox/scripts/` + `get_script()` pattern.

- **Call-based problems** (LeetCode) invoke a named function on a `Solution` instance and compare parsed values; a tuple result is not counted as a mismatch, since ground truth is never a tuple.
- **Whole-program problems** get stdin fed and stdout compared line by line, with a decimal fallback so numerically equal output is not failed over formatting. `if __name__ == '__main__'` guards are unwrapped so the program can be called per test case.
- Compile failures, missing entry points, wrong answers and timeouts are distinguished in the recorded error code. A solution that does not compile is a failed solution, not a failure of the grader.

## Verification

- **Prompts byte-identical to oe-eval** across 61 real problems (52 starter-code, 9 stdin) in both the thinking and non-thinking regimes, checked against an independently constructed copy of oe-eval's template.
- **Grader correct on real dataset problems** in both modes: correct solutions pass, wrong answers and missing functions fail with the right codes, `__main__` guards unwrap, infinite loops time out.
- **End to end in Docker** (mock model, `release_v3`): the grader ran in the container, decoded the compressed blobs there — reporting 14, 14 and 15 cases for the three problems, the real per-problem counts — and returned verdicts the scorer turned into pass@1.
- **Integration test** (`tests/integration/test_livecodebench_sandbox.py`, skipped without Docker) covers what stubs cannot: correct solutions to a whole-program and a call-based problem both score 1.0 through the real staging path, a wrong one scores 0.0.
- 31 unit tests for prompts, release selection, variants, the scorer, and the grader. Full suite: 2193 passed, 8 skipped.

## Still open

No real model has generated solutions yet — a short vLLM run would confirm the last mile and give a first pass@1 to compare against oe-eval's.

Unrelated, but noticed while testing: the `codex_python` harness preset cannot start locally. Its image has no `swerex-remote` and the preset sets neither `inject_swerex` nor a startup timeout above 60s, so all four instances fail. `codex_universal` sets both. I worked around it with an inline sandbox override rather than change the preset here.

### DIFF
--- a/src/olmo_eval/common/execution/__init__.py
+++ b/src/olmo_eval/common/execution/__init__.py
@@ -1,6 +1,11 @@
 """Execution helpers for scoring and sandboxed code execution."""
 
-from .environment import ExecutionEnvironment, ExecutionResult, ScoringContext
+from .environment import (
+    ExecutionEnvironment,
+    ExecutionResult,
+    ScoringContext,
+    StagingExecutionEnvironment,
+)
 from .process_pool import (
     ProcessOutputScore,
     ProcessPoolManager,
@@ -20,6 +25,7 @@ __all__ = [
     "ProcessScoringPoolConfig",
     "ScoringContext",
     "SerializedProcessScorer",
+    "StagingExecutionEnvironment",
     "default_process_pool_workers",
     "serialize_process_scorer",
 ]
--- a/src/olmo_eval/common/execution/environment.py
+++ b/src/olmo_eval/common/execution/environment.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
@@ -37,6 +38,23 @@ class ExecutionEnvironment(Protocol):
 
     async def execute_code(
         self, code: str, language: str = "python", timeout: float | None = None
+    ) -> ExecutionResult: ...
+
+
+@runtime_checkable
+class StagingExecutionEnvironment(ExecutionEnvironment, Protocol):
+    """Execution environment that can place files before running a command.
+
+    Command text reaches the container as a process argument, which the
+    operating system caps at a small size. Payloads too large for that limit
+    are staged as files instead, in the same environment the command runs in.
+    """
+
+    async def execute_with_files(
+        self,
+        command: str,
+        files: Mapping[str, str],
+        timeout: float | None = None,
     ) -> ExecutionResult: ...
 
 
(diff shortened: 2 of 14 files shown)


gregorybchris:
Superseded by #354 — branch renamed to `chrisg/livecodebench-task`. Same commits, same description.

PR #352, opened 2026-09-03
Let sandboxes receive files too large for a command

gregorybchris:
## Problem

A scorer reaches a sandbox only through a command string. swe-rex runs that through `execve`, which caps a **single argument at 128KB** (`MAX_ARG_STRLEN`). Fixtures larger than that cannot reach the container at all — the request succeeds and the failure happens inside, at the exec boundary.

That is fine for the code evals we have today, where the test payload is a few hundred bytes of assertions. It rules out any task whose fixtures are megabytes rather than kilobytes.

## Approach

swe-rex already serves a `POST /write_file` endpoint whose payload travels in the request body rather than in argv. It was simply never surfaced: `SandboxExecutor` keeps its runtime private, and `CapabilityExecutionEnvironment` published only `execute` / `execute_command` / `execute_code`.

- `SandboxExecutor.write_files()` exposes it, going through the same transport retries as every other runtime call.
- `SandboxManager.execute_with_files()` stages files and runs a command against them **while holding a single lease**.

The lease is the crux. `execute_command` leases an executor from the pool per call, so a separate write and run can land on different containers, and the command would execute somewhere its files were never written. Both steps now happen inside one `_execute_with_lease`.

`StagingExecutionEnvironment` is declared separately from `ExecutionEnvironment` rather than added to it, so environments that cannot stage files still satisfy the base protocol, and a scorer needing the capability fails a type check instead of an attribute lookup at runtime.

## Tests

`tests/core/harness/test_sandbox_file_staging.py` asserts the properties that matter:

- files are present before the command runs
- with several executors and concurrent calls, each command runs on the executor that received *its* files
- staging respects per-executor concurrency limits

## Notes

No existing task changes behavior; this only adds a capability. First caller is the LiveCodeBench task, whose test payloads are 10–20MB at p90 — see #354.

### DIFF
--- a/src/olmo_eval/common/execution/__init__.py
+++ b/src/olmo_eval/common/execution/__init__.py
@@ -1,6 +1,11 @@
 """Execution helpers for scoring and sandboxed code execution."""
 
-from .environment import ExecutionEnvironment, ExecutionResult, ScoringContext
+from .environment import (
+    ExecutionEnvironment,
+    ExecutionResult,
+    ScoringContext,
+    StagingExecutionEnvironment,
+)
 from .process_pool import (
     ProcessOutputScore,
     ProcessPoolManager,
@@ -20,6 +25,7 @@ __all__ = [
     "ProcessScoringPoolConfig",
     "ScoringContext",
     "SerializedProcessScorer",
+    "StagingExecutionEnvironment",
     "default_process_pool_workers",
     "serialize_process_scorer",
 ]
--- a/src/olmo_eval/common/execution/environment.py
+++ b/src/olmo_eval/common/execution/environment.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Protocol, runtime_checkable
 
@@ -37,6 +38,23 @@ class ExecutionEnvironment(Protocol):
 
     async def execute_code(
         self, code: str, language: str = "python", timeout: float | None = None
+    ) -> ExecutionResult: ...
+
+
+@runtime_checkable
+class StagingExecutionEnvironment(ExecutionEnvironment, Protocol):
+    """Execution environment that can place files before running a command.
+
+    Command text reaches the container as a process argument, which the
+    operating system caps at a small size. Payloads too large for that limit
+    are staged as files instead, in the same environment the command runs in.
+    """
+
+    async def execute_with_files(
+        self,
+        command: str,
+        files: Mapping[str, str],
+        timeout: float | None = None,
     ) -> ExecutionResult: ...
 
 
--- a/src/olmo_eval/harness/sandbox/executor.py
+++ b/src/olmo_eval/harness/sandbox/executor.py
@@ -11,7 +11,7 @@ import os
 import random
 import time
 import uuid
-from collections.abc import Awaitable, Callable
+from collections.abc import Awaitable, Callable, Mapping
 from dataclasses import dataclass
 from typing import Any, TypeVar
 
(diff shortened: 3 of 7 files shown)


gregorybchris:
Thanks — two of the three were right and are fixed in d25b955.

**Retry label doesn't name the path** — correct, and it mattered: a failed write reported `file write` whichever file it was, which is useless precisely when the payload is large. The label is now `file write to {path}`, so it carries into the retry logs and into the `SandboxTransportError` raised once retries are exhausted.

**Capacity test didn't test capacity** — correct, and the sharper point is that the assertion (`len(commands) == 4`) holds for *any* capacity, so the test's name was a promise it never checked. The stub now counts operations in flight, following `test_sandbox_manager_capacity.py`, and asserts `peak == 1` at `max_concurrency=1`. I added a second test at `max_concurrency=3` asserting `1 < peak <= 3`, so neither test can pass on a counter that is stuck at one.

Worth noting what the assertion actually protects: the lease has to bound *writing* as well as executing, because that is what stops a concurrent caller replacing a file between another caller's write and its command.

**`capabilities or Capability.DEFAULT` conflating `frozenset()` with `None`** — the reading is right, but I've left it. `execute`, `execute_command` and `execute_code` all contain that line verbatim (manager.py:486, 511, 541), so changing it only in `execute_with_files` would leave the manager inconsistent with itself and give four sibling methods two different semantics. It belongs in a change that treats all four together, which is a wider behavioral change than this PR should carry.

abhishekraok (approved):
Looks good. 

Only one issue remaining according to Claude:

`write_files` is the only runtime call with no timeout.**
`execute_command` wraps its runtime call in `asyncio.wait_for(..., timeout + 1.0)`;
`write_files` doesn't. swerex's `_request` posts with no `ClientTimeout`, so
aiohttp's 5-minute default applies — and the resulting `TimeoutError` *is*
classified by `_is_transport_error`, so it retries: 4 attempts, ~20 minutes,
then a quarantine health check. Separately, `execute_with_files(timeout=...)`
bounds only the command, so a caller passing `timeout=900` can block well past
that in staging.

Naming the path makes the eventual error good; it doesn't bound how long you
wait for it. With #354's payloads as the first real user, this is the stall
you'd actually hit. Suggest an explicit `wait_for` around the write, and a
docstring note that `timeout` covers the command only.

Nit: `_StagingExecutor` now duplicates the `active`/`peak` bookkeeping that
`_TrackingExecutor` in `test_sandbox_manager_capacity.py` already has, on top of
the duplicated `_manager()`. A shared fixture would carry both.

Otherwise good to merge.

gregorybchris:
Both addressed in 4404c9a. The timeout point was right, and worse in practice than the description: I confirmed swerex's `_request` sets no `ClientTimeout` at all, and `_is_transport_error` does classify `TimeoutError` as retryable, so the sequence really was aiohttp's 5-minute default × 4 attempts before `_quarantine_if_unresponsive` even ran.

**Bounding the write.** Each write is now wrapped in `asyncio.wait_for`, mirroring `execute_command` — per attempt, inside the retry loop — so the worst case is the retry budget × timeout rather than open-ended. The default is `config.command_timeout`, the same bound commands get.

**The docstring.** I went one step further than "timeout covers the command only": `execute_with_files` now passes its `timeout` through to `write_files`, so a caller's bound covers *every* step — each file write and the command, applied separately. The docstring says exactly that. I preferred this to a hidden second default for writes because a caller who passes `timeout=900` for a heavy job gets a predictable bound on everything with one knob, and a large payload that legitimately needs longer than 30s to upload can ask for it. A single value applied to each step is a different promise from one applied to the whole, so that distinction is spelled out.

**Tests that would have caught this.** Two tests run `write_files` against a runtime whose writes never finish. With an explicit timeout and with the config default, each gives up after the retry budget (4 attempts), within a couple of seconds, with the path in the error. Worth noting how I confirmed they test the real thing: I stashed the fix and ran them — they don't fail, they *hang*, which is precisely the behavior being fixed. (Retry backoff is patched to zero in that file so the retries themselves don't add wall-clock time.)

**The nit.** Agreed. Both files now import a single `TrackingExecutor` and `make_manager` from `tests/core/harness/sandbox_stubs.py`. The stub records every kind of operation (`execute_code`, `execute_command`, `write_files`) through one `_run()` that does the active/peak bookkeeping, so either test file can assert on capacity, routing or staging without a second copy. I used a plain module rather than a pytest fixture because the tests construct executors with different arguments (`max_concurrency`, failure modes) per test — a class import carries that more naturally than a fixture would. The capacity tests are unchanged in behavior; only the helper moved.

#354 is rebuilt on this commit and re-verified, including the integration tests against real containers.

PR #353, opened 2026-09-03
Load the data files a caller asked for from script-based repos

gregorybchris:
## The bug

`datasets` 4.x refuses to run a repository's legacy loading script, so `HuggingFaceBackend` falls back to reading the repository's data files directly. That fallback chose files by testing whether the **subset name appears in the filename**:

```python
subset = source.subset or ""
candidates = [f for f in repo_files if subset in f and ...]
```

With no subset configured, `subset` is `""` — and `"" in filename` is true for every file in the repository. Any explicitly declared `data_files` was ignored, and a request for one file quietly returned all of them.

Measured on `livecodebench/code_generation_lite`, whose six files are separate benchmark releases: asking for the 400 problems in `test.jsonl` returned **all 1055** across all six, with nothing to indicate the request had been widened. A caller gets a plausible-looking dataset that is not the one they asked for.

## Fix

The fallback now loads exactly the files a caller declared, and falls back to subset matching only when none were declared.

`DataSource.data_files` also accepts a `tuple[str, ...]` as well as a `str`, since one split can span several files — something the underlying `load_dataset` has always supported (`str | Sequence[str] | Mapping[...]`) and that `DataSource` had narrowed since the initial commit.

## Why the union type, and not tuple everywhere

`tuple` alone would be a quiet breaking change. `DataSource.to_dict()` feeds `TaskConfig.to_dict()`, which feeds `compute_task_hash()`, and results are stored under that hash. Serializing `"olympiad/test.jsonl"` as `["olympiad/test.jsonl"]` moves the hash:

```
frontierscience_olympiad, data_files as str   : 5901a8f78e0afbee
frontierscience_olympiad, data_files as tuple : 04f9e11d001651d9
```

**11 registered tasks** declare `data_files` today (`frontierscience_*`, `astabench_*`, eight `basic_skills_*`). Moving their hashes would detach every stored result from future runs, and the analysis tooling joins on `task_hash`.

So existing single-file declarations stay strings and serialize as strings. A separate consideration: `data_source` is reachable from CLI overrides, where values arrive as strings, and an always-tuple field would silently `list()` a string into seven single-character filenames.

## Tests

- `tests/data/test_hub_file_selection.py` — declared files are the only ones loaded; a single declared file is not widened; subset matching still works when nothing is declared; a missing subset still raises.
- `tests/data/test_sources.py` — tuple `data_files` survives `with_split`/`with_subset`, serializes as a list, leaves single-file declarations untouched, and keeps `DataSource` hashable.

Independent of #352 — the two share no files and can be reviewed in either order. #354 (LiveCodeBench) depends on both and contains them.

### DIFF
--- a/src/olmo_eval/data/backends/huggingface.py
+++ b/src/olmo_eval/data/backends/huggingface.py
@@ -48,7 +48,11 @@ class HuggingFaceBackend:
 
         kwargs: dict[str, Any] = {}
         if source.data_files is not None:
-            kwargs["data_files"] = source.data_files
+            kwargs["data_files"] = (
+                list(source.data_files)
+                if isinstance(source.data_files, tuple)
+                else source.data_files
+            )
         if source.revision is not None:
             kwargs["revision"] = source.revision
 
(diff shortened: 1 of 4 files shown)


gregorybchris:
Thanks — all three were right and are fixed in 3bd8b5a.

**`revision` ignored in the fallback** (both the listing and the file fetch) — correct, and this was the most important of the three. The primary path honors `revision`; the fallback silently didn't, so a pinned dataset that fell back read whatever the default branch holds. That is the same class of bug this PR exists to fix: a caller's selection quietly replaced by something else. It is not hypothetical either — `frontierscience` pins a revision alongside `data_files` today.

I had actually written the `@{revision}` URL handling earlier and removed it as speculative while trimming scope. That was the wrong call: since the primary path honors it, the fallback dropping it is an inconsistency, not an unused feature. Both call sites now use it, and there's a test asserting the revision reaches the listing *and* the URLs — the listing matters because selection computed against one revision then read from another is silently wrong.

**Empty `data_files` → `IndexError`** — correct. An empty selection is a caller mistake rather than a missing file, so it now raises a `ValueError` saying so instead of failing on `candidates[0]`.

**No extension validation for declared files** — correct, and slightly worse than described. The loader module was taken from the *first* file's extension with no check, so a mixed set like `("a.jsonl", "b.parquet")` was read entirely as whatever the first file happened to be. Both cases now fail with the offending filenames named: unreadable types (your `README.md` typo example) and mixed types. The supported types live in one mapping the subset filter also uses, so they aren't spelled out twice.

Six new tests cover the revision on both paths, the unpinned case, and the three rejections.

One note on the review's file list: it covered 9 files including the sandbox staging ones, because this branch was stacked on #352 when the review ran. They're now both based on `main` independently, so this PR is 4 files.

abhishekraok (changes requested):
Thanks for the fixes. One issue remains.
 
Claude: Test coverage doesn't reach load(). Every test calls the private _load_from_hub_files and passes data_files=/revision= explicitly while also setting them on the source, so nothing asserts that load() forwards them into kwargs — the integration where the bug actually lived. No test drives load() through the fallback at all.

abhishekraok (approved):

PR #354, opened 2026-09-03
Add LiveCodeBench code generation task

gregorybchris:
Ports LiveCodeBench from [oe-eval](https://github.com/allenai/oe-eval). Contest problems are tagged by contest date, so a model can be scored on problems published after its training data was collected, and solutions are graded by running them against the contest's own test cases.

Builds on #352 (sandbox file staging) and #353 (data file selection), both now merged; this branch is rebased on main and contains only the LiveCodeBench commits.

## Tasks and variants

| spec | prompt and sampling |
|---|---|
| `livecodebench` | LiveCodeBench's system prompt plus a reasoning line that names no tags ("Provide CONCISE reasoning on how to arrive at the answer."); sampling as oe-eval's `::olmo3:adapt`: temp 0.6, top_p 0.95, context-bounded generation, 10 samples, pass@1/5/10 |
| `livecodebench:think_tags` | oe-eval's `::olmo3:adapt` prompt, which asks for reasoning inside literal `<think> </think>` tags |
| `livecodebench:paper` | oe-eval-internal's `::tulu-thinker_deepseek_no_think_tags`, the regime reported in the OLMo 3 paper: the default prompt with no system message |
| `livecodebench:tulu` | oe-eval's `::tulu` — no reasoning instruction, temp 0.2, 2048 tokens |
| `livecodebench:lite` | oe-eval's `..._lite` — the default regime on a single sample, pass@1, for cheap iteration |
| `livecodebench:grpo` | oe-eval's `:grpo::tulu-thinker` — temp 1.0, top_p 1.0, 16384 tokens, primary pass@10 |

**Why the default prompt names no tags.** The `<think>`-tag instruction dates from models trained on literal tags whose chat template did not insert them. A template that already opens a thinking block gets asked for something it has done, and a model with no such training is asked to invent a format, so the instruction measured template compliance alongside coding ability. The tag-free line gives both kinds of model the same instruction. The system prompt is the benchmark's own and is kept; the paper regime dropped it, which is available as `:paper` for reproducing published numbers.

`livecodebench` is release_v3 (612 problems); `livecodebench_hidden` is the v4–v6 window after the v3 cutoff (443 problems), registered separately so problems postdating a model's training data can be scored on their own. Both carry all six variants.

Left out: `::tulu-thinker` (temp 0.8, superseded by the deepseek regime) and `::olmo3:midtrain` (the base regime plus stop sequences). Easy to add if wanted.

## What makes this task different

Its fixtures. HumanEval ships a few hundred bytes of assertions per problem; LiveCodeBench ships the contest's real input/output pairs. Measured across release_v3:

- median problem: tens of KB
- p90: **10–20MB**
- largest: **92MB** compressed, 199MB decoded
- release total: 2.6GB compressed

Neither oe-eval nor upstream caps them, and capping is not viable: even a 4MB budget truncates 26% of release_v3 problems, keeping a median of 43% of their tests, which would break comparability with published numbers.

Two consequences:

**They reach the container as staged files** rather than in the command string — the capability added in #352.

**They are kept off the instances.** `build_requests` splats `instance.metadata` into the run's request records, so carrying payloads there would write gigabytes per run. Instances carry a dataset row reference and the scorer reads the payload when it grades.

That row reference is an implicit coupling, so the scorer refuses to grade a row whose `question_id` does not match the instance. Grading against another problem's tests would score every solution wrong while still producing a plausible-looking number. Alignment was checked across both releases: 612 and 443 instances, all unique ids, zero mismatches.

## Grading

A port of the reference harness, staged into the container as a script, following the existing `harness/sandbox/scripts/` + `get_script()` pattern.

- **Call-based problems** (LeetCode) invoke a named function on a `Solution` instance and compare parsed values; a tuple result is not counted as a mismatch, since ground truth is never a tuple.
- **Whole-program problems** get stdin fed and stdout compared line by line, with a decimal fallback so numerically equal output is not failed over formatting. `if __name__ == '__main__'` guards are unwrapped so the program can be called per test case.
- Compile failures, missing entry points, wrong answers and timeouts are distinguished in the recorded error code. A solution that does not compile is a failed solution, not a failure of the grader.
- **The solution runs in a forked child**, as in the reference, under the reference's overall budget of `(timeout + 1) * num_tests + 5` seconds. A solution that segfaults, is OOM-killed, or silences the per-case alarm is a failed solution, not a missing verdict. Per-case timeouts use `setitimer`, so sub-second budgets are enforced rather than dropped.
- **A grading run that cannot be scored raises.** No verdict (no `python3` in the image, a killed container), a payload that decoded to zero test cases, or an error inside the grader itself raises a `SandboxInfrastructureError` subclass, which the runner records against the task as lost scores. Previously an undecodable payload scored 1.0 (every solution passes an empty test set) and a missing verdict scored 0.0 with a warning.
- **Answer extraction takes the last fenced block**, as the reference does. A reasoning model writes code while it thinks, so earlier blocks are attempts it discarded. The shared `extract_code` helper takes the first block and falls back to the whole reply, which reliably extracts a draft from such a model; it also cannot strip `<think>` blocks when a chat template supplied the opening tag and the reply only closes one. Taking the last block sidesteps both, which is why the reference strips no tags at all. This matches oe-eval's `extract_code` for this task, which also takes the last block, so it is a parity fix rather than a divergence. `extract_code` is unchanged, so `humaneval` and `mbpp` keep their behavior.

## Verification

- **Prompts byte-identical to oe-eval** across 61 real problems (52 starter-code, 9 stdin) for the `think_tags` and `tulu` regimes, checked against an independently constructed copy of oe-eval's template. The default and `paper` prompts differ from those only in the reasoning line and the presence of the system message, both covered by unit tests.
- **Grader correct on real dataset problems** in both modes: correct solutions pass, wrong answers and missing functions fail with the right codes, `__main__` guards unwrap, infinite loops time out.
- **End to end in Docker** (mock model, `release_v3`): the grader ran in the container, decoded the compressed blobs there — reporting 14, 14 and 15 cases for the three problems, the real per-problem counts — and returned verdicts the scorer turned into pass@1.
- **Integration test** (`tests/integration/test_livecodebench_sandbox.py`, skipped without Docker) covers what stubs cannot: correct solutions to a whole-program and a call-based problem both score 1.0 through the real staging path, a wrong one scores 0.0, and the grader in the container saw more test cases than the row's public ones.
- **A real model on GPU** (`Qwen/Qwen2.5-Coder-7B-Instruct`, vLLM on one L40, `livecodebench:lite`, 50 problems of release_v3): **pass@1 = 0.4200**, 50/50 instances scored in 51s.

  The verdict breakdown is what makes that number trustworthy rather than merely plausible:

  | outcome | count |
  |---|---|
  | passed | 21 |
  | wrong answer (-2) | 26 |
  | timeout (-3) | 2 |
  | compile error (-4) | 1 |

  Failures are overwhelmingly wrong answers, with a single compile error. A run whose failures are mostly compile errors means extraction is broken, not that the model is weak — which is exactly what an earlier run with a different model showed before the extraction fix above (37 compile errors, 13 with no verdict, pass@1 = 0.0). All 50 replies contained a fenced block and all 50 arrived as clean text.
- 44 unit tests for prompts, release selection, variants, extraction, the scorer, and the grader, including the zero-test, crashed-interpreter, disarmed-alarm and sub-second-timeout cases. Full suite: 2252 passed, 8 skipped.
- Test-case counts across both releases, from the decoded payloads: release_v3 has a median of 15 and a maximum of 103 cases per problem, v4–v6 a median of 42 and a maximum of 52. The reference budget for the largest problem is 726s, under the scorer's 900s ceiling.

## Still open

**Parity is not yet demonstrated.** The run above proves the pipeline is sound end to end, but I have no oe-eval number for this model on release_v3 to compare against. A true parity check needs the same model and regime through both harnesses; cheap to do now that the path works.

## Unrelated issues noticed while testing

None of these are caused by this PR, and all three cost a failed launch:

- `ai2/oe-base` is **deprecated**. It is the Beaker launcher's built-in default and appears in the README seven times, so launching from the documented example fails with `[code=422] deprecated budget`.
- `job_assembler.py:470` derives the vLLM venv layout from the **harness** provider kind while the runtime uses the **model preset's** kind. For a `kind: vllm` preset with a sandbox harness, vLLM is installed in an isolated venv the worker cannot import from, and the job dies at startup. This affects 10 of the 11 vLLM model presets and cannot be worked around from the CLI — a raw HuggingFace model id has to be passed instead of the preset, which is what this PR's testing did.
- The `codex_python` harness preset cannot start locally: its image has no `swerex-remote`, and the preset sets neither `inject_swerex` nor a startup timeout above 60s, so every instance fails. `codex_universal` sets both. Worked around with an inline sandbox override rather than changing the preset here.

### DIFF
--- a/src/olmo_eval/common/scorers/__init__.py
+++ b/src/olmo_eval/common/scorers/__init__.py
@@ -22,7 +22,12 @@ from .citation import (
     score_citations_for_sections,
 )
 from .code_execution import CodeExecutionScorer, MultiplEScorer
-from .execution import ContextScorer, ExecutionScorer, SandboxRequiredError
+from .execution import (
+    ContextScorer,
+    ExecutionScorer,
+    SandboxRequiredError,
+    ScoringIncompleteError,
+)
 from .ifeval import IFEvalScorer
 from .llm_judge import (
     JudgeFn,
@@ -74,6 +79,7 @@ __all__ = [
     "RubricJudgeScorer",
     "SafetyScorer",
     "SandboxRequiredError",
+    "ScoringIncompleteError",
     "score_citation_group",
     "score_citations_for_sections",
     "Scorer",
--- a/src/olmo_eval/common/scorers/code_execution/scripts/__init__.py
+++ b/src/olmo_eval/common/scorers/code_execution/scripts/__init__.py
@@ -1,0 +1,19 @@
+"""Python programs that are staged into a sandbox and run there.
+
+These are loaded as text rather than imported: they execute inside the
+container, against its interpreter, not in the harness process.
+"""
+
+from importlib import resources
+
+
+def get_script(name: str) -> str:
+    """Load a sandbox script by name.
+
+    Args:
+        name: Script name without the .py extension (e.g., "livecodebench_grader").
+
+    Returns:
+        The script source as a string.
+    """
+    return resources.files(__package__).joinpath(f"{name}.py").read_text()
--- a/src/olmo_eval/common/scorers/code_execution/scripts/livecodebench_grader.py
+++ b/src/olmo_eval/common/scorers/code_execution/scripts/livecodebench_grader.py
@@ -1,0 +1,403 @@
+"""Grade one LiveCodeBench solution against a problem's test cases.
+
+Runs inside a sandbox, not in the harness process. Reads ``problem.json`` and
+``solution.py`` from its own directory and prints a single JSON verdict to
+stdout.
+
+Problems come in two shapes. A problem with a function name expects the
+solution to define that function (on a ``Solution`` class for LeetCode
+problems); its inputs and outputs are JSON values compared directly. A problem
+without one is a whole program: inputs are fed to stdin and the captured
+stdout is compared line by line, falling back to decimal comparison so that
+numerically equal output is not rejected over formatting.
+
+Ported from the reference implementation:
+https://github.com/LiveCodeBench/LiveCodeBench/blob/main/lcb_runner/evaluation/testing_util.py
+"""
+
+import ast
+import base64
+import faulthandler
+import json
+import os
+import pickle
+import signal
+import sys
+import time
+import zlib
+from decimal import Decimal
+from io import StringIO
+from types import ModuleType
+from typing import Any
+from unittest.mock import mock_open, patch
+
+# Prelude the reference harness prepends to every solution, so that solutions
+# relying on names from a bare `from x import *` still resolve.
+IMPORT_PRELUDE = (
+    "from string import *\nfrom re import *\nfrom datetime import *\n"
+    "from collections import *\nfrom heapq import *\nfrom bisect import *\n"
+    "from copy import *\nfrom math import *\nfrom random import *\n"
+    "from statistics import *\nfrom itertools import *\nfrom functools import *\n"
+    "from operator import *\nfrom io import *\nfrom sys import *\nfrom json import *\n"
+    "from builtins import *\nfrom typing import *\nimport string\nimport re\n"
+    "import datetime\nimport collections\nimport heapq\nimport bisect\nimport copy\n"
+    "import math\nimport random\nimport statistics\nimport itertools\n"
+    "import functools\nimport operator\nimport io\nimport sys\nimport json\n"
+    "sys.setrecursionlimit(50000)\n"
+)
+
+
+class TimeoutException(Exception):
+    """Raised by the alarm handler when a test case runs too long."""
+
+
+def timeout_handler(signum, frame):
+    raise TimeoutException
+
+
+class Capturing(list):
+    """Collect everything a block writes to stdout."""
+
+    def __enter__(self):
+        self._stdout = sys.stdout
+        sys.stdout = self._stringio = StringIO()
+        return self
+
+    def __exit__(self, *args):
+        self.append(self._stringio.getvalue())
+        del self._stringio
+        sys.stdout = self._stdout
+
+
+class MockBuffer:
+    """Byte-oriented view of the stdin stand-in."""
+
+    def __init__(self, inputs):
+        self.lines = inputs.split("\n")
+        self.index = 0
+
+    def read(self, *args):
+        return "\n".join(self.lines).encode()
+
+    def readline(self, *args):
+        if self.index >= len(self.lines):
+            return b""
+        line = self.lines[self.index]
+        self.index += 1
+        return (line + "\n").encode()
+
+
+class MockStdin:
+    """Stand-in for sys.stdin that also exposes a .buffer."""
+
+    def __init__(self, inputs):
+        self.stringio = StringIO(inputs)
+        self.buffer = MockBuffer(inputs)
+
+    def read(self, *args):
+        return self.stringio.read()
+
+    def readline(self, *args):
+        return self.stringio.readline()
+
+    def readlines(self, *args):
+        return self.stringio.readlines()
+
+    def __getattr__(self, name):
+        return getattr(self.stringio, name)
+
+
+def clean_if_name(code):
+    """Drop an ``if __name__ == '__main__'`` guard, keeping its body."""
+    try:
+        tree = ast.parse(code)
+        last_block = tree.body[-1]
+        if isinstance(last_block, ast.If):
+            condition = last_block.test
+            if ast.unparse(condition).strip() == "__name__ == '__main__'":
+                code = ast.unparse(tree.body[:-1]) + "\n" + ast.unparse(last_block.body)
+    except Exception:
+        pass
+    return code
+
+
+def make_function(code):
+    """Wrap a whole program in a function so it can be called per test case."""
+    try:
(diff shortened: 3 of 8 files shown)


abhishekraok (commented):
Claude Code · https://claude.ai/code/session_01RcHPnv8oDU2PHDofoNBQDz

Reviewed at `f0c765f` (the LiveCodeBench commits are unchanged from the original push; only the base moved). ruff clean, `ty` at 11 diagnostics (same as `main`), full unit suite passes.

**Parity, verified independently against oe-eval and the dataset's loading script:** templates and sampling params for `::olmo3:adapt`, `::tulu`, `:grpo::tulu-thinker` match `tasks.py` character-for-character; the last-block extraction matches oe-eval's `extract_code` (`indexlines[-2]:indexlines[-1]`), so the second commit is a parity fix, not a divergence — worth saying so in the body; `RELEASE_V3_FILES` / `RELEASE_V4_V6_FILES` match `ALLOWED_FILES["release_v3"]` / `["v4_v6"]`, order preserved.

### Major — each turns a grading failure into a plausible number

**1. A test payload that fails to decode scores 1.0.** `decode_test_cases` returns `[]` on any exception, both grade functions return `True` on an empty loop, and the scorer reads only `passed`. Reproduced with the grader as shipped:
```
undecodable blob, call-based:           {"passed": true, "num_tests": 0}
undecodable blob, stdio, wrong program: {"passed": true, "num_tests": 0}
```
The grader already reports `num_tests`; have the scorer treat `num_tests == 0` as a grading failure. oe-eval has the same hole (`all(r is True for r in [])`), but reading the row lazily in the container adds a new way to reach it — a `MemoryError` decoding a 92MB blob lands in that bare `except Exception`.

**2. `timeout` below 1s disables the per-case timeout entirely.** `int(problem.get("timeout", 6))` → `signal.alarm(0)`, which *cancels* the alarm. `LiveCodeBenchScorer.timeout` is a `float` field, so `timeout=0.5` looks legal. Reproduced: an endless loop with `timeout=0.5` never terminates; `timeout=2` correctly gives `-3`. `signal.setitimer`, or `max(1, int(...))`.

**3. A missing verdict is scored 0.0, not raised.** No `python3` in the image, an OOM-killed grader, a segfault (plausible — the prelude sets `setrecursionlimit(50000)` and, unlike oe-eval's forked child, the solution runs in the grader's own process) → `_parse_verdict` is `None` → 0.0 with a `logger.warning`. A misconfigured image publishes a clean pass@1 of 0.0 for the whole run. The row-alignment `RuntimeError` already sets the precedent; raising here engages `max_hard_failure_rate` instead.

### Should fix before a real run

**4. `ascore` does its heavy lifting on the event loop.** `_test_case_rows(...)`, the Arrow row read, and `json.dumps(problem)` are synchronous inside `async def`. Measured `json.dumps` alone: 75 ms at a 20MB payload, ~490 ms at the 92MB worst case — and swerex re-serializes the same string into the request body, so roughly double. Over 612 × 10 samples that's minutes of aggregate loop stall blocking every in-flight inference request. `asyncio.to_thread` for the read + dumps.

### Minor

**5.** `overall_timeout` is a flat 900s where oe-eval uses `(timeout+1)*num_tests + 5`; a correct-but-slow solution on a problem with >~150 cases is killed here and allowed there.
**6.** `rm -rf` rides on the command string, so it doesn't run when the outer timeout kills the command — up to a few hundred MB of `problem.json` leaks per timeout on a pooled executor. Prepend a sweep of stale `/tmp/lcb-*`.
**7.** `zip(inputs, outputs, strict=False)` silently truncates a length mismatch (matches upstream; combined with #1 an under-length `outputs` quietly shrinks the test set). A module-level `while True` in call-based mode reports `-4` rather than `-3`.
**8.** `test_private_test_cases_reach_the_container` asserts only that `'"num_tests"'` appears in the output, not that the count exceeds the public cases, which is what its docstring claims. Nothing covers the zero-test case.

1–3 before merge; 4 before the first vLLM run. The `codex_python` preset issue in the body deserves its own issue so it doesn't get buried.

gregorybchris:
@abhishekraok thanks for your review! I've addressed your comments and run an extra agentic review for good measure. Ready for your eyes again when you get a sec

abhishekraok (approved):
Codex · codex://threads/01a0881f-9448-77b0-aa9c-77a3e437171a

Approved as requested. CI is passing; Docker and GPU integration checks were skipped.

mgmorgan23:
I'm not sure if we want to match exactly, but I just want to flag that I think the Olmo3 paper used the `livecodebench_codegeneration::tulu-thinker_deepseek_no_think_tags `formulation from oe-eval, with sampling parameters set to 
`"generation_kwargs": {"max_gen_toks": 999999, "truncate_context": false, "do_sample": true, "top_p": 0.95, "temperature": 0.6, "repeats": 10, "stop_sequences": []}, "metric_kwargs": {"pass_at_ks": [1, 5, 10]}},`

mgmorgan23 (approved):

PR #281, opened 2026-08-14
Add HELMET task suite: all seven categories at 4k-128k

jopetty:
> **Scope change (2026-09-03):** per review, this PR is now the port of HELMET at its own 4k–128k context lengths. The long-context recall extension (256k–2m json_kv tiers) has been split out to a follow-up PR based on this branch. The base of the branch is unchanged; the split is a single subtractive commit on top.

Adds support for [HELMET](https://github.com/princeton-nlp/HELMET) (ICLR 2025), the long-context benchmark: **120 tasks across all seven categories** at HELMET's standard 4k–128k lengths.

| category | tasks | primary metric |
|---|---|---|
| recall | `json_kv` | substring recall |
| rag | `kilt_{nq,triviaqa,hotpotqa,popqa}` | substring exact match |
| rerank | `msmarco_rerank_psg` | NDCG@10 |
| longqa | `infbench_{qa,choice}_eng`, `narrativeqa` | ROUGE-L F1 / EM / GPT-4 judge |
| summ | `infbench_sum_eng`, `multi_lexsum` | GPT-4 judge (fluency-gated F1) |
| icl | `trec_{coarse,fine}`, `banking77`, `clinic150`, `nlu` | HELMET-normalized EM |
| cite | `alce_{asqa,qampari}` (+`_nocite`) | str_em / qampari_rec_top5 |

Suites: `helmet_{category}__{size}`, `helmet_all__{size}` (mean over category averages), and `helmet_nojudge__{size}` (excludes the three LLM-judged tasks, following the `science:judge`/`science:nojudge` split).

**Data**: `allenai/helmet-plus` on the Hub hosts HELMET's pre-retrieved KILT/MS MARCO/ALCE files (re-hosted unpacked from its 10GB `data.tar.gz` so consumers fetch per-task files), the summarization keypoints, the json_kv tiers, and per-directory manifests so loaders never hardcode filenames. Downloads are pinned to a fixed dataset revision. The json_kv tiers are regenerated with HELMET's own `generate_json_kv_data.py` (same generator and parameters, calibrated against the Olmo 3 tokenizer), so they match HELMET's distribution but not its exact instances. LongQA/ICL sources load from the Hub directly; three ICL datasets are repointed to parquet mirrors because `datasets` ≥4 dropped script loading (split sizes and label counts verified against HELMET's hardcoded values).

[...]

PR #356, opened 2026-09-03
Extend HELMET recall (json_kv) to 256k-2m context lengths

jopetty:
Follow-up to #281, which ports HELMET at its standard 4k–128k lengths. This PR adds the long-context extension: four extra `json_kv` recall tiers at **256k, 512k, 1m, and 2m** tokens, generated with HELMET's own `generate_json_kv_data.py` and calibrated against the Olmo 3 tokenizer. Only the synthetic recall task extends; every other HELMET category is capped by real documents, fixed retrieval depth, or finite demonstration pools.

The data is already on `allenai/helmet-plus` (the pinned revision in #281 includes it), and the json_kv loader is manifest-driven, so no loader changes are needed. The diff is the tier list, the docstrings that describe the ragged length coverage, and the tests for it.

Suites: `helmet_recall__{262144,524288,1048576,2097152}`. No `helmet_all__*` is registered above 128k, since it would silently be recall alone and break the trend line against shorter lengths.

Stacked on #281: until that merges, this PR shows its commits too. The extension itself is the last commit on the branch (5 files, +41/-21).


gregorybchris (commented):

Looks good overall! The extension is small and well scoped, and leaving `helmet_all__*` unregistered above 128k is the right call. Nothing blocks the merge. I have one structural suggestion, one question about token headroom, and a request for run evidence.

Scope: this review covers only the extension commit (`51b64c91`, i.e. `git diff pr-281-head..pr-356-head`, 5 files). The #281 content is reviewed on #281. I ran `ruff check`, `ruff format --check`, `ty check`, and the HELMET, registry, and suite tests on the PR head, and all pass. The one `ty` warning (an unused ignore in `harness/sandbox/executor.py`) also appears on `main` and is unrelated. CI is green.

### Non-blocking

1. `src/olmo_eval/data/helmet_tasks.py:244`
   The fallback in `base_config.get("context_sizes", CONTEXT_SIZES)` now means "all ten tiers, up to 2m". Before this PR, `CONTEXT_SIZES` equaled the standard list, so the fallback was harmless. Now a future base task that forgets `context_sizes` silently gets 256k-2m tasks. I reproduced this by adding a base `infbench` config with no `context_sizes`: it generated `hypothetical_new__262144` through `hypothetical_new__2097152`, with `max_context_tokens=2096942` at the 2m tier. For a book-backed task, that tier would hold the same untruncated book as a shorter one but carry a 2m label. As a second effect, a second category above 128k would register a `helmet_all__262144` suite, which is exactly the misleading trend line the suite docstring warns about. Could the fallback be the standard sizes, so that reaching past 128k is always an explicit opt-in?
   ```python
   for size in base_config.get("context_sizes", STANDARD_CONTEXT_SIZES):
   ```
   A test that locks in the invariant would also help, for example asserting that every task above `131072` has `tag == "recall"`.

2. `src/olmo_eval/data/helmet_tasks.py:98`
   Question about headroom. The KILT tiers budget `size - max_gen_toks` for the prompt, and LongQA reserves `200 + max_gen_toks`. `json_kv` has no reserve, and the docstrings say its tiers are calibrated to hit their nominal lengths under the Olmo 3 tokenizer. If the 2m prompts really are about 2,097,152 tokens, then prompt plus the 100 generation tokens exceeds the window of a model whose max length is exactly 2m. vLLM would reject those requests. At 1m and 2m the model window often equals the tier, so this matters more here than at 128k. What target did the calibration use: `size` or `size - max_gen_toks`? I couldn't check this myself because `allenai/helmet-plus` is private and I don't have access.

3. Evidence request (PR description)
   The description doesn't say what was run. I couldn't confirm that the pinned revision `f46550958fafdff9340d4c24c50a98aaccd5d202` has `256k`, `512k`, `1m`, and `2m` entries in `json_kv/manifest.json`, again because the dataset is private. Before merge, could you paste:
   - the four new manifest entries at the pinned revision,
   - min/max Olmo 3 token counts for the rendered prompts at each new tier (this also answers item 2),
   - one completed run of `helmet_recall__262144`, and ideally `helmet_recall__2097152`, with the download size and peak memory of the loader. The loader streams and applies `limit` while reading, which is good. At 2m tokens, though, 100 instances are on the order of a gigabyte of prompt text held in memory, so a real figure would be useful.

### Nits

- Nit: `src/olmo_eval/data/helmet_tasks.py:33`: "the tiers added by the long-context extension" describes this PR more than the code. Something like "Standard HELMET lengths, and the longer tiers that only json_kv supports" will still read correctly after merge.

### Before merge

Nothing here blocks. I'd like to see the evidence in item 3 posted in the thread, and item 1 is a one-line change worth taking. Once #281 merges and this branch is rebased onto `main`, you may merge after CI is green.

gregorybchris (approved):

PR #358, opened 2026-09-08
State bench lc clean

jopetty:
Adds support for state-bench, a suite of long-context state tracking evaluations.


gregorybchris (commented):

Thanks for this! The per-config stratum table and the skipped empty suites are careful work. A few things we should address before merging.

I found two blocking issues. Both affect whether a stored StateBench result means what it says. The first is about prompts longer than the model's context. The second is about the unpinned dataset. The non-blocking items cover portability, memory, reproducibility of the 10% sample, structure, and missing parity and test evidence.

Lint, format, and the unit tests pass locally and in CI. `ty` reports one warning in `harness/sandbox/executor.py:788`, and `main` has the same warning, so it does not come from this PR.

### Blocking

1. `src/olmo_eval/evals/tasks/state_bench_lc.py:184` (with `src/olmo_eval/inference/providers/vllm.py:401-404`)
   `format_request` passes `max_length=self.config.max_length`, which is `None` for every task. The vLLM provider then falls back to the model's `max_model_len`. It left-truncates any longer prompt to fit and scores the continuations as usual. The provider computes `overflow` but never records it in the output. So `state_bench` on a 128k model runs the `tokens_256k`, `tokens_512k`, and `tokens_1m` tasks with the start of each prompt removed. For a state-tracking task, the start of the prompt holds the initial state. Those rows get stored as ordinary accuracy numbers near chance, and the stored result cannot show that the prompt was truncated. The full `state_bench` suite includes every stratum, so this is the default outcome for most models.
   Part of this is a gap in the framework: silent left-truncation is the right default for short MC tasks, but no provider reports it. Could we do both of these?
   - Framework: record the truncated token count in the output metadata (for example `prompt_truncated_tokens`) in each logprob provider, so it survives into saved predictions.
   - This PR: make over-context StateBench instances fail visibly. One option is to raise when a prompt exceeds the provider limit. Another is to document that users should select stratum suites (`state_bench:tokens_64k`) within their model's context. The class docstring or the suite descriptions should say so either way.
   A regression test that runs an over-length prompt through the mock provider and checks the recorded flag would lock this in.

2. `src/olmo_eval/evals/tasks/state_bench_lc.py:209-213`
   The `DataSource` has no `revision`, so the task hash (`data_source.revision: None`) stays the same when the Hub data changes. This PR's own history shows the data changing: the repo moved from `jacksonp-ai2/state-bench-lc` to `allenai/state-bench`, and the strata were re-cut twice. `STATE_BENCH_CONFIGS_BY_STRATUM` (line 60) also hard-codes which splits exist, and that is only true for one staging of the dataset. Please pin `revision=` to the commit these tasks were built against. Future re-stagings should then land as a new revision, bumped on purpose.
   ```python
   STATE_BENCH_REVISION = "<commit sha>"
   ...
   DataSource(STATE_BENCH_REPO, subset=_config_name, split=_dataset_split, revision=STATE_BENCH_REVISION)
   ```

### Non-blocking

1. `src/olmo_eval/evals/tasks/state_bench_lc.py:15`
   An anonymous request to `https://huggingface.co/api/datasets/allenai/state-bench` returns 401, so the dataset is private or gated right now. The old `jacksonp-ai2/state-bench-lc` copy also returns 401. Is it going public before merge? If not, could the module docstring say the data is not yet public, with a follow-up issue filed to track the release? Otherwise 342 registered tasks fail at load for anyone outside Ai2.

2. `src/olmo_eval/evals/tasks/state_bench_lc.py:193`
   `StateBench10Percent` decodes the whole split into a list before sampling. `_load_instances_cached` keeps that list for the life of the task. The runner's `limit` is also applied only after a full load (`runners/asynq/preparation.py:56-60`), and that part is a framework limitation. A 1M-token prompt is several MB of text, so the `tokens_1m` splits may be large. Could you share the peak memory for `state_bench:tokens_1m` (and its `_10pct` form)? I could not measure it without Hub access.

3. `src/olmo_eval/evals/tasks/state_bench_lc.py:195`
   The 10% sample uses a fixed `random.Random(42)` and ignores `config.seed`. I built a `_10pct` task with `seed=42` and another with `seed=7` over the same 50 instances. Both returned the same sample, even though `seed` is in `to_dict` and the two task hashes differ. Please use `self.config.seed`:
   ```python
   yield from random.Random(self.config.seed).sample(instances, sample_size)
   ```

4. `src/olmo_eval/evals/tasks/state_bench_lc.py:198-237`
   The two `type(...)` plus `globals()[...]` blocks are near copies of each other. Could we switch to `__init_subclass__` as `LabBenchTask` does (`src/olmo_eval/evals/tasks/lab_bench.py:123`), or use one helper that builds both classes? Also, `register()` already adds dynamically created classes to the module namespace (`evals/tasks/common/registry.py:64-66`), so the `globals()` assignment is not needed.

5. `src/olmo_eval/evals/tasks/state_bench_lc.py:143-148`
   `process_doc` returns `None` for a row with a missing prefix, bad choices, or an out-of-range gold index. The loader drops those rows without logging anything. The denominator then shrinks without notice. Since this dataset is curated, should a malformed row raise `ValueError` with the `example_id` instead?

6. Parity and testing evidence
   The description does not name a reference implementation or say what was run. Could you add:
   - the reference the scoring matches (paper or official repo), and why per-token normalized MC accuracy (`LogprobPerTokenMCAccuracyMetric`) is the right metric;
   - a small table with reference numbers beside this port's numbers for one model on a few strata;
   - confirmation that the `" {choice}"` leading space matches how the reference joins `prefix` and choice.

7. Tests
   There are no tests. Could you add unit tests for `process_doc` (valid row, each malformed case), `format_request` continuations, the 10% sample's determinism and size, and the stratum helpers (`state_bench_configs_for_stratum`, `state_bench_strata_for_config`, and the fact that no suite is registered empty)?

### Nits

- Nit: `STATE_BENCH_OMITTED_SPLITS` (line 97) and `STATE_BENCH_CONFIGS_SET` outside its helpers have no consumers. Could we drop the first until something reports coverage?
- Nit: The module is `state_bench_lc`, but the tasks, suites, and dataset are all `state_bench`. Since this replaces the `-lc` copy, should the modules be renamed to `state_bench.py`?
- Nit: The PR description says "family" and "regime", while the code says `_formatter` and `_complexity`. Using one set of terms would make the suite code easier to follow.
- Nit: `dataset_split` duplicates the split on `data_source`. It exists only because `get_data_source` overwrites a `DataSource`'s split with `config.split`. A one-line comment saying so would stop someone from removing it. Note that `to_dict` still records `split: "test"`.
- Nit: The suite module repeats the same full/10% pair of `make_suite` calls four times. A small helper that registers both would shorten it.

### Before merge

Please pin the dataset revision, and make over-context truncation visible for StateBench. Recording it in the provider output is the preferred fix. Please also answer the dataset-visibility question and add the parity table. After that, you're welcome to merge once CI is green. If the provider-side truncation flag is deferred, please file an issue for it ("Record prompt truncation in logprob provider outputs") and link it here.

jopetty:
Thanks for the review! Here's where things stand on the blocking items and the evidence you asked for.

### Blocking items

1. **Over-context truncation is now visible.** The vLLM, vLLM-server, and OLMo-core logprob providers record the number of dropped prompt tokens as `prompt_truncated_tokens` in each output's metadata, so it survives into saved predictions (6b3d6199). The HF provider never truncates, so it is unchanged. On the StateBench side, the task docstring and the six suites that span every stratum now say to select stratum suites within the model's context (932d1e92). A regression test drives an over-length prompt through the OLMo-core logprob path and checks the recorded count (98f69084). I used the OLMo-core path rather than the mock provider because the mock never truncates, so it cannot exercise the flag.
2. **Dataset revision is pinned.** `STATE_BENCH_REVISION = "683d249128a0d387eda27c8cfc8c1f9df0389495"` is passed to every `DataSource`, so future re-stagings change the task hash (932d1e92).

### Dataset visibility

`allenai/state-bench` is private to Ai2 for now.

### Run evidence

StateBench is new with this work, so there is no external harness to compare against. Instead, here is a full run of the `state_bench:tokens_16k` suite on the public [`allenai/Olmo-Hybrid-7B`](https://huggingface.co/allenai/Olmo-Hybrid-7B) (Beaker experiment `01M12DFTQ3DRDA8VMQ57188DT7`, vLLM, `max_model_len=18432`). Metric: `accuracy:logprob`. Each item has 5 choices, so chance is 0.20. There are 96-100 instances per task.

| family | aperiodic | periodic | r-trivial |
|---|---|---|---|
| `cube-painting` | 0.1979 | 0.1979 | 0.9688 |
| `integer-code` | 0.4200 | 0.1700 | 1.0000 |
| `people-in-rooms` | 0.4271 | 0.1959 | 1.0000 |
| `ruler` | 0.5000 | 0.1600 | 1.0000 |
| `spreadsheet-cells` | 0.2500 | 0.1667 | 0.9375 |
| `status-lights` | 0.3229 | 0.1667 | 0.9688 |
| **mean** | **0.353** | **0.176** | **0.979** |

Suite average (`state_bench:tokens_16k`): **0.503**.

`r-trivial` configs are near ceiling, `periodic` configs sit at chance, and `aperiodic` falls in between. Notes on provenance:
- The run predates the repo move and loaded `jacksonp-ai2/state-bench-lc`. I checked that all 43 `tokens_16k` parquet files there have the same LFS sha256 as `allenai/state-bench` at the pinned revision, so the data is identical.
- Across all 2,028 `tokens_16k` rows, the longest prompt is 16,383 tokens under the Olmo 3 tokenizer, inside the 18,432-token window, so nothing in this run was truncated.
- Each continuation is scored as the choice text after a single space. StateBench's own tooling defines no eval-time joiner. Its token-count utility takes the joiner as a parameter, and its training-document form joins prompt and completion with a newline.

The non-blocking items and nits are not addressed in this push. I'll follow up on them separately.

PR #359, opened 2026-09-08
Add FrontierScience Olympiad multiple-choice eval for base-model tracking

donovanr:
## Description

Adds `frontierscience_olympiad:mc` and the base-model alias `frontierscience_olympiad:mc:olmo3base`. The existing free-response `frontierscience_olympiad` task is unchanged.

Four curated choices per problem, scored by the log-likelihood of the labels `A`–`D`. Th MC scoring does not use a judge, so it is cheap and easy for per-checkpoint tracking.

Our MC variant is generated using the correct answers to the original Olympiad questions, and three synthesized plausible distractors.

> [!NOTE]
> **In working through the 100 original questions to verify the accepted answers and generate plausible wrong answers, 17 out of the 100 original questions were flagged as either ambiguously worded or having wrong gold answers. I am leaving these out of MC variant for now.**

Beyond this excision, scores on the MC variant are not comparable with published FrontierScience accuracy; since the MC task measures recognition among curated distractors, not free-response generation, and random guessing gives a floor of 25% accuracy.

## What's in it

**Task** (`src/olmo_eval/evals/tasks/frontierscience.py`). Strips out the benchmark's free-response answer-format instruction and builds a completion-style prompt:

```
Problem:
{problem}

Choices:
A. {choice_a}
B. {choice_b}
C. {choice_c}
D. {choice_d}

Answer:
```

One `LOGLIKELIHOOD` request with continuations `" A"`–`" D"`. Rows are validated on load: four nonempty choices, gold index in range, index/label agreement, `reference_answer` matching the gold choice text, known subject.

**Dataset** (`src/olmo_eval/evals/tasks/frontierscience_olympiad_mc.jsonl`). 83 of the 100 Olympiad problems, versioned here. The 17 with unresolved source-answer concerns are excluded. That leaves counts of:

- physics: 40
- chemistry: 34
- biology: 9

Gold labels are balanced across MC option: A 21, B 20, C 21, D 21. Each row carries `distractor_provenance` naming the error mechanism that is used to generate each distractor.

**Metrics**, overall and per subject: `accuracy`, `gold_probability` (softmax mass on gold), `logprob_margin` (gold minus best incorrect). The continuous two are useful because 83 items makes accuracy coarse — binomial SE is about ±5.3%.

**Tests**: registration, dataset invariants, prompt/continuation construction, and the three metrics with subject slices.

**Dependency pins** in `pyproject.toml`. `huggingface-hub>=1.10.0`, because `ruler_loader.py` imports `silent_tqdm` and Beaker's unlocked resolve could otherwise install a hub version without it. `transformers<5.13` in the `vllm` and `hf` extras, because 5.13 nests `rope_parameters` per layer type while vLLM 0.19.1 reads it flat, which breaks Olmo-3. Both are repo-wide fixes, not specific to this task. 

## Results

Chance is 0.250, the implied binomial SE is about ±0.05.

| model | accuracy | gold prob | margin | physics | chemistry | biology |
|---|---:|---:|---:|---:|---:|---:|
| Qwen3-8B | 0.325 | 0.333 | −0.760 | 0.325 | 0.324 | 0.333 |
| Qwen2.5-7B | 0.349 | 0.293 | −0.266 | 0.250 | 0.471 | 0.333 |
| Olmo-3-7B `stage2-step47684` | 0.386 | 0.287 | −0.292 | 0.375 | 0.353 | 0.556 |
| Qwen3.8-27B | 0.542 | 0.451 | +0.331 | 0.550 | 0.559 | 0.444 |

Qwen3.8-27B results are about 6 SE above chance and have the only positive margin; the 7–8B models sit within about 1.5 SE of chance. That ordering is a decent sanity check: the task separates a strong model from weak ones, and the distractors are strong enough that small models fall back on label priors (Olmo3 likes `C`) rather than beating chance.

A couple of caveats on the table:

- Raw accuracy at 7–8B is largely label position bias — Olmo-3 puts 48 of 83 picks on C, scoring 81% on C-gold items and 24% on the rest, so its 6pp margin over Qwen3-8B is not meaningful.
- `margin` is probably the most useful metric: negative means the model isn't separating gold from distractors at all.

### Olmo-3-7B

```bash
uv run olmo-eval beaker launch \
  -n frontierscience-mc-olmo3-7b \
  -m olmo-3-1025-7b \
  -t 'frontierscience_olympiad:mc:olmo3base' \
  -c titan -p urgent -T 2h -w ai2/olmo-eval-debug \
  -H default -o provider.kind=vllm \
  -o provider.kwargs.attention_backend=FLASH_ATTN \
  -o 'provider.dependencies=[nvidia-cutlass-dsl==4.4.2,quack-kernels==0.4.0]' \
  --no-follow -y
```

### Qwen3-8B

```bash
uv run olmo-eval beaker launch \
  -n frontierscience-mc-qwen3-8b \
  -m Qwen/Qwen3-8B \
  -t 'frontierscience_olympiad:mc:olmo3base' \
  -c titan -p urgent -T 2h -w ai2/olmo-eval-debug \
  -H default -o provider.kind=vllm \
  -o provider.kwargs.attention_backend=FLASH_ATTN \
  -o 'provider.dependencies=[nvidia-cutlass-dsl==4.4.2,quack-kernels==0.4.0]' \
  --no-follow -y
```

### Qwen2.5-7B

```bash
uv run olmo-eval beaker launch \
  -n frontierscience-mc-qwen2.5-7b \
  -m qwen2.5-7b \
  -t 'frontierscience_olympiad:mc:olmo3base' \
  -c titan -p urgent -T 2h -w ai2/olmo-eval-debug \
  -H default -o provider.kind=vllm \
  -o provider.kwargs.attention_backend=FLASH_ATTN \
  -o 'provider.dependencies=[nvidia-cutlass-dsl==4.4.2,quack-kernels==0.4.0]' \
  --no-follow -y
```

### Qwen3.8-27B

```bash
uv run olmo-eval beaker launch \
  -n frontierscience-mc-qwen3.8-27b \
  -m Qwen/Qwen3.8-27B \
  -t 'frontierscience_olympiad:mc:olmo3base' \
  -c titan -p urgent -T 4h -w ai2/olmo-eval-debug \
  -H default -o provider.kind=vllm \
  -o provider.max_model_len=8192 \
  -o provider.kwargs.attention_backend=TRITON_ATTN \
  -o provider.kwargs.gdn_prefill_backend=triton \
  -o provider.kwargs.language_model_only=true \
  -o 'provider.dependencies=[nvidia-cutlass-dsl==4.4.2,quack-kernels==0.4.0]' \
  --no-follow -y
```

For this last config, we need `TRITON_ATTN` because `head_dim` is 256; `FLASH_ATTN` only works for `head_dim` of 128 on this image. `gdn_prefill_backend=triton` is needed for the Gated DeltaNet layers, `language_model_only=true` to skip the vision tower, and `max_model_len=8192` instead of the native 262k.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Integration tests pass (if applicable)

## Checklist


src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -634,6 +721,85 @@ async def _score_output(
         return scores
 
 
+@register("frontierscience_olympiad:mc")
+class FrontierScienceOlympiadMC(Task):
+    """Judge-free multiple-choice adaptation of the verified Olympiad subset."""
+
+    data_source = DataSource(path=FRONTIERSCIENCE_VERIFIED_MC_FILE, split="test")
```
gregorybchris, line 731: **Blocking (1).** The vendored dataset is not pinned by anything in the config.

This serializes to `{'path': 'frontierscience_olympiad_mc.jsonl', ..., 'revision': None}`. I truncated the JSONL to 50 rows and re-hashed: the task loaded 50 instances and the hash stayed `6b7c356ab07d67e9`. A later revision of the file — adding back some of the 17 excluded questions, or another normalization pass — produces a different eval that is indistinguishable from this one in storage.

`DataSource.revision` is already serialized into `to_dict()`, so the fix is small:

```python
FRONTIERSCIENCE_MC_RELEASE = "2026-09-v1"  # bump on any edit to the JSONL

data_source = DataSource(
    path=FRONTIERSCIENCE_VERIFIED_MC_FILE,
    revision=FRONTIERSCIENCE_MC_RELEASE,
)
```

Please add a test asserting that changing the release identifier changes the task hash. (The `split="test"` here is also overwritten by `get_data_source()` from the class-level `split = Split.TEST` — see the note on the `instances` property.)

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -318,9 +333,81 @@ def _subject_metrics(score_key: str) -> tuple[FrontierScienceMetric, ...]:
...
+            not isinstance(gold_idx, int)
+            or not response.outputs
+            or len(response.outputs) != len(choices)
+        ):
+            return None
+
+        scorer = self.scorer() if isinstance(self.scorer, type) else self.scorer
+        scores = [scorer.score(response.instance, output) for output in response.outputs]
+        if self.metric_kind == "accuracy":
+            return 1.0 if scores.index(max(scores)) == gold_idx else 0.0
```
gregorybchris, line 363: **Blocking (2).** Missing logprobs are scored as a correct answer.

`LogprobScorer` returns `-inf` when a provider returns no logprobs. With all four continuations at `-inf`, `scores.index(max(scores))` returns `0`, so every instance whose gold is A is credited correct. I reproduced this with a mock response carrying no logprobs: `compute_instance` returns `1.0` for a gold-A instance, `0.0` for a gold-C one, and `accuracy.compute([good_response, all_missing])` returns `1.0`.

A completely broken run over this set lands at 0.25 — chance — with 21 of 83 items marked correct. An infrastructure failure becomes a plausible-looking score.

This one is partly ours: `LogprobMCAccuracyMetric` in `src/olmo_eval/common/metrics/base.py:552` has the same argmax-over-all-`-inf` shape, so it is worth fixing in both places.

```python
if all(score == float("-inf") for score in scores):
    return None
```

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -318,9 +333,81 @@ def _subject_metrics(score_key: str) -> tuple[FrontierScienceMetric, ...]:
...
+        if self.metric_kind == "gold_probability":
+            largest = max(scores)
+            weights = [math.exp(score - largest) for score in scores]
+            return weights[gold_idx] / sum(weights)
+        if self.metric_kind == "logprob_margin":
+            return scores[gold_idx] - max(
+                score for index, score in enumerate(scores) if index != gold_idx
+            )
```
gregorybchris, line 371: **Blocking (3).** These two emit NaN and ±inf, and one bad instance poisons the task-level number.

Reproduced with one good response and one whose gold logprob is missing: `logprob_margin.compute_instance` returns `-inf`, and `logprob_margin.compute([good, that_one])` returns `-inf` for the whole task. With all four continuations missing, `gold_probability` returns `nan` per instance and `nan` in aggregate — `math.exp(score - largest)` is `nan` when both are `-inf`, and `sum(values) / len(values)` propagates it.

NaN is also not valid JSON, so it does not round-trip cleanly to the predictions file or to Postgres.

Please return `None` from `compute_instance` when the gold or the best distractor has no logprob, and have `compute` skip non-finite values.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -318,9 +333,81 @@ def _subject_metrics(score_key: str) -> tuple[FrontierScienceMetric, ...]:
...
+        raise ValueError(f"unknown FrontierScience MC metric kind: {self.metric_kind}")
+
+    def compute(self, responses: Sequence[Response]) -> float:
+        values = [
+            value
+            for response in responses
+            if (value := self.compute_instance(response)) is not None
+        ]
+        return sum(values) / len(values) if values else 0.0
```
gregorybchris, line 380: **Blocking (4).** The denominator here is the scorable subset, so a mostly-failed run reports a full-looking score.

With one scored response and two that returned no outputs, this metric gives `1.0` where `LogprobMCAccuracyMetric` on the same input gives `0.333` — the existing metric divides by `len(responses)`. Two thirds of the task failing should not read as 100%.

Either divide by `len(responses)` to match the existing metric, or record the scored count so the shortfall is visible in the result.

The `else 0.0` on line 377 is the other half of this: an all-failed run reports `0.0` rather than failing, which is the "all-failed run reported as Success with 0.0" pitfall.

Separately (non-blocking, finding 9): an empty subject slice also lands on that `0.0`. Biology is 9 of 83 rows, so a routine `limit=10` smoke run will often persist `accuracy_biology = 0.0`, which reads as a real zero. `SubsetAccuracyMetric` uses `-1` as its sentinel. This follows the existing module convention that `FrontierScienceMetric` documents at line 288, so a follow-up issue covering both metric families is fine rather than PR scope.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -701,3 +867,4 @@ async def _score_output(
     "paper",
     sampling_params=replace(_RESEARCH_SAMPLING, temperature=1.0, num_samples=30),
 )
+register_variant("frontierscience_olympiad:mc", "olmo3base")
```
gregorybchris, line 873: **Non-blocking (5).** This variant is a no-op alias, and its name is model-specific.

`register_variant` is called with no overrides, and `get_task` never rewrites `config.name` for a variant. I built both specs and compared:

```
base  name: frontierscience_olympiad:mc   hash: 6b7c356ab07d67e9
alias name: frontierscience_olympiad:mc   hash: 6b7c356ab07d67e9
identical config: True
```

So anyone who launched with `:olmo3base` — as all four commands in the description do — will find their results filed under `frontierscience_olympiad:mc`.

A model-named variant is also what §7 of the review guide rules out: variants describe a format or metric modifier, and model-specific presets compose from variants. Could we drop this line and launch the plain `:mc` spec?

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -634,6 +721,85 @@ async def _score_output(
...
+
+    data_source = DataSource(path=FRONTIERSCIENCE_VERIFIED_MC_FILE, split="test")
+    split = Split.TEST
+    metrics = MC_METRICS
+    primary_metric = MC_ACCURACY[0]
+    num_fewshot = 0
+    sampling_params = SamplingParams(temperature=0.0, max_tokens=1)
+
+    @property
+    def instances(self) -> Iterator[Instance]:
```
gregorybchris, line 739: **Non-blocking (6).** Because `instances` reads the file through `importlib.resources`, `config.data_source` is never consulted — a `-o data_source=...` override would be silently ignored, and `task._load_instances()` raises `FileNotFoundError: Dataset path not found: frontierscience_olympiad_mc.jsonl`.

If finding 1 is taken, `data_source` becomes the place the release identifier lives and this mostly resolves itself. Please drop the redundant `split=` on line 728 either way.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -318,9 +333,81 @@ def _subject_metrics(score_key: str) -> tuple[FrontierScienceMetric, ...]:
...
 
 
+@dataclass(frozen=True)
+class FrontierScienceMCMetric(Metric):
+    """A label-logprob metric, optionally restricted to one science subject."""
+
+    name: str = ""
+    scorer: type[Scorer] | Scorer = LogprobScorer
+    metric_kind: str = "accuracy"
+    subject: str | None = None
```
gregorybchris, line 346: **Non-blocking (7 and 8).** Two things on this class.

**Reuse.** `LogprobMCAccuracyMetric` at `src/olmo_eval/common/metrics/base.py:530` is the same argmax-over-`LogprobScorer` computation. The subject slices and the two continuous metrics are genuinely new, but the overall `accuracy` could reuse the existing metric so a future fix to the `-inf` handling lands once. Worth a comment explaining the divergence if you keep a separate class.

**`name` and `metric_kind` must agree but are independent.** `Metric.to_dict()` serializes only `type`, `name`, and `scorer`, so `FrontierScienceMCMetric(name="accuracy", metric_kind="gold_probability")` hashes identically to the real accuracy metric while computing something else. Deriving `name` from `metric_kind` and `subject` in `__post_init__` closes it.

`metric_kind` is also a fixed set — `Literal["accuracy", "gold_probability", "logprob_margin"]` would let `ty` catch the typo that line 369's `ValueError` currently catches at runtime.

src/olmo_eval/evals/tasks/frontierscience.py
```diff
@@ -634,6 +721,85 @@ async def _score_output(
...
+    def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance | None:
+        raw_problem = str(doc.get("problem") or "")
+        problem, substitutions = _MC_ANSWER_INSTRUCTION.subn("", raw_problem)
+        choices = tuple(str(choice).strip() for choice in doc.get("choices", ()))
+        if not problem.strip() or not choices:
+            return None
+        if substitutions != 1:
+            raise ValueError(f"expected one answer-instruction suffix at MC row {index}")
+        if len(choices) != 4 or any(not choice for choice in choices):
+            raise ValueError(f"expected four nonempty choices at MC row {index}")
```
gregorybchris, line 761: **Non-blocking (11).** Inconsistent handling of malformed rows.

A row with an empty `problem` or an empty `choices` list is silently skipped here, while every other defect below raises. So the one failure mode that changes the item count without a trace is the one that stays quiet. Raising in both cases would make the "validated on load" claim uniform.

Relatedly, `usable` is present on every row of the JSONL and never read. If a future regeneration marks a row unusable, it will still be scored.

tests/evals/tasks/test_frontierscience.py
```diff
@@ -833,3 +863,69 @@ def test_metric_subject_slice_stays_out_of_pairwise_scorer_fallback():
...
+        response = Response(
+            instance=instance,
+            request=LMRequest(request_type=RequestType.LOGLIKELIHOOD, prompt="q"),
+            outputs=[LMOutput(text="", logprobs=[{"logprob": score}]) for score in scores],
+        )
+        metrics = {metric.name: metric for metric in olympiad_mc.config.metrics}
+
+        assert metrics["accuracy"].compute([response]) == 1.0
+        assert metrics["accuracy_physics"].compute([response]) == 1.0
+        assert metrics["accuracy_biology"].compute([response]) == 0.0
```
gregorybchris, line 925: **Non-blocking (9).** This assertion locks in the empty-subject-slice sentinel.

`compute` returns `0.0` when no instance matches the slice, which is indistinguishable from a real zero. Biology is 9 of 83 rows, so a routine `limit=10` smoke run will often persist `accuracy_biology = 0.0`. `SubsetAccuracyMetric` uses `-1` for exactly this case.

This follows the existing module convention that `FrontierScienceMetric` documents at line 288, so I am not asking you to change it in this PR — but if the sentinel changes later, this test needs to move with it. A linked follow-up issue is enough.

pyproject.toml
```diff
@@ -59,6 +62,9 @@ vllm = [
     "vllm[runai]==0.19.1 ; platform_system != 'Darwin'",  # runai extras enable S3 model loading
     "opencv-python-headless==4.13.0.92 ; platform_system != 'Darwin'",  # Required to avoid libxcb import error
     "torch-c-dlpack-ext ; platform_system != 'Darwin'",  # Required for TVM tensor allocation in vLLM
+    # transformers 5.13 nests rope_parameters per layer type; vLLM 0.19.1 reads it flat,
+    # so Olmo-3 (mixed sliding/full attention) fails to load with a rope_theta KeyError.
+    "transformers<5.13",
```
gregorybchris: **Non-blocking (12).** Two repo-wide dependency pins riding along in a task PR.

You say so yourself in the description, and both look correct to me. `huggingface-hub>=1.10.0` is clearly justified by the module-level `from huggingface_hub.utils import silent_tqdm` in `src/olmo_eval/data/ruler_loader.py:16`. But `transformers<5.13` is a cap that will silently block future upgrades, and the reason lives only in this comment and the PR body.

Could we split them into their own PR so they are findable later? Failing that, please file an issue to remove the cap once vLLM handles the nested `rope_parameters`, and link it from the comment. I could not verify the vLLM 0.19.1 / transformers 5.13 interaction without a GPU — if you have the failing traceback, pasting it into that issue would help whoever revisits this.

Two more things here:

- The `olmo_core` extra pulls `ai2-olmo-core[torchao,transformers]==2.4.0` with no cap. If the rope breakage is repo-wide, does that path need the same constraint?
- Nit: line 67 has no `platform_system != 'Darwin'` marker, unlike every other entry in this extra, so `pip install olmo-eval[vllm]` on macOS now pulls transformers where it previously pulled nothing.

gregorybchris (commented):

Nice work — the curation effort behind the dataset is visible, the prompt construction is clean, the metric set is well chosen for 83 items, and the results table with its label-bias caveat is exactly the kind of honesty we want here. A few things to sort out before merging.

I found twelve issues and four nits. The first four affect persisted-result correctness, task identity, and failure visibility, so they should be fixed before merge. The next five are structure, reuse, and naming. The last three cover dataset provenance and the scope of the dependency pins. Everything below is also left inline where it has a useful line anchor; this comment carries the complete set so it can be handed to an agent in one piece.

Verification I ran against `a1c3c15d`: `ruff check` and `ruff format --check` clean, `ty check` clean apart from one pre-existing unrelated warning in `harness/sandbox/executor.py`, and `pytest tests/ --ignore=tests/integration` at 2288 passed / 8 skipped. CI is green.

### Blocking

1. **`src/olmo_eval/evals/tasks/frontierscience.py:728` — the vendored dataset is not pinned by anything in the config.**

   `data_source` serializes to `{'path': 'frontierscience_olympiad_mc.jsonl', 'subset': None, 'split': 'test', 'source_type': 'local', 'data_files': None, 'revision': None}`. I truncated the JSONL to 50 rows and re-hashed: the task loaded 50 instances and the hash stayed `6b7c356ab07d67e9`. A later revision of the file — adding back some of the 17 excluded questions, or another LaTeX normalization pass — produces a different eval that is indistinguishable from this one in storage. That breaks the property the result store rests on.

   `DataSource.revision` is already serialized into `to_dict()`, so the fix is small:

   ```python
   FRONTIERSCIENCE_MC_RELEASE = "2026-09-v1"  # bump on any edit to the JSONL

   data_source = DataSource(
       path=FRONTIERSCIENCE_VERIFIED_MC_FILE,
       revision=FRONTIERSCIENCE_MC_RELEASE,
   )
   ```

   Please add a test asserting that changing the release identifier changes the task hash.

2. **`src/olmo_eval/evals/tasks/frontierscience.py:359` — missing logprobs are scored as a correct answer.**

   `LogprobScorer` returns `-inf` when a provider returns no logprobs. With all four continuations at `-inf`, `scores.index(max(scores))` returns `0`, so every instance whose gold is A is credited correct. I reproduced this with a mock response carrying no logprobs: `compute_instance` returns `1.0` for a gold-A instance and `0.0` for a gold-C one, and `accuracy.compute([good_response, all_missing])` returns `1.0`. A completely broken run over this set lands at 0.25 — chance — with 21 of 83 items marked correct.

   That is an infrastructure failure turning into a plausible-looking score. This one is partly ours: `LogprobMCAccuracyMetric` in `src/olmo_eval/common/metrics/base.py:552` has the same argmax-over-all-`-inf` shape, so it is worth fixing in both places.

   ```python
   if all(score == float("-inf") for score in scores):
       return None
   ```

3. **`src/olmo_eval/evals/tasks/frontierscience.py:361` — `gold_probability` and `logprob_margin` emit NaN and ±inf, and one bad instance poisons the task-level number.**

   Reproduced with one good response and one whose gold logprob is missing: `logprob_margin.compute_instance` returns `-inf` and `logprob_margin.compute([good, that_one])` returns `-inf` for the whole task. With all four continuations missing, `gold_probability` returns `nan` per instance and `nan` in aggregate, because `math.exp(score - largest)` is `nan` when both are `-inf` and `sum(values) / len(values)` propagates it.

   NaN is also not valid JSON, so it does not round-trip cleanly to the predictions file or to Postgres. Please return `None` from `compute_instance` when the gold or the best distractor has no logprob, and have `compute` skip non-finite values.

4. **`src/olmo_eval/evals/tasks/frontierscience.py:371` — the accuracy denominator hides a mostly-failed run.**

   `compute` averages only the instances that produced a value. With one scored response and two that returned no outputs, `FrontierScienceMCMetric` accuracy is `1.0` where `LogprobMCAccuracyMetric` on the same input is `0.333`, because the existing metric divides by `len(responses)`. Two thirds of the task failing should not read as 100%.

   Either divide by `len(responses)` to match the existing metric, or record the scored count so the shortfall is visible in the result. The same `else 0.0` on line 377 means an all-failed run reports `0.0` rather than failing, which is the "all-failed run reported as Success with 0.0" pitfall.

### Non-blocking

5. **`src/olmo_eval/evals/tasks/frontierscience.py:870` — `frontierscience_olympiad:mc:olmo3base` is a no-op alias, and its name is model-specific.**

   `register_variant` is called with no overrides, and `get_task` never rewrites `config.name` for a variant. I built both specs and compared: identical `to_dict()`, and both hash to `6b7c356ab07d67e9`. Anyone who launched with `:olmo3base` — as all four commands in the description do — will find their results filed under `frontierscience_olympiad:mc`.

   A model-named variant is also what §7 of the review guide rules out: variants describe a format or metric modifier, and model-specific presets compose from variants. Could we drop the registration and launch the plain `:mc` spec?

6. **`src/olmo_eval/evals/tasks/frontierscience.py:736` — `data_source` is dead, and its `split=` is overwritten.**

   The `instances` property reads the file through `importlib.resources`, so `config.data_source` is never consulted and a `-o data_source=...` override would be silently ignored. `task._load_instances()` raises `FileNotFoundError: Dataset path not found: frontierscience_olympiad_mc.jsonl`. The `split="test"` argument on line 728 is also overwritten by `get_data_source()` from the class-level `split = Split.TEST`.

   If finding 1 is taken, `data_source` becomes the place the release identifier lives and this mostly resolves itself. Please drop the redundant `split=` either way.

7. **`src/olmo_eval/evals/tasks/frontierscience.py:337` — `LogprobMCAccuracyMetric` already does the accuracy case.**

   `src/olmo_eval/common/metrics/base.py:530` is the same argmax-over-`LogprobScorer` computation. The subject slices and the two continuous metrics are genuinely new, but the overall `accuracy` could reuse the existing metric so a future fix to the `-inf` handling lands once. Worth a comment explaining the divergence if you keep a separate class.

8. **`src/olmo_eval/evals/tasks/frontierscience.py:340` — `name` and `metric_kind` must agree but are independent fields.**

   `Metric.to_dict()` serializes only `type`, `name`, and `scorer`, so `FrontierScienceMCMetric(name="accuracy", metric_kind="gold_probability")` hashes identically to the real accuracy metric while computing something else. Deriving `name` from `metric_kind` and `subject` in `__post_init__` closes it. `metric_kind` is also a fixed set — `Literal["accuracy", "gold_probability", "logprob_margin"]` would let `ty` catch the typo that line 369's `ValueError` currently catches at runtime.

9. **`src/olmo_eval/evals/tasks/frontierscience.py:377` and `tests/evals/tasks/test_frontierscience.py:925` — an empty subject slice reports 0.0.**

   This follows the existing module convention, which `FrontierScienceMetric` documents at line 288, so it is not new. It bites harder here: biology is 9 of 83 rows, so a routine `limit=10` smoke run will often have zero biology instances and persist `accuracy_biology = 0.0`, which reads as a real zero. `SubsetAccuracyMetric` uses `-1` as its sentinel for exactly this. The new test asserting `metrics["accuracy_biology"].compute([response]) == 0.0` locks the behavior in, so it would need updating alongside.

   Happy for this to be a follow-up issue covering both metric families rather than PR scope. The issue should say: give `FrontierScienceMetric` and `FrontierScienceMCMetric` a sentinel for an empty subject slice so a `limit` run cannot persist a fake zero, and update the tests that assert `0.0`.

10. **`src/olmo_eval/evals/tasks/frontierscience_olympiad_mc.jsonl` — the vendored file has drifted from upstream in ways nothing records.**

    I loaded `openai/frontierscience` at the pinned revision `25ed67db7da8f4591484e764008ff585544f5a30` and diffed all 83 rows. All 83 `task_group_id`s exist upstream and are unique. Three things:

    - **3 problem stems differ from upstream.** Two of them — `bb0539ef` (`\bullet` → `\cdot`) and `328c6a47` (a stray `'` → a backtick, plus delimiter tightening) — have `display_override: null`, so there is no recorded reason for the edit. The third, `16318af5`, does record one. These stems are also now out of sync with the free-response `frontierscience_olympiad` task, which reads them from the Hub.
    - **~55 `reference_answer` values differ from the upstream `answer`**, almost all LaTeX normalization (`\Big[` → `\left[`, `\dfrac` → `\frac`, backtick stripping). One is a content edit: `dbe975d5` goes from `\vec{F}^\mathrm{}_Y = ... T_1(0) ...` to `|\vec F_Y| = ... T_1 ...`. The description says the MC variant is "generated using the correct answers to the original Olympiad questions", which undersells how much rewriting happened. Worth a sentence in the module docstring or the PR body.
    - **19 of 83 rows have `distractor_provenance` entries whose `answer` text no longer matches any rendered choice**, from the normalization pass running after provenance was recorded. The description says each row names the error mechanism for each distractor, but for those 19 rows you cannot join provenance to choice by answer text. Adding a `choice_index` to each entry would make it durable.

    Two things I checked that turned out fine, so you can ignore them: no row has duplicate or normalization-equivalent choices (my first pass flagged three, but they are SMILES strings where parentheses are meaningful), and gold is not identifiable by length — argmin- and argmax-length baselines both score 0.217, below the 0.25 floor.

11. **`src/olmo_eval/evals/tasks/frontierscience.py:753` — inconsistent handling of malformed rows.**

    A row with an empty `problem` or an empty `choices` list is silently skipped, while every other defect raises. So the one failure mode that changes the item count without a trace is the one that stays quiet. Raising in both cases would make the "validated on load" claim uniform. Relatedly, `usable` is present on every row and never read — if a future regeneration marks a row unusable, it will still be scored.

12. **`pyproject.toml:65` — two repo-wide dependency pins ride along in a task PR.**

    You say so yourself in the description. Both look correct to me, and `huggingface-hub>=1.10.0` is clearly justified by the module-level `from huggingface_hub.utils import silent_tqdm` in `src/olmo_eval/data/ruler_loader.py:16`. But `transformers<5.13` is a cap that will silently block future upgrades, and the reason lives only in a code comment and this PR body.

    Could we split them into their own PR so they are findable later? Failing that, please file an issue to remove the cap once vLLM handles the nested `rope_parameters`, and link it from the comment. I could not verify the vLLM 0.19.1 / transformers 5.13 interaction without a GPU — if you have the failing traceback, pasting it into that issue would help whoever revisits this. Also worth checking: the `olmo_core` extra pulls `ai2-olmo-core[torchao,transformers]==2.4.0` with no cap. If the rope breakage is repo-wide, does that path need the same constraint?

### Nits

- Nit: `pyproject.toml:67` — the new `transformers<5.13` in the `vllm` extra has no `platform_system != 'Darwin'` marker, unlike every other entry in that extra, so `pip install olmo-eval[vllm]` on macOS now pulls transformers where it previously pulled nothing.
- Nit: `src/olmo_eval/evals/tasks/frontierscience_olympiad_mc.jsonl` — `source_answer_flag` is `null` on all 83 rows. Dead column; drop it or document what a non-null value would mean.
- Nit: `src/olmo_eval/evals/tasks/frontierscience.py:773` — `metadata["index"]` is written but only used for the `id` fallback, which cannot trigger since all 83 `task_group_id`s are non-empty.
- Nit: `tests/evals/tasks/test_frontierscience.py` — no test covers the missing-logprob path. If findings 2 and 3 are fixed, a regression test asserting `compute_instance` returns `None` when logprobs are absent would be the natural companion.

### Before merge

Findings 1 through 4 should be fixed, each with a regression test — a hash test for the dataset pin, and metric tests for the `-inf`, NaN, and denominator cases. Finding 5 is a one-line deletion and worth doing in the same pass. For 9 and 12, a linked follow-up issue each is enough. Everything else can land now or later at your discretion.

Once those are in, you're good to merge whenever CI is green.

gregorybchris (approved):

donovanr:
@gregorybchris I’ve addressed the provenance links and the missing platform marker. Most of the remaining feedback I think is more about metric edge cases, shared infrastructure, or conventions already used elsewhere in the repo. Those are better handled in focused follow-up issues if you think they are worth opening. I’m keeping this PR scoped to the MC eval and don’t think those broader concerns should block it.

PR #360, opened 2026-09-08
Add DeepSearchQA task (two grading implementations)

fabrahman:
## Summary

- Adds the [DeepSearchQA](https://huggingface.co/datasets/google/deepsearchqa) benchmark (Google DeepMind, arXiv:2601.20975) — 900 hand-crafted, multi-step information-seeking questions across 17 domains, each requiring either a single answer or a complete set of items, scored by item-set precision/recall/F1.
- Registers **two** tasks sharing the same dataset/loading code (`DeepSearchQABase`), differing only in generation prompt and grading mechanic:
  - `deepsearchqa` — model ends with a discrete `FINAL ANSWER: ...` line, judged by index-matching two extracted item lists.
  - `deepsearchqa_official_judge` — mirrors the official paper prompt (Appendix A) verbatim: the judge grades the model's raw free-form response directly, no special output format required.
  - Both grade with an OpenAI model via `build_openai_judge_fn` (the official grader is Gemini 2.5 Flash, which olmo-eval's judge infra doesn't support), so absolute numbers from either task aren't directly comparable to the public Kaggle leaderboard — see `DEEPSEARCHQA.md` for the full writeup of what each deviates from and why.
- Fixes a data-handling bug: 4 of the 900 rows encode "no items satisfy every constraint" as the literal text `None` in the source CSV, which the HF CSV loader silently coerces to a null; these were being dropped as malformed instead of scored as an empty gold-answer set. Both tasks now handle this correctly (900/900 instances load, not 896/900).
- `:mini` variants (50-instance subset) registered for both tasks for quick iteration.
- `DEEPSEARCHQA.md` documents both tasks, the data quirk above, and the secrets each requires (`OPENAI_API_KEY` always; `S2_API_KEY`/`SERPER_API_KEY` additionally for `--harness dr_tulu`).

## Test plan

- [x] `uv run ruff check src/ tests/` — passes
- [x] `uv run ruff format --check src/ tests/` — passes
- [x] `uv run ty check src/ alembic/` — passes (no new diagnostics)
- [x] `uv run pytest tests/ --ignore=tests/integration` — 2247 passed
- [x] Verified against the real `google/deepsearchqa` dataset end-to-end: all 900 rows load (including the 4 previously-dropped empty-answer-set rows), `task inspect` prompts render correctly for both tasks
- [x] Ran a live 5-instance smoke test with `Qwen/Qwen2.5-7B-Instruct` end-to-end (generation + judge scoring with `OPENAI_API_KEY` set)
- [ ] Not yet run with search-tool keys for the intended `--harness dr_tulu` setup

### DIFF
--- a/src/olmo_eval/evals/tasks/deepsearchqa.py
+++ b/src/olmo_eval/evals/tasks/deepsearchqa.py
@@ -1,0 +1,392 @@
+"""DeepSearchQA: deep-research agent evaluation (Google DeepMind, arXiv 2601.20975).
+
+DeepSearchQA is a 900-problem benchmark of hand-crafted, multi-step
+information-seeking questions spanning 17 domains (Politics, Finance, Science,
+Health, History, Geography, Media, ...). Each problem is either a
+``Single Answer`` (one entity/value) or a ``Set Answer`` (an enumeration or
+composite answer with multiple required items); on HuggingFace both are
+stored as one comma-joined ``answer`` string
+(``google/deepsearchqa``, config ``deepsearchqa``, split ``eval``).
+
+Grading follows the paper's outcome-based, set-comparison methodology: an
+LLM judge decides, per item, whether a submitted answer is semantically
+equivalent to a ground-truth answer (and vice versa), and the task reports
+Precision (``|S ∩ G| / |S|``), Recall (``|S ∩ G| / |G|``), their harmonic
+mean F1 (the primary metric), and an all-or-nothing Exact Match rate. Unlike
+the paper/starter-notebook's official grader (a Gemini 2.5 Flash autorater
+whose exact prompt is not published), this task uses olmo-eval's OpenAI-based
+judge infrastructure with an independently written judge prompt, so absolute
+numbers are not directly comparable to the public leaderboard. Set
+``OLMO_EVAL_JUDGE=<model>[:<effort>]`` to change the judge model.
+
+The task asks the model to end its response with a ``FINAL ANSWER: ...`` line
+listing every item in its answer set, comma-separated; this line (rather than
+the full response) is what gets judged. This benchmark is about finding
+information via search rather than about knowledge already in the model's
+parameters, so it is intended to be run with search tools attached via the
+Harness abstraction (e.g. ``--harness dr_tulu``); running without tools mostly
+measures parametric recall on questions designed to require live search.
+
+Usage:
+    # Tool-augmented (intended way to run this benchmark)
+    olmo-eval run -m llama3.1-8b -t deepsearchqa --harness dr_tulu
+
+    # Baseline, no search tools (for reference only)
+    olmo-eval run -m llama3.1-8b -t deepsearchqa
+
+    # Quick iteration on a 50-instance subset
+    olmo-eval run -m llama3.1-8b -t deepsearchqa:mini --harness dr_tulu
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
+import re
+from collections.abc import Iterator, Sequence
+from dataclasses import dataclass, field
+from typing import Any
+
+from olmo_eval.common.metrics import Metric
+from olmo_eval.common.scorers.base import Scorer
+from olmo_eval.common.scorers.llm_judge import JudgeFn, build_openai_judge_fn
+from olmo_eval.common.types import (
+    Instance,
+    LMOutput,
+    LMRequest,
+    RequestType,
+    Response,
+    SamplingParams,
+)
+from olmo_eval.data import DataSource
+from olmo_eval.evals.tasks.common import Task, register, register_variant
+
+logger = logging.getLogger(__name__)
+
+DEEPSEARCHQA_REPO = "google/deepsearchqa"
+DEEPSEARCHQA_CONFIG = "deepsearchqa"
+DEEPSEARCHQA_SPLIT = "eval"
+
+DEEPSEARCHQA_DEFAULT_JUDGE_SPEC = "gpt-5.5:medium"
+DEEPSEARCHQA_JUDGE_ATTEMPTS = 3
+
+DEEPSEARCHQA_GENERATION_PROMPT = """\
+Answer the following question. It may require an exhaustive search across \
+multiple sources rather than a single lookup. If you have access to search \
+tools, use them to find and verify each candidate answer instead of \
+answering from memory alone.
+
+The question may have exactly one correct answer, or it may require a \
+complete set of items that together satisfy every constraint it states \
+(watch for words like "all", "each", or "list"). Do not stop at the first \
+plausible answer if the question implies there may be more than one.
+
+Finish your response with a line of exactly this form:
+FINAL ANSWER: <item 1>, <item 2>, ...
+
+List every item belonging to your answer set on that line, separated by \
+commas, and put nothing else on it. If there is a single correct answer, \
+list just that one item.
+
+Question: {question}"""
+
+# Tolerates markdown emphasis and a missing/extra colon around the marker.
+_FINAL_ANSWER_MARKER = re.compile(r"\**\s*FINAL\s+ANSWER\s*\**\s*:?\s*", re.IGNORECASE)
+
+
+def extract_final_answer(text: str) -> str:
+    """Return the text after the last ``FINAL ANSWER`` marker.
+
+    Falls back to the full response when the marker is absent, so the judge
+    still sees the model's attempt.
+    """
+    matches = list(_FINAL_ANSWER_MARKER.finditer(text))
+    if not matches:
+        return text.strip()
+    return text[matches[-1].end() :].strip() or text.strip()
+
+
+def split_answer_set(text: str) -> list[str]:
+    """Split a comma-delimited answer string into non-empty, stripped items."""
+    return [item.strip() for item in text.split(",") if item.strip()]
+
+
+def _numbered_list(items: Sequence[str]) -> str:
+    return "\n".join(f"{index}. {item}" for index, item in enumerate(items))
+
+
+DEEPSEARCHQA_JUDGE_PROMPT = """\
+You are grading a submitted answer against a ground-truth answer for the \
+question below. The answer may be a single item or a set of items. Two \
+items are equivalent if they refer to the same real-world entity, value, or \
+fact, even when worded differently (abbreviations, alternate spellings, \
+reordering, or trivial formatting differences do not matter). Judge \
+semantic equivalence, not surface string similarity.
+
+Question: {question}
+
+Ground-truth answer set ({num_gold} item(s)):
+{gold_list}
+
+Submitted answer set ({num_pred} item(s)):
+{pred_list}
+
+For every ground-truth item, decide whether it is semantically covered by \
+at least one submitted item. Separately, for every submitted item, decide \
+whether it matches at least one ground-truth item.
+
+Respond with only a JSON object of exactly this form, and nothing else:
+{{"matched_gold_indices": [...], "matched_submitted_indices": [...]}}
+Both lists hold the 0-based indices (into the lists above) of the items \
+that were matched."""
+
+_JSON_OBJECT = re.compile(r"\{.*\}", re.DOTALL)
+
+
+def build_deepsearchqa_judge_fn() -> JudgeFn:
+    """Build the task's judge from ``$OLMO_EVAL_JUDGE`` or its own default."""
+    import os
+
+    spec = os.getenv("OLMO_EVAL_JUDGE", DEEPSEARCHQA_DEFAULT_JUDGE_SPEC)
+    model, separator, effort = spec.partition(":")
+    return build_openai_judge_fn(
+        model=model,
+        temperature=0.0,
+        max_tokens=1024,
+        scorer_name="DeepSearchQA",
+        reasoning_effort=(effort if separator else None),
+    )
+
+
+def build_deepsearchqa_judge_prompt(
+    question: str, gold_items: Sequence[str], pred_items: Sequence[str]
+) -> str:
+    """Format the item-matching judge prompt for one instance."""
+    return DEEPSEARCHQA_JUDGE_PROMPT.format(
+        question=question,
+        num_gold=len(gold_items),
+        gold_list=_numbered_list(gold_items),
+        num_pred=len(pred_items),
+        pred_list=_numbered_list(pred_items),
+    )
+
+
+def parse_deepsearchqa_judge_response(
+    raw: str, num_gold: int, num_pred: int
+) -> tuple[set[int], set[int]] | None:
+    """Parse the judge's matched-index JSON, or None if it is unparseable."""
+    match = _JSON_OBJECT.search(raw)
+    if match is None:
+        return None
+    try:
+        data = json.loads(match.group(0))
+    except json.JSONDecodeError:
+        return None
+    if not isinstance(data, dict):
+        return None
+
+    gold_raw, pred_raw = data.get("matched_gold_indices"), data.get("matched_submitted_indices")
+    if not isinstance(gold_raw, list) or not isinstance(pred_raw, list):
+        return None
+
+    try:
+        matched_gold = {int(i) for i in gold_raw if 0 <= int(i) < num_gold}
+        matched_pred = {int(i) for i in pred_raw if 0 <= int(i) < num_pred}
+    except (TypeError, ValueError):
+        return None
+    return matched_gold, matched_pred
+
+
+def compute_deepsearchqa_scores(
+    num_gold: int, num_pred: int, num_matched_gold: int, num_matched_pred: int
+) -> dict[str, float]:
+    """Compute precision/recall/F1/exact-match from item-level match counts."""
+    recall = num_matched_gold / num_gold if num_gold else 1.0
+    precision = num_matched_pred / num_pred if num_pred else 0.0
+    f1 = 0.0 if precision + recall == 0 else 2 * precision * recall / (precision + recall)
+    exact_match = 1.0 if (num_pred > 0 and num_matched_pred == num_pred and recall == 1.0) else 0.0
+    return {
+        "deepsearchqa_precision": precision,
+        "deepsearchqa_recall": recall,
+        "deepsearchqa_f1": f1,
+        "deepsearchqa_exact_match": exact_match,
+    }
+
+
+@dataclass(frozen=True)
+class DeepSearchQAScorer(Scorer):
+    """Placeholder scorer; DeepSearchQA scores are computed in ``score_responses``."""
+
+    name: str = "deepsearchqa_f1"
+    score_key: str = "deepsearchqa_f1"
+
+    def score(self, instance: Instance, output: LMOutput) -> float:
+        return (output.metadata or {}).get(self.score_key, 0.0)
+
+
+@dataclass(frozen=True)
+class DeepSearchQAMetric(Metric):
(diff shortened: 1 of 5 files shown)


src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+that were matched."""
+
+_JSON_OBJECT = re.compile(r"\{.*\}", re.DOTALL)
+
+
+def build_deepsearchqa_judge_fn() -> JudgeFn:
+    """Build the task's judge from ``$OLMO_EVAL_JUDGE`` or its own default."""
+    import os
+
+    spec = os.getenv("OLMO_EVAL_JUDGE", DEEPSEARCHQA_DEFAULT_JUDGE_SPEC)
```
gregorybchris, line 159: > Agent review per `REVIEW_GUIDE.md`.

**Blocking.** This judge spec never reaches `TaskConfig`, so `to_dict()` (and the task hash) is identical with and without `OLMO_EVAL_JUDGE` set; I reproduced that with `gpt-4.1-mini` vs the default. `TaskConfig` already has `judge_model` / `judge_reasoning_effort` / `judge_max_tokens`, and `frontierscience.py` resolves the env override in `__init__` via `dataclasses.replace` so it is hashed. Please do the same on `DeepSearchQABase` and add a test that changing the judge changes `to_dict()`.

src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+            # The source CSV encodes "no items satisfy every constraint" as the literal
+            # text "None" on Set Answer rows; the HF CSV loader coerces that string to a
+            # null, so it arrives here indistinguishable from a missing field. A missing
+            # answer only makes sense as an intentional empty answer set for Set Answer
+            # rows; treat anything else as a malformed row.
+            if answer_type != "Set Answer":
+                return None
+            gold_items: list[str] = []
+        else:
+            gold_items = split_answer_set(answer)
```
gregorybchris, line 321: > Agent review per `REVIEW_GUIDE.md`.

**Blocking.** On the real dataset 22 of 316 `Single Answer` rows contain commas ("Lincoln, NE", "Hasbro, Inc.", "July 3, 2001"), so this produces 2-3 gold items and caps recall for a correct answer at 0.5 or 0.33. Set rows use semicolons and parenthetical commas too ("Rebecca West (1969, 1970), Antonia Fraser (1970, 1971), ..." becomes 28 items for a 14-item set). The paper scores single-answer questions as binary exact match, so a `Single Answer` row should be a one-item gold set regardless of punctuation.

src/olmo_eval/evals/tasks/deepsearchqa_official_judge.py
```diff
@@ -0,0 +1,362 @@
...
+        pred_text = pred_text.strip()
+
+        base_metadata: dict[str, Any] = {"deepsearchqa_official_gold_items": gold_items}
+
+        if not pred_text:
+            # An empty response can't contain any gold item or excessive claim.
+            scores = compute_deepsearchqa_official_scores(len(gold_items), 0, 0)
+            return scores, base_metadata, 0
+
+        answer_text = ", ".join(gold_items) if gold_items else "None"
```
gregorybchris, line 338: > Agent review per `REVIEW_GUIDE.md`.

**Blocking (same root cause).** Rejoining the comma-split items changes the text the "verbatim" Appendix A prompt sees (semicolons and newlines in the source become commas), and `len(gold_items)` becomes a wrong recall denominator on the rows noted in the summary. The official grader takes the raw answer string; please pass `answer` through unchanged and derive `num_gold` from the judge's `Correctness Details` length instead of a client-side split.

src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+        parsed = None
+        for attempt in range(DEEPSEARCHQA_JUDGE_ATTEMPTS):
+            raw = await judge_fn(prompt)
+            parsed = parse_deepsearchqa_judge_response(raw, len(gold_items), len(pred_items))
+            if parsed is not None:
+                break
+            if attempt < DEEPSEARCHQA_JUDGE_ATTEMPTS - 1:
+                await asyncio.sleep(2**attempt)
+
+        matched_gold, matched_pred = parsed if parsed is not None else (set(), set())
```
gregorybchris, line 423: > Agent review per `REVIEW_GUIDE.md`.

**Blocking.** After three unparseable replies this scores 0.0 on all four metrics with no per-instance record: no flag in `output.metadata`, no raw judge text. In saved predictions that is indistinguishable from a genuine zero. Please write an explicit parse-failed flag and the last raw response into `output.metadata` (and the raw response on success); `build_predictions` already forwards a `judge_result` key.

src/olmo_eval/evals/tasks/deepsearchqa_official_judge.py
```diff
@@ -0,0 +1,362 @@
...
+        parsed = None
+        for attempt in range(DEEPSEARCHQA_JUDGE_ATTEMPTS):
+            raw = await judge_fn(prompt)
+            parsed = parse_deepsearchqa_official_judge_response(raw, len(gold_items))
+            if parsed is not None:
+                break
+            if attempt < DEEPSEARCHQA_JUDGE_ATTEMPTS - 1:
+                await asyncio.sleep(2**attempt)
+
+        num_matched, num_excessive = parsed if parsed is not None else (0, 0)
```
gregorybchris, line 352: > Agent review per `REVIEW_GUIDE.md`.

**Blocking (mirror of the same finding in `deepsearchqa.py`).** Exhausted retries score 0.0 with no per-instance parse-failure flag or raw judge response in `output.metadata`.

src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+        "deepsearchqa_f1": f1,
+        "deepsearchqa_exact_match": exact_match,
+    }
+
+
+@dataclass(frozen=True)
+class DeepSearchQAScorer(Scorer):
+    """Placeholder scorer; DeepSearchQA scores are computed in ``score_responses``."""
+
+    name: str = "deepsearchqa_f1"
```
gregorybchris, line 243: > Agent review per `REVIEW_GUIDE.md`.

All four scorer instances share this `name` and differ only in `score_key`; `task._get_scorers()` returns one entry, and `score()` reads the unprefixed key while `score_responses` writes `score:...`, so it returns 0.0 whenever called. `analysis/pairwise.py` matches `metric.scorer().name` against the stored `metric:scorer` key, so precision/recall/exact-match will not resolve their display profile. Deriving `name` from `score_key` fixes both.

src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+        return extract_final_answer(output.text)
+
+    async def score_responses(
+        self,
+        responses: Sequence[Response],
+        context: Any = None,
+    ) -> Sequence[Response]:
+        """Judge each response's answer set against the gold answer set."""
+        self._extract_answers(responses)
+        judge_fn = build_deepsearchqa_judge_fn()
```
gregorybchris, line 363: > Agent review per `REVIEW_GUIDE.md`.

The runner calls `score_responses` with a single response (`runners/asynq/results.py:224`), so this builds a fresh OpenAI client and a fresh adaptive request-shape cache per instance; every instance re-learns that `gpt-5.5` rejects `max_tokens` and `temperature`. `frontierscience.py` caches the judge on the task for this reason. Also, `max_tokens=1024` at `medium` effort is a shared reasoning+output budget on chat completions; how many judge calls needed a retry in the smoke test?

src/olmo_eval/evals/tasks/deepsearchqa_official_judge.py
```diff
@@ -0,0 +1,362 @@
...
+
+
+F1_METRIC = _metric("deepsearchqa_official_f1")
+PRECISION_METRIC = _metric("deepsearchqa_official_precision")
+RECALL_METRIC = _metric("deepsearchqa_official_recall")
+EXACT_MATCH_METRIC = _metric("deepsearchqa_official_exact_match")
+DEEPSEARCHQA_OFFICIAL_METRICS = (F1_METRIC, PRECISION_METRIC, RECALL_METRIC, EXACT_MATCH_METRIC)
+
+
+@register("deepsearchqa_official_judge")
```
gregorybchris, line 267: > Agent review per `REVIEW_GUIDE.md`.

Question: the implementation that matches the paper's grader lives under the qualified name, while the independently designed mechanic owns `deepsearchqa`. The guide's parity rule gives the plain name to the reference-matching port and names the divergence. Could we swap (e.g. `deepsearchqa` = official prompt, `deepsearchqa_final_answer` = index matching) before the first stored results exist?

src/olmo_eval/evals/tasks/deepsearchqa.py
```diff
@@ -0,0 +1,435 @@
...
+For every ground-truth item, decide whether it is semantically covered by \
+at least one submitted item. Separately, for every submitted item, decide \
+whether it matches at least one ground-truth item.
+
+Respond with only a JSON object of exactly this form, and nothing else:
+{{"matched_gold_indices": [...], "matched_submitted_indices": [...]}}
+Both lists hold the 0-based indices (into the lists above) of the items \
+that were matched."""
+
+_JSON_OBJECT = re.compile(r"\{.*\}", re.DOTALL)
```
gregorybchris, line 152: > Agent review per `REVIEW_GUIDE.md`.

Nit: unused after the `raw_decode` change (same in `deepsearchqa_official_judge.py:131`).

gregorybchris (commented):

Looks good overall! The data-loading fix for the four `None` rows is right (I confirmed 900/900 instances load with unique ids), the two-implementation split is clearly documented, and the tests cover the scoring arithmetic well. A few things we should address before merging.

I found three issues that affect task identity or persisted-result correctness, so they should be fixed before merge. The rest are non-blocking structure and diagnostics questions, plus nits. Verification: ruff, ruff format, and the PR's 55 tests pass locally; `ty` reports one diagnostic that is also present on `main` and is unrelated to this PR.

### Blocking

1. `src/olmo_eval/evals/tasks/deepsearchqa.py:159` (and `deepsearchqa_official_judge.py:138`)
   The judge model and reasoning effort are read from `OLMO_EVAL_JUDGE` at scoring time and never reach `TaskConfig`. I reproduced this: `get_task("deepsearchqa").config.to_dict()` is byte-identical with `OLMO_EVAL_JUDGE` unset and with `OLMO_EVAL_JUDGE=gpt-4.1-mini`, and no `judge_*` key appears in the serialized config. Two runs judged by different models therefore collide on the same task hash and artifact name, and neither the default `gpt-5.5:medium` nor an override is recoverable from a stored result. `TaskConfig` already has serialized `judge_model`, `judge_reasoning_effort`, and `judge_max_tokens` fields for exactly this, and `frontierscience.py` shows the pattern: resolve the env override in `__init__` via `dataclasses.replace(config, ...)`, then build the judge from `self.config.judge_*` in scoring. `researchqa.py` has the same defect as this PR; that is a pre-existing fault, not a reason to repeat it. Please move the judge spec into the config on `DeepSearchQABase` (both tasks share it) and add a test asserting that changing the judge model changes `to_dict()`.
   ```python
   class DeepSearchQABase(Task):
       judge_model = "gpt-5.5"
       judge_reasoning_effort = "medium"
       judge_max_tokens = 1024

       def __init__(self, config: TaskConfig) -> None:
           if spec := os.getenv("OLMO_EVAL_JUDGE"):
               model, _, effort = spec.partition(":")
               config = replace(config, judge_model=model, judge_reasoning_effort=effort or None)
           super().__init__(config)
   ```

2. `src/olmo_eval/evals/tasks/deepsearchqa.py:321` (shared `process_doc`; consumed at `deepsearchqa_official_judge.py:338`)
   Splitting the gold `answer` on commas produces wrong gold sets on the real data. I checked the HF dataset: 22 of the 316 `Single Answer` rows contain commas ("Lincoln, NE", "Hasbro, Inc.", "July 3, 2001", "209,550,294"), so each becomes a two- or three-item gold set and a fully correct one-item answer can score at most recall 0.5 (or 0.33). Set rows are worse: "Wild Water Buffalo, 1965; Komodo Dragon, 1986; ..." is semicolon-delimited, and "Rebecca West (1969, 1970), Antonia Fraser (1970, 1971), ..." splits into 28 items for a 14-item set, halving the attainable recall. Since `len(gold_items)` is the recall denominator in both tasks, these persisted per-instance and aggregate numbers are wrong, not just noisy. The paper scores single-answer questions as binary exact match, so at minimum a `Single Answer` row must be a one-item gold set regardless of punctuation. For the official-judge task, the Appendix A prompt is designed to take the raw answer string, and rejoining `gold_items` with ", " also changes the text the "verbatim" prompt sees (newlines and semicolons become commas). Please pass `answer` through unchanged there and derive `num_gold` from the judge's `Correctness Details` length rather than a client-side split. For the index-matching task, could we either split on the delimiter the row actually uses or document that set rows with nested commas are mis-scored and list how many there are?

3. `src/olmo_eval/evals/tasks/deepsearchqa.py:423` (and `deepsearchqa_official_judge.py:352`)
   When the judge returns three unparseable replies, the instance is scored 0.0 across all four metrics and the only record is an aggregate log line. Nothing in `output.metadata` marks the instance as a judge failure, and the raw judge text is not stored, so in saved predictions a parse failure is indistinguishable from a genuine zero. Note that `build_openai_judge_fn` returns `""` when a reasoning model spends its whole `max_completion_tokens` budget before emitting content, so with a 1024-token budget at `medium` effort this path may fire often (see non-blocking 3). Please write an explicit flag (for example `deepsearchqa_judge_parse_failed: True`) and the last raw judge response into `output.metadata` on this path, and record the raw response on success too; `build_predictions` already forwards a `judge_result` key so this survives to disk. A regression test that inspects the persisted metadata after exhausted retries would lock it in.

### Non-blocking

1. `src/olmo_eval/evals/tasks/deepsearchqa_official_judge.py:267`
   The implementation that matches the paper's grader is registered under the qualified name, and the independently designed mechanic owns `deepsearchqa`. The guide's parity rule is that the reference-matching port gets the plain name and divergences are named. Could we swap them, so `deepsearchqa` is the official-prompt grader and the FINAL ANSWER variant becomes `deepsearchqa_final_answer` (or similar)? This is much cheaper to change before the first stored results exist than after.

2. `src/olmo_eval/evals/tasks/deepsearchqa.py:243` (and `deepsearchqa_official_judge.py:225`)
   All four `DeepSearchQAScorer` instances share `name="deepsearchqa_f1"` and differ only in `score_key`. I confirmed `task._get_scorers()` returns a single entry, and `Scorer.score()` reads `output.metadata["deepsearchqa_f1"]` while `score_responses` writes `score:deepsearchqa_f1`, so the scorer returns 0.0 whenever anything invokes it. Nothing on the scoring path does today, but `analysis/pairwise.py` resolves a metric profile by matching `metric.scorer().name` against the stored `metric:scorer` key, so precision, recall, and exact-match will fail to resolve their display format. Deriving `name` from `score_key` (or passing `name=name` in `_metric`) fixes both.

3. `src/olmo_eval/evals/tasks/deepsearchqa.py:164` and `:363` (mirrored in the official-judge module)
   Two related judge-construction concerns. First, the runner calls `score_responses` with one response at a time (`runners/asynq/results.py:224`), so building the judge inside `score_responses` creates a fresh client and a fresh copy of the adaptive request-shape cache per instance; each of 900 instances rediscovers that `gpt-5.5` rejects `max_tokens` and `temperature`. `frontierscience.py` caches the judge on the task for this reason. Second, `max_tokens=1024` with `reasoning_effort=medium` is a shared budget for reasoning and output on chat completions; frontierscience uses 8192 and notes truncation as its parse-failure cause. From the five-instance smoke test, how many judge calls needed a retry? If any did, please raise the budget.

4. `src/olmo_eval/evals/tasks/deepsearchqa.py:393`
   Only `outputs[0]` is judged; other samples get no `score:` metadata and `output_score_aggregation` is ignored. That is fine for a temperature-0 task, but please say so in a comment or raise on `len(outputs) > 1` so a future `:n8` variant does not silently score one sample.

5. `src/olmo_eval/evals/tasks/deepsearchqa.py:114`
   When the marker is absent, the whole response is split on commas and sent to the judge as the "submitted set", which will usually produce dozens of items and a near-zero precision. That is a reasonable penalty, but nothing records that it happened. Could we write a `deepsearchqa_format_correct` boolean to `output.metadata` so format failures can be separated from wrong answers after the run?

6. Both modules duplicate the placeholder scorer and metric classes, the judge builder, the JSON-object scan loop, the `score_responses` loop, and the retry loop nearly verbatim. Since the PR already introduces `DeepSearchQABase`, could the shared pieces move there so a future fix lands once? The JSON scan helper also looks generally useful next to `common/scorers/citation.py::extract_json_from_response`, which it improves on.

### Nits

- Nit: `_JSON_OBJECT` at `deepsearchqa.py:152` and `deepsearchqa_official_judge.py:131` is unused after the `raw_decode` change.
- Nit: `DEEPSEARCHQA.md:38` ("No other files were modified") is PR-description content that will be false after the next commit; the "Files" section as a whole reads as a changelog rather than documentation. This is also the first task doc under `src/olmo_eval/evals/tasks/`; consider whether it belongs beside the module or in the README's task docs.
- Nit: `extract_final_answer` does not strip a leading `<think>...</think>` block, so a marker inside the trace is matched (I checked: `"<think>FINAL ANSWER: x</think> ... y"` returns `"x</think> ... y"`). `researchqa.py` strips the block first.

### Before merge

Please address the three blocking items (judge config in the hash, gold-answer splitting on real rows, and per-instance judge-failure metadata), each with the regression test described. If the naming swap in non-blocking 1 is declined, that is fine, but please say so in the thread so the decision is on record. After that you may merge once CI is green and the branch is rebased on `main`.

wu-ming233 (changes requested):
I went through the automated reviews, and I agree with the major blocking items:
- the comma-based splitting does not work for the set-based answers, and instead should be giving the raw string to LLM to judge
- judge information should be included in TaskConfig for hashing purposes
- the failed judgements should explicitly warn

The other automated reviews are non-blocking to me.

I implemented these fixes [here](https://github.com/allenai/olmo-eval/commit/45005ce1ba5b41ebd3e2fb207d2c2ca08106af76), for your reference.

Additionally, I changed our own judge to also use the official judge prompt (with a small tweak) to try to follow original eval as much as possible, so both deepsearchqa.py and deepsearchqa_official_judge.py use official prompt, except we extract FINAL ANSWER: in deepsearchqa.py.

fabrahman:
> I went through the automated reviews, and I agree with the major blocking items:
> 
> * the comma-based splitting does not work for the set-based answers, and instead should be giving the raw string to LLM to judge
> * judge information should be included in TaskConfig for hashing purposes
> * the failed judgements should explicitly warn
> 
> The other automated reviews are non-blocking to me.
> 
> I implemented these fixes [here](https://github.com/allenai/olmo-eval/commit/45005ce1ba5b41ebd3e2fb207d2c2ca08106af76), for your reference.
> 
> Additionally, I changed our own judge to also use the official judge prompt (with a small tweak) to try to follow original eval as much as possible, so both deepsearchqa.py and deepsearchqa_official_judge.py use official prompt, except we extract FINAL ANSWER: in deepsearchqa.py.

@wu-ming233 The fix looks great and clean. Thanks!
Merging the fix.

Issue #348, opened 2026-09-02
Reasoning-regime ports don't strip <think> content before extraction (oe-eval applies model-level r1_style)


abhishekraok:
oe-eval's Olmo 3 paper runs for Think models set `process_output: "r1_style"` on the **model** config (`oe_eval/models/output_processors.py`): everything through `</think>` (plus `<answer>` tags) is removed from the continuation before any task scores it. olmo-eval has no model-level equivalent, and the ported reasoning-regime tasks — `gpqa:cot`, `popqa`, `omega_500` — run their extractors on the raw output.

Measured on the paper's `mmlu:cot::hamish_zs_reasoning_deepseek` run for Olmo-3-7B-Think (Beaker result dataset `01KAD4QTVWCR9RZSS48YZ52040`, 14,042 instances): 13,999 outputs contain `</think>`; scoring the raw output disagrees with oe-eval on 132 verdicts (99.06% agreement), while stripping first agrees on 14,042/14,042. `olmo_eval.evals.extract.extract_think_answer` is regex-for-regex identical to oe-eval's processor and is a no-op on outputs without `</think>`, so applying it inside each extractor is safe for Instruct models.

#339 does this task-locally for `mmlu:cot`. Remaining:
- apply the same in `gpqa:cot`, `popqa`, `omega_500` extractors (each was GPU-certified on Olmo-3-7B-Instruct only, so the gap went unobserved), or
- add a model-level output processor (`process_output=r1_style`) so it need not be repeated per task — closer to oe-eval's design.

Any published Think-model number from these tasks should be treated as unverified until one of these lands.

abhishekraok:
Claude Code · https://claude.ai/code/session_01LMFjgUqHjQA3aCeB4QmEHZ

#361 landed the mechanism: `TaskConfig.strip_thinking`, applied by the runner before scoring (oe-eval's `r1_style`, model-level). It defaults to `False` and sets nothing on any task, so this issue stays open until the reasoning-regime tasks turn it on — oe-eval applied `r1_style` to every Olmo 3 PAPER post-training run, Instruct models included:

- `gpqa:cot`, `mmlu_*:cot`, `popqa`, `omega_500`, `ifeval`, `ifeval_ood`, `ifeval_mt*` → `strip_thinking = True`

`mmlu:cot` (#339) and `ifeval` (#297) already strip inside their extractors; that becomes redundant (idempotent) once the flag is on.

PR #363, opened 2026-09-09
strip_thinking: keep the post-trace text verbatim, as oe-eval does

abhishekraok:
Follow-up to #361, toward #348. The runner's strip did `.lstrip()` after removing the trace; oe-eval's `r1_style` processor keeps the post-`</think>` text byte-for-byte (and removes `<answer>` tags). The trim is not cosmetic for IFEval: the loose variants drop the response's *first line* — with the leading newline kept that line is empty, with it trimmed it's real content — and `nth_paragraph_first_word` indexes paragraphs before dropping empty ones.

Measured on the paper's Olmo-3-7B-Think `ifeval` run (Beaker results `01KAD4RGRYERJZY4SHEPXYPBX4`, 541 prompts), scoring the raw outputs through the runner transform then `IFEvalScorer`:

| transform | verdict agreement with oe-eval |
|---|---|
| oe-eval's `r1_style` (this PR) | 539/541 |
| #361 as merged (`.lstrip()`) | 528/541 |

The 12 extra disagreements all flip a **loose** verdict to pass (`keywords:forbidden_words` ×4, `nth_paragraph_first_word` ×3, `repeat_prompt`, `two_responses`, `no_comma`, `number_words`, `keywords:frequency`), so the merged behavior inflates Think-model IFEval loose scores by ~2 points relative to oe-eval. Instruct-model runs (no trace) are unaffected.

Change: `strip_thinking_traces` now reuses `extract_think_answer`, the existing regex-for-regex mirror of `r1_style`. Tests adapted for the preserved whitespace; two added (verbatim text, `<answer>` tags).


gregorybchris:
I realized maybe I should have run an agent review on your previous strip_thinking PR, so I did that after you merged and Claude had a pretty long writeup. The thing about BPB sounded plausible as something worth fixing. Same with the task hash. 

I think you could consider these out of PR scope or pull fixes into this PR or tell me that these are not real concerns.. but wanted to add them here in case helpful.

```
Review of PR #361 (abhishekr/strip-thinking, +263/−4 across 6 source and 4 test files). Lint, type checks, and the touched tests all pass. Two findings are worth blocking on. The rest are worth fixing or at least discussing before merge.

Blocking

1. Every task hash changes on merge, even with the flag off. TaskConfig.to_dict() in src/olmo_eval/evals/tasks/common/base.py:296 unconditionally emits "strip_thinking": False. The task hash is computed over the full serialized dict, so pre- and post-merge runs of the same task get different hashes. That breaks pairwise dedup and best-suite intersection in analysis/pairwise.py, changes prediction filename suffixes, and invalidates saved results-viewer filters. The convention elsewhere in that function is to emit optional fields only when set. Fix: emit the key only when strip_thinking is true.
2. The strip regex is quadratic and runs on the event loop. re.sub(r"(?s).*</think>", "", text) at base.py:620 backtracks fully at each position after the last tag. Measured cost per output:

| Answer length | re.sub | rpartition |
|---------------|--------|------------|
| 5k chars      | 5 ms   | ~µs        |
| 50k chars     | 371 ms | ~µs        |
| 100k chars    | ~1.8 s | ~µs        |
   This runs synchronously inside score_and_store under the scoring semaphore, so a long-form task with many large answers stalls concurrent scoring and worker health checks. text.rpartition("</think>")[2].lstrip() is semantically identical. The same regex already exists in evals/extract/reasoning.py::extract_think_answer, so a shared helper would fix both.

Should fix

3. BPB metrics disagree with the scorer. BitsPerByteScorer now normalizes by original_text bytes, but BPBMetricByteAvg.compute and _select_gold_output in common/metrics/base.py still count the stripped text. Corpus BPB weights no longer match per-instance values, and an output that is entirely a trace becomes empty text and is silently dropped. Latent today since BPB tasks are loglikelihood-only, but the PR's own scorer test makes the combination nominally supported.
4. The vLLM reasoning-parser path is not covered. With --reasoning-parser set, the provider reads only message.content and drops reasoning_content, so no tag remains and the strip no-ops. Logprobs then span the full generation while bytes count only the answer, which is exactly the skew this PR fixes for the tag case. Pre-existing gap, but the tag-keyed design cannot cover it. Prepending reasoning_content into original_text at the provider, or a first-class reasoning field on LMOutput, would.
5. Three normalizations for logits_per_char. The prediction row in runners/io/builders.py:69 divides by original chars, the live per-char MC metric divides by stripped chars, and the offline Postgres recompute divides by the row's num_chars. The stored field no longer equals sum_logits / num_chars. Low realism, but pick one convention.
6. The strip has no single choke point, and the tested path is dead in production. It is bolted on before two score_responses call sites. The preparation.py call only runs for errored trackers, and the one runner-level test exercises only that path. A regression in the live results.py:224 path would pass CI. A Task.score() template method that strips then awaits score_responses would centralize it, with a test through process_results.

Design and cleanup

7. Last-tag semantics are lossy. A literal </think> in a code answer strips the real answer. A generation truncated right after </think> scores as empty, while an unterminated <think> scores as the full raw trace. No finish_reason is stored to tell these apart. This matches oe-eval's behavior and the PR's tests, so it is a defensible choice, but sage.py already has a more careful strip_think that could be shared.
8. Only outputs[*].text is stripped. Assistant turns in trajectory and request_trace keep the trace. No current scorer reads them, but harness scaffolds fill turn content with raw text, so a future agentic judge would silently see the trace. Extend the strip or document the limitation.
9. Small duplication and dead code. num_chars_all is emitted only when original_text exists, unlike its sibling num_tokens_all, so downstream readers will KeyError on non-thinking runs. Two separate if "original_text" in meta blocks guard the same key. The original_text or text expression and the BPB formula are duplicated between builders and the scorer. The if output.metadata is None guard is dead since metadata defaults to a dict. A full_text property on LMOutput would give one definition.
10. TASK_CONFIG_FIELDS is hand-maintained and already stale. The PR adds one entry to a frozenset in cli/utils.py:130 that omits eight real TaskConfig fields, so -o judge_model=x is rejected as an invalid override. Deriving it from dataclasses.fields(TaskConfig) plus priority is a drop-in, and the same function already does this for SamplingParams.
```

abhishekraok:

Thanks — went through all ten against the code. Verdicts and where each landed:

**Pulled into this PR**
- (1) Task hashes: confirmed — `compute_task_hash` is sha256 over the full `to_dict()`, and the unconditional `strip_thinking: False` changed every hash when #361 merged. Now emitted only when set, so hashes of runs that never used the option return to their pre-#361 values (test added).
- (2) Quadratic regex: confirmed and worse than measured here — 1.2 s at 50k chars, 3.9 s at 100k. `extract_think_answer` (which this PR routes the strip through) now uses `rpartition`; tests pin equivalence with the reference regexes on the edge cases, plus a timing test.
- (6) Test coverage: confirmed — `finalize_task` only runs for errored trackers. Added a test through `process_results` → `score_and_store`.
- (7)(8)(9-guard): documented in the docstring (last-tag semantics, only `outputs[*].text` is touched); dead `metadata is None` guard removed.

**Split out**
- (3)(5)(9-rest) → #365: `original_text` vs `text` normalization in the BPB metric, `acc_per_char`, and the prediction row; `num_chars_all` emitted conditionally. All latent (loglikelihood tasks carry no trace). I couldn't find the "Postgres recompute" of `logits_per_char` — the only `per_char` sites are `builders.py` and `MCAccuracyPerChar`.
- (4) → #366: `vllm_server` drops `reasoning_content`, so the parser path can't be stripped or normalized — also a parity risk of the same class as the lstrip fix here.
- (10) → separate PR (unrelated to strip_thinking): `TASK_CONFIG_FIELDS` is missing 8 real fields (`judge_model`, `max_length`, `answer_extractor`, …); deriving it from `dataclasses.fields(TaskConfig)` like the sampling fields.

gregorybchris (approved):

PR #370, opened 2026-09-11
Remove the plans folder

gregorybchris:
The only file in `plans/` was `001_oracle_no_gpu_task_pipeline_tests.md`, which described work that already landed in `tests/evals/tasks/test_oracle_no_gpu_pipeline.py`. Nothing else references it, so the folder is removed.

Note: `src/olmo_eval/common/scorers/citation_validation.py` mentions `plans/002_science_literature_evals.md` in two comments. That file was never in the repo, so those pointers were already dangling; left alone here to keep this change to a plain deletion.


abhishekraok (approved):
Thanks for the cleanup.

PR #383, opened 2026-09-17
Add MMLU-Pro tasks (5-shot MC, 5-shot cloze, 0-shot CoT)

gregorybchris:
## Description

Ports MMLU-Pro (`TIGER-Lab/MMLU-Pro`, 12,032 test questions, up to 10 options, 14 categories) from oe-eval. Three recipes are registered per category, following the naming used by the existing `mmlu` and `gpqa` ports:

| oe-eval config | olmo-eval task | Recipe |
|---|---|---|
| `mmlu_pro_{cat}:mc::none` | `mmlu_pro_{cat}` (aliases `:mc`, `:olmo3base`) | 5-shot logprob MC, letters A–J as continuations, acc_raw |
| `mmlu_pro_{cat}:rc::none` | `mmlu_pro_{cat}:rc` (alias `:olmo3base`, variant `:bpb`) | 5-shot logprob cloze, acc_per_char |
| `mmlu_pro_{cat}:0shot_cot::tulu3` | `mmlu_pro_{cat}:cot` | 0-shot chat CoT, temperature 0.6 / top_p 0.95, no token cap, shared OLMo 3 answer cascade (A–J) |

Suites: `mmlu_pro`, `mmlu_pro:bpb`, `mmlu_pro:cot`, each averaging over the 14 category tasks.

The dataset is a single HF config with a `category` column, so each task loads the split and filters in `process_doc`. Few-shot exemplars are the dataset's `validation` rows for the category, in dataset order, as in oe-eval. The dataset category `"computer science"` maps to the task slug `computer_science`.

Files:
- `src/olmo_eval/evals/tasks/mmlu_pro.py`: MC, RC, and CoT task classes, the CoT extractor and scorer, and the registration loop.
- `src/olmo_eval/evals/suites/mmlu_pro.py`: the three suites.
- `tests/evals/tasks/test_mmlu_pro.py`: registration, prompt shapes, category filtering, few-shot slicing, extraction cascade, suite sizes.

### Parity with oe-eval

- Prompts, continuations, and gold labels are byte-identical to oe-eval's own request builder on all 12,032 test questions for all three recipes (36,096 comparisons, 0 mismatches). oe-eval's classes were run directly in its own environment and diffed against `format_request` output here.
- The CoT recipe follows the `mmlu_{subject}:cot` and `gpqa:cot` ports rather than oe-eval's MMLU-Pro config, per review: sampling is temperature 0.6 / top_p 0.95 with no `max_tokens` cap (oe-eval uses greedy decoding with 2048 tokens, which degrades reasoning models), and the answer is extracted with the shared `OLMO_3_ANSWER_REGEX_TEMPLATES` cascade using an A–J answer regex, so it accepts the same phrasings MMLU CoT does. The exact "Therefore, the answer is (X)" phrasing still scores 1.0 first. `extract_think_answer` is applied first, as in `mmlu_cot.py`.

### Beaker runs against OLMo 2

Full `mmlu_pro` suite (5-shot MC), all 12,032 questions, 1×H100 each (group `chrisg-mmlu-pro-olmo2`, workspace `ai2/olmo-eval-debug`):

| Model | Instance-weighted (micro) | OLMo 2 model card (oe-eval) | Unweighted suite avg |
|---|---|---|---|
| OLMo-2-1124-7B | 30.77 | 31.0 | 33.02 |
| OLMo-2-1124-13B | 34.66 | 35.1 | 37.18 |

The 13B model leads the 7B model in 13 of 14 categories.

`mmlu_pro:cot` on OLMo-2-1124-7B-Instruct with 50 questions per category, 1×H100 (same group):

| CoT configuration | Responses with a letter | Unweighted suite avg |
|---|---|---|
| Greedy, 2048 tokens, MMLU-Pro cascade (initial) | 699 / 700 | 37.6 |
| Temperature 0.6 / top_p 0.95, shared cascade (current) | 700 / 700 | 35.1 |

### Known divergences

- oe-eval's `mmlu_pro` suites report instance-weighted micro averages (categories range from 381 to 1,351 questions). The suite registry only supports unweighted per-task averages, so the suite roll-up differs while per-category scores match. A weighted aggregation strategy is left for a separate PR.
- The RC recipe skips oe-eval's unconditioned "Answer:" pass (acc_uncond), matching the existing `mmlu_{subject}:rc` port.
- The `:cot` sampling parameters and answer cascade match the MMLU and GPQA CoT ports instead of oe-eval's MMLU-Pro config (see Parity above), so CoT scores are not directly comparable to oe-eval's `mmlu_pro:0shot_cot::tulu3`.
- The 5-shot CoT recipes (`:cot::none`, `:cot::llama3.1`) are not ported.

`build_predictions` now exports the `answer_format_correct` value the scorer records in output metadata (this also covers GPQA CoT), and CoT instances no longer carry `gold_idx`, so the prediction `label` is the gold letter.

## Type of Change

- [ ] Documentation update

## Testing

- [x] Unit tests pass locally (`pytest tests/ --ignore=tests/integration/`): 2,380 passed, 23 skipped
- [x] New tests added for new functionality: 36 tests in `tests/evals/tasks/test_mmlu_pro.py`, plus an `answer_format_correct` export test in `tests/runners/io/test_builders.py`

Also: `ruff check`, `ruff format --check`, and `ty check` clean (one pre-existing ty diagnostic in an unrelated sandbox file); `olmo-eval suite inspect` and `task inspect` on real data; the oe-eval parity scripts and Beaker runs described above.

## Checklist

- [x] I have added/updated documentation as needed (module docstrings; README has no per-task table)


src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,352 @@
...
+class MMLUProCoTExactMatchScorer(Scorer):
+    """Case-insensitive exact match on the extracted answer letter."""
+
+    name: str = "exact_match"
+
+    def score(self, instance: Instance, output: LMOutput) -> float:
+        answer, format_correct = _extract_cot_answer(output.text or "")
+        # Whether the letter was stated in the requested format, independent
+        # of correctness: a low score means fallback extraction was needed.
+        output.metadata["answer_format_correct"] = format_correct
```
gregorybchris, line 257: Non-blocking 1: this value never reaches the predictions file. `build_predictions` (`runners/io/builders.py:96`) copies only a fixed set of metadata keys; I confirmed 0 of 700 CoT prediction rows from the Beaker run contain `answer_format_correct`. Framework gap (GPQA CoT has it too); a two-line addition beside the `judge_result` block plus a `test_builders.py` case fixes it, or file an issue and link it before merge. Details in the summary comment.

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,352 @@
...
+            return None
+        question, options, gold_idx = parsed
+        query = f"Question: {question}\n" + "".join(
+            f" ({label}) {option}\n" for label, option in zip(_CHOICE_LABELS, options, strict=False)
+        )
+        return Instance(
+            question=query,
+            gold_answer=_CHOICE_LABELS[gold_idx],
+            choices=tuple(options),
+            metadata=_metadata(doc, index, self.category, gold_idx),
```
gregorybchris: Non-blocking 2: `gold_idx` here makes `build_predictions` (`builders.py:102`) write an int `label` (e.g. `2`) next to a letter `extracted_answer` (`"C"`) for every CoT row; oe-eval's label for this task is the letter. `gold_idx` is only consumed for multi-output logprob responses, so please drop it from the CoT instance metadata (or fix the preference in `build_predictions`, which would also cover GPQA CoT).

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,352 @@
...
+        formatter = self.config.formatter
+        if formatter is None:
+            raise ValueError("MMLU-Pro CoT task requires a formatter")
+        return formatter.format(instance, None)
+
+
+for _category in MMLU_PRO_CATEGORIES:
+    _slug = category_slug(_category)
+    _name = f"mmlu_pro_{_slug}"
+    _source = DataSource(path=MMLU_PRO_PATH, split="test")
```
gregorybchris: Non-blocking 3: no `revision` pin. Current Hub sha is `b189ec765aa7ed75c8acfea42df31fdae71f97be` (last modified 2026-05-02) and MMLU-Pro has been edited in place before. `revision` is in `DataSource.to_dict()`, so pinning after results exist changes every task hash; pinning now to the sha your parity run used keeps the numbers. `frontierscience.py` and `astabench_sqa.py` do this.

Nit: `split="test"` is overwritten by `get_data_source(split=config.split)`, so it is inert (same in `mmlu.py`).

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,352 @@
...
+        return 1.0 if answer.upper() == gold.upper() else 0.0
+
+
+_COT_ACCURACY = AccuracyMetric(name="exact_match", scorer=MMLUProCoTExactMatchScorer)
+
+
+class MMLUProCoTTask(MMLUProTask):
+    """0-shot chain-of-thought in chat format for one category."""
+
+    formatter = _COT_FORMATTER
```
gregorybchris, line 268: Non-blocking 4: `mmlu_cot.py` and `gpqa.py` set `strip_thinking = True` on their CoT classes; this task strips inside the extractor instead. Scores agree, but with `strip_thinking` the runner records `original_text` and stores the stripped answer as `text`, so prediction shapes differ across the CoT families. It enters the hash only when True, so worth deciding before merge. Should this task set it too?

src/olmo_eval/evals/suites/mmlu_pro.py
```diff
@@ -0,0 +1,25 @@
+from olmo_eval.evals.suites.registry import AggregationStrategy, make_suite
+from olmo_eval.evals.tasks.mmlu_pro import MMLU_PRO_CATEGORIES, category_slug
+
+_SLUGS = tuple(category_slug(category) for category in MMLU_PRO_CATEGORIES)
+
+MMLU_PRO = make_suite(
```
gregorybchris, line 6: Non-blocking 5: oe-eval's `mmlu_pro:mc::none` is a micro (instance-weighted) average; this suite is macro (33.02 vs 30.77 on OLMo-2-7B). The PR names this. Please file the weighted-aggregation follow-up as an issue and link it before merge, and consider noting the divergence in this suite's description.

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,352 @@
...
+            "formatter": _MC_FORMATTER,
+            "metrics": (LogprobMCAccuracyMetric(),),
+            "primary_metric": LogprobMCAccuracyMetric(),
+            "num_fewshot": 5,
+            "sampling_params": SamplingParams(max_tokens=1, temperature=0.0),
+            "__module__": __name__,
+            "__qualname__": f"MMLUPro_{_slug}",
+        },
+    )
+    globals()[_mc_cls.__name__] = _mc_cls
```
gregorybchris: Nit: `register()` already `setattr`s dynamically created classes onto their module for pickling, so this line (and the two like it at 331 and 351) is redundant. The `type(...)` loop follows `mmlu.py`/`gpqa.py`, so not asking to change that here.

gregorybchris (commented):

Looks good overall! Nothing blocking. Five non-blocking items, two of which change the task hash if done after results are stored, so they are worth deciding before merge.

**What I verified (no need to re-check):**
- `ruff check`, `ruff format --check` clean; `ty` shows only the pre-existing swerex diagnostic in an untouched file; `pytest tests/ --ignore=tests/integration`: 2318 passed, 8 skipped. CI is green.
- Parity: I re-ran `diff_prompts.py` on your dumps and also regenerated the olmo-eval dump from this branch's HEAD (your dumps predate the final commit by about 90 minutes): 36,096 prompt comparisons, 0 mismatches; 40 extractor cases, 0 mismatches; 12,032 unique `question_id`s.
- Beaker: recomputed from `metrics.json`: 7B micro 30.77 / macro 33.02, 13B micro 34.66 / macro 37.18, 0 failed instances, category sizes 381 to 1351. CoT: 700 rows, 699 with an extracted letter, 37.57%.
- Task identity: `mmlu_pro_math`, `:mc`, and `:olmo3base` share hash `ae38cf9e7f6fae6d` (intended alias); `:rc`, `:rc:bpb`, `:cot`, and the other categories all differ. The CoT prompt text lives in `ChatFormatter.user_template`, so it is in the hash.
- Registration: task and suite auto-discovery pick up both new modules; `task inspect` and `suite inspect` work on real data. `limit` is applied after category filtering (`runners/asynq/preparation.py:57`), which is why the 50-per-category CoT run worked.
- Extractor: I read `extract_answer_with_format` and agree it cannot reproduce this cascade (first-match "answer is", the `.*[aA]nswer:` per-line rule, and 0.0 rather than 0.2/0.1 for the bare-letter fallback). `<think>` traces are dropped before matching.
- Sampling: 2048 tokens, temperature 0. vLLM maps `do_sample=True, temperature=0.0` to greedy (`providers/vllm.py:223`), matching oe-eval's `do_sample: False`; `stop_sequences=None` matches oe-eval's `[]`.
- No Ai2 paths in source; no open PRs overlap; the registries reject task/suite name collisions and the tests exercise them.

### Non-blocking

1. `src/olmo_eval/evals/tasks/mmlu_pro.py:249` (framework: `src/olmo_eval/runners/io/builders.py:96`)
   The scorer writes `answer_format_correct` to `output.metadata`, but `build_predictions` copies only a fixed set of keys, so it never reaches the predictions file. I confirmed 0 of 700 CoT prediction rows contain it. oe-eval reports it as an extra metric, so after a run here a parse failure cannot be told apart from a wrong answer. This is a deficiency in `build_predictions`, not in this PR, and GPQA CoT has the same gap. The fix is small; could we either include it here or file an issue and link it from the PR before merge?
   ```python
   # builders.py, beside the judge_result block
   if "answer_format_correct" in meta:
       out_data["answer_format_correct"] = meta["answer_format_correct"]
   ```
   plus a `tests/runners/io/test_builders.py` case asserting the key survives.

2. `src/olmo_eval/evals/tasks/mmlu_pro.py:279`
   The CoT instance carries `gold_idx` in metadata, and `build_predictions` (`builders.py:102`) prefers `gold_idx` over `gold_answer` for `label`. Result: every CoT prediction row has an int `label` (for example `2`) next to a letter `extracted_answer` (`"C"`). oe-eval's label for this task is the letter. `gold_idx` is only consumed for multi-output logprob responses, so the CoT instance does not need it. Please drop it from the CoT metadata. The framework alternative is for `build_predictions` to prefer `gold_answer` when a response has a single output, which would also fix GPQA CoT; an issue for that is fine.

3. `src/olmo_eval/evals/tasks/mmlu_pro.py:295`
   The `DataSource` has no `revision`. The Hub dataset was last modified 2026-05-02 (sha `b189ec765aa7ed75c8acfea42df31fdae71f97be`) and MMLU-Pro has been edited in place before. `revision` is part of `DataSource.to_dict()`, so pinning later changes every task hash once results exist. Could we pin now, as `frontierscience.py` and `astabench_sqa.py` do? Your parity run used the current revision, so pinning to that sha keeps the numbers.

4. `src/olmo_eval/evals/tasks/mmlu_pro.py:260`
   `mmlu_cot.py` and `gpqa.py` set `strip_thinking = True` on their CoT classes; this one strips inside the extractor instead. Scores agree either way, but with `strip_thinking` the runner records `original_text` and stores the stripped answer as `text`, so predictions from the three CoT task families would differ in shape. `strip_thinking` enters the hash only when True, so this is another decide-before-merge item. Should this task set it too?

5. `src/olmo_eval/evals/suites/mmlu_pro.py:6`
   oe-eval's `mmlu_pro:mc::none` is a micro (instance-weighted) average; this suite is macro, and on OLMo-2-7B that is 33.02 vs 30.77. The PR names the divergence. Please file the weighted-aggregation follow-up as an issue (`AggregationStrategy` has no weighted option) and link it here before merge, and consider noting the divergence in the `mmlu_pro` suite description so a reader comparing against the model card is not surprised.

### Nits

- Nit: `src/olmo_eval/evals/tasks/mmlu_pro.py:312` (also 331, 351): `register()` already `setattr`s dynamically created classes onto their module for pickling, so the `globals()[...]` lines are redundant. The `type(...)` registration loop follows `mmlu.py` and `gpqa.py`, so I am not asking to change it here.
- Nit: `src/olmo_eval/evals/tasks/mmlu_pro.py:49`: this is the second module importing `_format_rc` and `_make_mcq_prompt` from `mmlu.py`; dropping the underscore or moving them to a shared helper would make the dependency explicit.
- Nit: `src/olmo_eval/evals/tasks/mmlu_pro.py:295`: `split="test"` on the `DataSource` is overwritten by `get_data_source(split=config.split)`, so it is inert (same in `mmlu.py`).
- Nit: `src/olmo_eval/evals/tasks/mmlu_pro.py:1` and `src/olmo_eval/evals/suites/mmlu_pro.py:10,17,24`: the exact count ("14 categories") in the docstring and suite descriptions will go stale; `MMLU_PRO_CATEGORIES` and the test already pin it.

### Before merge

Nothing blocks. Please decide items 3 and 4 (both change the hash if done after results are stored), address item 2, and link follow-up issues for items 1 and 5 (or fix 1 inline). After that, merge once CI is green.

gregorybchris:
Thanks for the review. Addressed in 038c574a:

- **answer_format_correct not in predictions**: fixed at the framework level. `build_predictions` now exports `answer_format_correct` from output metadata next to `judge_result`, with a test in `tests/runners/io/test_builders.py`. This also covers GPQA CoT.
- **CoT label is an index**: `gold_idx` is no longer set on CoT instances (only the logprob tasks need it), so the prediction `label` is now the gold letter that scoring compares against.
- **Dataset revision**: pinned to `b189ec765aa7ed75c8acfea42df31fdae71f97be`, the revision the parity check and the Beaker runs used.
- **strip_thinking**: set on the CoT task to match `mmlu_cot` and `gpqa` CoT. Scores are unchanged since the extractor already strips; prediction text shape is now consistent.
- **Micro vs macro**: filed #384 for an instance-weighted aggregation strategy. Suite descriptions now state the unweighted roll-up and the reference harness's micro average.
- **Nits**: removed the redundant `globals()` assignments and the inert `split="test"` argument; the module docstring no longer states the category count. The import of `_make_mcq_prompt` and `_format_rc` from `mmlu.py` stays as is: both reproduce oe-eval's shared `make_mcq_prompt` / `make_cloze_prompt`, and one copy keeps the two ports from drifting apart. Happy to promote them to public names in a follow-up if preferred.

Full suite: 2,319 passed, 8 skipped. The revision, `strip_thinking`, and metadata changes alter the task hashes, which is intended before any results are stored.

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Fixed in 038c574a: `build_predictions` now exports `answer_format_correct` from output metadata (next to `judge_result`), with a test in `tests/runners/io/test_builders.py`. GPQA CoT gets the same fix.

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Fixed in 038c574a: CoT instances no longer carry `gold_idx`, so the predictions `label` is the gold letter.

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Fixed in 038c574a: pinned to `b189ec765aa7ed75c8acfea42df31fdae71f97be`, the revision the parity check and Beaker runs used.

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Fixed in 038c574a: `strip_thinking = True` on the CoT task, matching `mmlu_cot` and `gpqa`. Scores unchanged; prediction text shape now consistent.

src/olmo_eval/evals/suites/mmlu_pro.py
gregorybchris: Filed #384 for an instance-weighted aggregation strategy. In 038c574a the suite descriptions state the unweighted roll-up and the reference harness's micro average.

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Removed all three `globals()` assignments in 038c574a; verified the module attributes and task pickling still work through `register()`.

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,355 @@
...
+    "\nAnswer the above question and REMEMBER to finish your response with the exact phrase "
+    '"Therefore, the answer is (ANSWER_LETTER)" where (ANSWER_LETTER) is one of (A), (B), (C), '
+    "(D), (E), etc."
+)
+
+_COT_FORMATTER = ChatFormatter(
+    user_template=_COT_DESCRIPTION + "{question}" + _COT_FINAL_DESCRIPTION,
+)
+
+_COT_SAMPLING = SamplingParams(max_tokens=2048, temperature=0.0)
```
mgmorgan23: Not blocking -- This matches the current implementation in oe-eval, but I think the chat version was not updated recently -- MMLU Pro was run on the base model but not the posttrained one for Olmo3. I think we should consider matching these sampling parameters to the ones that we use for MMLU. Temp = 0 for reasoning models can really degrade responses.

src/olmo_eval/evals/tasks/mmlu_pro.py
```diff
@@ -0,0 +1,355 @@
...
+
+_ANSWER_FORMAT_RE = re.compile(r"Therefore, the answer is \(([A-J])\)")
+_ANSWER_IS_RE = re.compile(r"(?i)answer is:?\s*\(?([A-J])\)?")
+_ANSWER_COLON_RE = re.compile(r".*[aA]nswer:\s*([A-J])")
+_STANDALONE_LETTER_RE = re.compile(r".*\b([A-J])\b")
+
```
mgmorgan23, line 225: Not blocking -- the formatting in the prompt for the question is almost exactly the same as for MMLU, but we accept a lot fewer regex matches here I think. That matches the oe-eval implementation, but again it might be worth consolidating/matching

mgmorgan23:
While I think we probably don't want to give up the ability to run a single MMLU-Pro category at a time, one possible additional solution to the weighting problem would be to run the eval as a single task and use `SubsetAccuracyMetric(Metric)` to get the per-category results. It also might be worth running a newer reasoning model through oe-eval and this to check for parity on those extractions. Otherwise, this looks good to me!

mgmorgan23 (approved):

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Ah good call! Updated to match MMLU

src/olmo_eval/evals/tasks/mmlu_pro.py
gregorybchris: Updated to use the shared extraction utils. Thanks!

PR #385, opened 2026-09-17
Add an instance-weighted (micro) suite aggregation strategy

### LINKED ISSUE #384: Suite aggregation: add an instance-weighted (micro) average strategy
## Problem

`AggregationStrategy` in `src/olmo_eval/evals/suites/registry.py` offers `NONE`, `AVERAGE`, `AVERAGE_OF_AVERAGES`, and `DISPLAY_ONLY`. `AVERAGE` is an unweighted mean over the suite's leaf tasks. oe-eval's suite configs also support `"primary_metric": "micro"`, an instance-count-weighted mean across tasks, and several benchmarks are reported that way.

MMLU-Pro (#383) is the first port where this matters: its 14 category tasks range from 381 to 1,351 questions, and oe-eval's `mmlu_pro:mc::none` suite is micro-averaged. On OLMo-2-1124-7B the unweighted suite average is 33.02 while the micro average is 30.77 (model card: 31.0). Per-task scores match oe-eval; only the roll-up differs.

## What is needed

- A new `AggregationStrategy` value (for example `WEIGHTED_AVERAGE`) that weights each leaf task by its instance count.
- Instance counts plumbed to where suite scores are computed. `compute_scope_score` in `src/olmo_eval/analysis/scope_scores.py` currently receives only per-task aggregate floats, so the count has to come along from results storage and the viewer, and the CLI `suite` commands need to display the strategy.
- Switch `mmlu_pro`, `mmlu_pro:bpb`, and `mmlu_pro:cot` to the weighted strategy once it exists, and audit other ported suites whose oe-eval counterparts are micro-averaged.

gregorybchris:
Closes #384.

Rebased onto `main` after `chrisg/mmlu-pro` (#383) merged; the diff now contains only the aggregation change.

## What changed

`AggregationStrategy.WEIGHTED_AVERAGE` weights each of a suite's leaf tasks by its instance count, matching oe-eval's `"primary_metric": "micro"` — `sum(metric * num_instances) / sum(num_instances)`, as computed in `oe_eval/tasks/aggregate_tasks.py`. `mmlu_pro`, `mmlu_pro:bpb` and `mmlu_pro:cot` now use it. The 14 category tasks range from 381 to 1,351 questions, so per issue #384 the MC suite roll-up on OLMo-2-1124-7B moves from 33.02 to 30.77, against a model card figure of 31.0.

Instance counts had to reach every place a suite score is computed:

- `compute_suite_aggregations` reads `num_instances` from the task results it already receives.
- `compute_scope_score` takes a new `task_instance_counts_by_name` argument, keyed and ordered to match `task_scores_by_name` so task-hash variants collapse the same way.
- The viewer results table selects `TaskResult.num_instances` and publishes a `task_instance_counts` map per model.
- `PairwiseResult` carries `model_task_instance_counts` alongside `model_task_scores`.

When any contributing task lacks a usable instance count, the aggregate falls back to the unweighted mean. Mixing weighted and unweighted terms would produce a number that belongs to neither convention, and silently dropping the unweighted tasks would change what the average covers.

That fallback is decided **once per scope**, not per model. `TaskResult.num_instances` is nullable, so a run imported from an older `metrics.json` can land with no counts; deciding per model would put a micro average for one run beside a macro average for another in the same sorted column. If any compared run is missing a count, none of them are weighted, and the viewer's aggregate tooltip says so. `scope_scores.weights_are_complete` answers that question for a scope. The runner, which publishes one run's number rather than a comparison, still decides per run and logs which tasks were missing.

A `WEIGHTED_AVERAGE` suite nested inside an `AVERAGE_OF_AVERAGES` parent is collapsed with its own weighting, in the runner (`_compute_child_average`), in the viewer (`_child_scope_score`), and in the runner's summary table, whose collapse check previously tested `aggregation == "average"` by string. No registered suite nests a weighted child today; this keeps the next one from being silently wrong.

The CLI already prints `suite.aggregation.value` in both `olmo-eval suites` and `olmo-eval suite inspect`, so the new strategy shows up there without further changes. `README.md` documents it.

## Parity notes

- **`mmlu_pro`** (5-shot MC) matches oe-eval's `mmlu_pro:mc::none`, which is micro. Unambiguous.
- **`mmlu_pro:bpb`** has no oe-eval counterpart suite. Instance weighting is exactly right here anyway: the task's metric is `BPBMetricInstanceAvg`, so weighting each category's BPB by its instance count reproduces the bits per byte of the pooled question set.
- **`mmlu_pro:cot`** ports the prompt of oe-eval's `mmlu_pro:0shot_cot::tulu3`, which **macro**-averages (its sampling and answer extraction follow the MMLU and GPQA CoT ports, per #383's review). oe-eval's two micro CoT suites (`mmlu_pro:cot::none`, `mmlu_pro:cot::llama3.1`) are 5-shot and are not what we ported. Per #384 this suite is still weighted, so that all three MMLU-Pro suites here roll up the same way as the benchmark's own pooled reporting, but its description now names the divergence rather than claiming a parity that does not hold.

## Audit of other ported suites

I enumerated oe-eval's `TASK_SUITE_CONFIGS` programmatically. Nineteen suites are micro: `bbh:*` (nine variants), `bfcl_live_ast_micro::std`, `bfcl_live_micro::std`, `deepmind_math::tulu`, `deepmind_math::olmo3:heldout`, `minerva_math::llama3.1`, `mmlu::olmo1`, and the four `mmlu_pro:mc/cot` suites.

**MMLU-Pro is the only one with a counterpart here.** We have no bbh, bfcl, or deepmind_math suites. Our `mmlu` follows `mmlu:*::olmes` (macro), not `mmlu::olmo1`. Our `minerva_math` follows `minerva_math::olmes` (macro), not `::llama3.1`.

I had earlier flagged `multipl_e_humaneval:olmo3base:v2` as a possible micro counterpart. That was wrong: oe-eval's `multipl_e_humaneval:6lang::olmo3:n32:v2` has no `primary_metric` key at all, and `TASK_GROUPS_RAW` only admits suites that have one, so oe-eval computes no aggregate for it. Nothing to reconcile.

## Testing

`uv run pytest tests/ --ignore=tests/integration` — 2402 passed, 23 skipped (after the rebase onto `main`). Lint, format, and `ty check` are clean apart from one pre-existing diagnostic in `harness/sandbox/executor.py`.

New coverage: weighted runner aggregation and its fallback, nested weighted children in both the runner and `compute_scope_score` plus the summary-table collapse, `compute_scope_score` weighting including hash-variant collapsing and every fallback path, `weights_are_complete`, the viewer results table exposing instance counts, and the per-scope fallback in both the viewer and the pairwise payload.

## Validation on a real run

Launched `olmo-2-7b` (`allenai/OLMo-2-1124-7B`) on the `mmlu_pro` suite at commit `b14272a`: [beaker.org/ex/01M2RG6VM9N6PA4D3JGY0J9XEN](https://beaker.org/ex/01M2RG6VM9N6PA4D3JGY0J9XEN). All 14 tasks succeeded, 12,032 instances scored, exit code 0.

| task | score | instances |
|---|---:|---:|
| mmlu_pro_biology | 59.41% | 717 |
| mmlu_pro_business | 24.33% | 789 |
| mmlu_pro_chemistry | 18.99% | 1132 |
| mmlu_pro_computer_science | 29.76% | 410 |
| mmlu_pro_economics | 45.85% | 844 |
| mmlu_pro_engineering | 20.54% | 969 |
| mmlu_pro_health | 38.14% | 818 |
| mmlu_pro_history | 38.58% | 381 |
| mmlu_pro_law | 24.16% | 1101 |
| mmlu_pro_math | 17.25% | 1351 |
| mmlu_pro_other | 34.20% | 924 |
| mmlu_pro_philosophy | 36.27% | 499 |
| mmlu_pro_physics | 21.71% | 1299 |
| mmlu_pro_psychology | 53.13% | 798 |
| **total** | | **12,032** |

Computed from those per-task rows:

- unweighted (macro) mean: **33.0237**
- instance-weighted (micro) mean: **30.7680**

The runner's published suite aggregate was **0.3077** (`mmlu_pro | Success | primary_score:average | 0.3077`), so it published the weighted mean, not the unweighted one. Both figures reproduce #384 exactly: 33.02 unweighted, 30.77 micro, against a model card figure of 31.0. The 2.26-point gap is the whole point of the change, and it falls out of the spread in task sizes — 381 questions for history against 1,351 for math, with the large tasks scoring well below average.

This run covers the runner's weighted aggregation end to end. It does not cover the storage-boundary check or the viewer/pairwise read paths: it ran with `--no-store` because the launching machine had no AWS credentials, and `--store` fetches the Postgres password from AWS Secrets Manager inside the job. Those paths are covered by unit tests only.


gregorybchris (commented):

Looks good overall! The weighted math is right and well tested, and the rebase left the MMLU-Pro suites consistent with the task module. One thing should be fixed before merge: the new "instance counts are required" storage check breaks persistence for external evals, which never carried a count.

I found one blocking issue and four non-blocking ones. The blocking one is a persistence regression introduced by the last commit. The others are documentation drift from that same commit and two test gaps.

### Blocking

1. `src/olmo_eval/storage/base.py:181`
   The external runner builds each task with `"num_instances": result.metadata.get("num_tasks")` (`src/olmo_eval/runners/external/runner.py:303`). Only some external benchmarks set `num_tasks`: `terminal_bench` and `deepscholar` always do, `tau2` and `asta` only inside a success branch, and `scicode` never does. For those results `num_instances` is `None`, so `convert_runner_results` now raises. `save_results` catches that at `src/olmo_eval/runners/io/storage.py:265`, logs an error, and `continue`s. The whole experiment is skipped from every storage backend and the run still exits normally. The external runner's failure path (`runner.py:119`) goes through the same `_save_results`, so a failed external eval's error record is dropped from storage too. That is the "error information dropped on persistence" case from section 3.4.

   I reproduced this by calling `convert_runner_results` on a result dict shaped exactly like the external runner's for `scicode` with `num_instances=None`: it raises `ValueError: num_instances is required for task 'scicode'`. On `origin/main` the same input converts and stores `num_instances=None`.

   Please make the external runner always supply an integer count before this check lands. `len(result.predictions)` when predictions exist, else `num_tasks`, else `0` would match what `_write_metrics_json` already does with `task_data.get("num_instances") or 0` at `runner.py:360`. Please also add a regression test that runs an external-shaped result without `num_tasks` through `convert_runner_results`, and a direct test in `tests/storage/test_base.py:265` (`TestConvertRunnerResults`) asserting that a missing `num_instances` raises.

   The silence here is a pre-existing framework fault, not this PR's: `save_results` swallows converter errors and the run exits 0 with nothing stored. Could you file a follow-up issue asking whether a converter failure should fail the run, or at least be reported alongside the hard-failure gate?

### Non-blocking

1. `src/olmo_eval/evals/suites/registry.py:35`
   The `WEIGHTED_AVERAGE` docstring still says "Falls back to the unweighted average when any contributing task has no instance count." The last commit removed that fallback. Since this docstring is what `help(AggregationStrategy)` and the next contributor will read, please change it to say the aggregate is omitted (runner) or `None` (viewer) when a count is missing, matching `README.md:185`.

2. `src/olmo_eval/runners/processing/aggregation.py:363` (and `:159` for a nested child)
   When a weighted suite is omitted, the only trace is a `logger.warning`. The suite key is absent from `results["suites"]` in `metrics.json`, and the summary table has no row, so a reader of `metrics.json` cannot tell "omitted for missing counts" from "suite not requested". Today this branch is unreachable from the async runner, which always sets an integer count (`src/olmo_eval/runners/asynq/results.py:451`), so I am not asking for a change now. Should the omitted suite get an entry with an `"error"` field and no `metrics`, the way failed tasks do, so the omission is visible in the persisted file? A follow-up issue would be fine.

3. `tests/test_runner.py:977`
   `test_suite_aggregation_weighted_average_without_counts` covers the flat case. The nested case (a `WEIGHTED_AVERAGE` child under `AVERAGE_OF_AVERAGES` where one leaf lacks a count, so `_compute_child_average` returns `None` and the parent averages the remaining children) is tested for `compute_scope_score` at `tests/analysis/test_scope_scores.py:117` but not for the runner. Please add the runner counterpart so the two implementations cannot drift.

4. PR description
   Several paragraphs describe behavior the last commit removed: "the aggregate falls back to the unweighted mean", "That fallback is decided once per scope", "`scope_scores.weights_are_complete` answers that question", "the viewer's aggregate tooltip says so", and the Testing section lists `weights_are_complete` and "the per-scope fallback". The PR body is the record of what merged, so please bring it in line with the code before merge.

### Nits

- Nit: `src/olmo_eval/cli/results/viewer_server.py:570` relies on operator precedence: `int(num_instances) or None if num_instances is not None else None` parses as `(int(x) or None) if x is not None else None`. It is correct, but parentheses would save the next reader a trip to the grammar.
- Nit: `_task_weight` exists in both `runners/processing/aggregation.py:66` (returns `0.0` for missing) and `analysis/scope_scores.py:24` (returns `None` for missing). Same name, different contract. Consider naming one to reflect the difference.

### What I verified

- `uv run ruff check src/ tests/` clean. `uv run ruff format --check` clean (535 files). `uv run ty check src/ alembic/` reports only the pre-existing unused `type: ignore` in `harness/sandbox/executor.py:788`. `uv run pytest tests/ --ignore=tests/integration`: 2402 passed, 23 skipped. All match the PR description.
- Rebase consistency: expanded all three MMLU-Pro suites (14 tasks each) and resolved every spec through `task_exists`. None missing, no task name collides with a suite name. Primary metrics are `LogprobMCAccuracyMetric` (`mmlu_pro`), `BPBMetricInstanceAvg` (`mmlu_pro:bpb`), and `AccuracyMetric` over `MMLUProCoTExactMatchScorer` (`mmlu_pro:cot`), so the "weighting an instance-averaged BPB reproduces the pooled BPB" claim holds.
- Weighted math: `_weighted_mean` uses `zip(strict=True)` over lists that are appended in lockstep, and `_tasks_missing_instance_counts` rejects `<= 0` before any division, so denominators are positive. Recomputed the PR body's table: 12,032 instances, macro 33.02, micro 30.77.
- Task identity and persisted formats: `aggregation` is a `Suite` attribute, not a `TaskConfig` field, so no task hash changes. `convert_runner_results` does not persist suites, so Postgres rows are unchanged. `metrics.json` `suites[*].aggregation` can now be `"weighted_average"`; the only consumer that switches on it (`log_summary`) was updated. `PairwiseResult.model_task_instance_counts` is built per request and not persisted. `TaskResult.num_instances` stays nullable and unchanged, so no Alembic migration is needed.
- Hash-variant collapse in `compute_scope_score`: `_group_model_task_scores_by_name` and `_group_model_task_instance_counts_by_name` iterate the same `columns` list, and the pairwise payload's two `_group_by_task_name` calls iterate the same `task_entries`, so scores and counts align by position. `zip_longest` treats a length mismatch as a missing count. The "mean of variant counts" choice is documented in the docstring and tested.
- Viewer query: `func.greatest`/`coalesce` is Postgres-only, and the repo has no SQLite path. The pairwise path already recovers counts from predictions in `_update_task_rows_num_instances`, so the two surfaces agree.
- Runner: `asynq/results.py:451` always writes an integer `num_instances`, including for failed tasks, so the "omit the aggregate" branch is reachable only from hand-built or imported result dicts.

### Before merge

Fix Blocking 1 (external runner supplies a count; regression tests for the converter) and update the `registry.py` docstring and the PR description. Please file the two follow-up issues named above (converter failures surfacing; omitted weighted suites leaving a trace in `metrics.json`). After that you may merge once CI is green.

src/olmo_eval/storage/base.py
```diff
@@ -175,6 +175,12 @@ def convert_runner_results(
         if not task_hash:
             raise ValueError(f"task_hash is required for task '{spec}'")
 
+        # num_instances is required: suite aggregation weights tasks by it, and a
+        # result stored without one can never be weighted afterwards.
+        num_instances = task_data.get("num_instances")
+        if num_instances is None:
```
gregorybchris, line 181: Blocking 1 in the summary. The external runner sets `num_instances` to `result.metadata.get("num_tasks")` (`runners/external/runner.py:303`), which is `None` for `scicode` and for `tau2`/`asta` outside their success branch. This raise is then caught by `save_results` (`runners/io/storage.py:265`), which logs and `continue`s, so the whole experiment is skipped from storage and the run exits normally. I reproduced it with an external-shaped `scicode` result: `ValueError: num_instances is required for task 'scicode'`. On `origin/main` the same input stores with `num_instances=None`.

Please have the external runner always supply an integer count (predictions length, else `num_tasks`, else 0, matching `runner.py:360`), and add a converter test for a result without `num_tasks` plus a direct test that a missing `num_instances` raises.

src/olmo_eval/evals/suites/registry.py
```diff
@@ -30,12 +30,17 @@ class AggregationStrategy(StrEnum):
     Attributes:
         NONE: No aggregation - just collect individual task results.
         AVERAGE: Compute simple average of all task scores.
+        WEIGHTED_AVERAGE: Average of all task scores, each weighted by the
+            task's instance count. This is the instance-weighted ("micro")
+            average that oe-eval reports for several suites. Falls back to the
```
gregorybchris, line 35: Non-blocking 1 in the summary. This still describes the fallback that the last commit removed. Please change it to say the aggregate is omitted (runner) or `None` (viewer) when a contributing task has no count, matching `README.md:185`.

src/olmo_eval/runners/processing/aggregation.py
```diff
@@ -289,44 +349,54 @@ def compute_suite_aggregations(
...
+            suite_specs = [f"{task_spec}{priority_suffix}" for task_spec in suite.expand()]
+            tasks_included = _contributing_task_specs(suite_specs, task_results)
+            if not tasks_included:
+                continue
 
-                if full_task_spec not in task_results:
+            if weighted:
+                missing = _tasks_missing_instance_counts(tasks_included, task_results)
+                if missing:
+                    _log_omitted_weighted_suite(spec, missing)
```
gregorybchris, line 363: Non-blocking 2 in the summary. When this branch fires, the suite key is absent from `results["suites"]` and the summary table, so a `metrics.json` reader cannot tell "omitted for missing counts" from "not requested". Unreachable from the async runner today (`asynq/results.py:451` always writes an int), so no change requested now. Should the omitted suite get an entry with an `"error"` and no `metrics`, the way failed tasks do? A follow-up issue would be fine.

src/olmo_eval/cli/results/viewer_server.py
```diff
@@ -783,6 +820,24 @@ def _group_model_task_scores_by_name(
     return grouped_scores
 
 
+def _group_model_task_instance_counts_by_name(
```
mgmorgan23, line 823: This function is the same as the one above (806-820) -- the only difference is the variable name

gregorybchris: Thanks for this catch :) wouldn't it be cool if coding agents kept a store of function embeddings to deduplicate code without huge context windows..

gregorybchris:
@mgmorgan23 could I please request your re-review when you get a chance?

mgmorgan23 (approved):
Looks good to me!

PR #387, opened 2026-09-18
Add BFCL v3 single-turn categories as a completion task

pdasigi:
Registers the BFCL v3 single-turn categories, reformulated as a completion so a
pretrained model with no chat template can be measured on them. The chat regimes
BFCL was built for (native function calling, and BFCL's own prompting mode) follow
in a separate PR.

## What this adds

- `common/scorers/bfcl/` — the vendored AST checker, decoder, Java/JavaScript
  parsers and value converters, and one scorer that dispatches on category.
- `evals/tasks/bfcl/` — 15 tasks (13 categories plus two pooled live tasks), the
  completion formatter, hand-written exemplars per language, and the benchmark's
  own README; the top-level README keeps a pointer rather than the detail.
- `evals/suites/bfcl.py` — five suites reproducing BFCL's own summaries.
- 50 tests, offline (no dataset download).

The exemplars are written in the task module rather than drawn from a dataset, so
`fewshot_source` names the set that produced a run's prompts — the way
`minerva_math` names its fixed set — with a digest of the exemplar text, so that
revising an exemplar cannot reuse an earlier run's task hash.

```bash
uv run olmo-eval run -m my-base-model -t bfcl
uv run olmo-eval run -m my-base-model -t bfcl_simple:2shot   # :0shot .. :5shot, default 5
```

## Why vendored

`bfcl-eval` on PyPI pins `numpy==1.26.4` and pulls faiss, sentence-transformers,
anthropic and cohere. That does not coexist with the vLLM environment here, so the
checker is vendored instead. Java and JavaScript need tree-sitter grammars, declared
as task-specific dependencies so no other category installs them.

## Fidelity

Replaying BFCL's own possible answers back through the checker reproduces
**999/1000** non-live Python, **1344/1351** live, **92/100** Java, **49/50**
JavaScript. Every remaining case is a dataset quirk the reference checker fails
identically — e.g. `simple_363`, whose possible answer names `find_closest` while its
function document names `restaurant_search.find_closest`.

Two deliberate departures, both commented in the source:

1. An arithmetic argument is folded rather than passed to `eval`, which would run
   whatever the model wrote.
2. The v3 prompt is reproduced verbatim, including its wording slips and its
   rendering of function documents as a Python repr under a sentence calling it
   JSON — changing either moves the scores.

## Results

Full suite, 3641 instances per run, a100. Default is 5 exemplars.

### 5-shot

| Category | n | Olmo-3-1025-7B | Qwen3.5-4B-Base |
|---|---|---|---|
| `simple` | 400 | **0.925** | 0.885 |
| `multiple` | 200 | 0.885 | **0.895** |
| `parallel` | 200 | 0.780 | **0.800** |
| `parallel_multiple` | 200 | 0.775 | **0.840** |
| `java` | 100 | 0.630 | **0.670** |
| `javascript` | 50 | 0.840 | **0.860** |
| `irrelevance` | 240 | 0.758 | **0.892** |
| `live` (pooled) | 2251 | 0.592 | **0.757** |
| **non-live** | | 0.799 | **0.846** |
| **overall** | | 0.696 | **0.802** |

### 0-shot

| Category | Olmo-3-1025-7B | Qwen3.5-4B-Base |
|---|---|---|
| `simple` | **0.848** | 0.720 |
| `multiple` | **0.745** | 0.670 |
| `parallel` | **0.635** | 0.605 |
| `parallel_multiple` | 0.380 | **0.495** |
| `java` | **0.560** | 0.440 |
| `javascript` | **0.700** | 0.680 |
| `irrelevance` | 0.525 | **0.921** |
| `live` | 0.421 | **0.565** |
| **overall** | 0.509 | **0.613** |

### Reading these

- **Exemplar dependence is near-identical** across models (−0.187 vs −0.189
  overall from 5-shot to 0-shot), so the exemplars are not advantaging either one.
- **The exemplars carry the behaviours they demonstrate.** The two largest 0-shot
  drops for Olmo are `parallel_multiple` (−0.395) and `irrelevance` (−0.233) — the
  multi-call and declining exemplars. Live decode failures go from 1.1% to 38.6%
  without them; the model narrates instead of calling.
- **`simple` is close to a formatting check.** It stays at 0.848 for Olmo at 0-shot,
  because the instruction block spells the `[func(arg=value)]` format out in words.
  Complying needs instruction-following, not tool-calling priors. The categories that
  do probe tool calling are `live`, `irrelevance` and `parallel_multiple`.
- **Not leniency.** Re-scoring every correct instance under strict exact match —
  no case/punctuation folding, no int→float widening, no variable-name handling —
  leaves 96.2% of correct `simple` and 96.8% of correct `live` intact. Dropping every
  tolerance would move `simple` from 0.925 to 0.890.

## Known issue, not introduced here

The `bfcl` suite's headline number is mis-weighted by a pre-existing framework bug:
`_compute_child_average` flattens a nested suite with `child.expand()` and averages
the leaves equally, ignoring that suite's own aggregation strategy. So `bfcl` reports
a flat mean of 7 leaves for its non-live term (Olmo 0.657, Qwen 0.835) instead of
BFCL's average-of-averages (0.677, 0.846). The tables above use the corrected values.

`bfcl:non_live` and `bfcl:non_live_ast` are correct when run directly — the
flattening only bites at three levels of nesting. The same bug affects
`olmobase:easy:qa:rc`, `science:biology`, `science:nojudge` and `science:all`, so
fixing it changes numbers for suites outside this PR and is left for its own change.

## Not implemented

- **Multi-turn** — scoring needs BFCL's stateful API backend.
- **Executable and REST** — these grade by running predicted calls against live
  third-party APIs.

Because the executable categories are missing, `bfcl:non_live` is BFCL's non-live
overall with those terms left out and will not match a published non-live number.
`bfcl:non_live_ast` and the live summaries are directly comparable.

https://claude.ai/code/session_01P9Qzbda9kPQePRC7MKJxnu


README.md
```diff
@@ -660,6 +660,90 @@ See the [Harness](#harness) section above for full documentation on:
 - Defining tools with the `@tool` decorator
 - Programmatic usage
 
+## Function Calling (BFCL)
```
gregorybchris, line 663: This is quite a lot of BFCL specific stuff for the root README. I don't know how we want to handle this, but I'd propose having a root folder that can hold specs/docs and we can have a separate file per eval (or task depending on how much detail is needed)?

src/olmo_eval/evals/tasks/bfcl/__init__.py
```diff
@@ -0,0 +1,865 @@
...
+                "prepared_functions": prepare_function_docs(functions, language),
+            },
+        )
+
+
+def _parameters(properties: dict[str, Any], required: list[str]) -> dict[str, Any]:
+    return {"type": "dict", "properties": properties, "required": required}
+
+
+PYTHON_EXEMPLARS: tuple[Exemplar, ...] = (
```
gregorybchris, line 255: src/olmo_eval/evals/tasks/bfcl.py:253 (and :393, :519) — the exemplars are output-affecting but reach neither TaskConfig.to_dict() nor the task hash. I reproduced this: editing PYTHON_EXEMPLARS[0] changes the prompt (same prompt: False) while compute_task_hash(config.to_dict()) stays at a07ec00a9080a846, and to_dict()["fewshot_source"] is null. Two runs whose prompts differ by five demonstrations would then collide on one task hash and one artifact name. minerva_math already handles this shape by naming the fixed set through the serialized fewshot_source field ("minerva_math_fixed", minerva_math.py:65). Please do the same, or move the exemplar tuples onto BFCLCompletionFormatter as a field so Formatter.to_dict's asdict(self) picks them up, and add a test asserting the hash changes when an exemplar changes.

src/olmo_eval/common/scorers/bfcl/java_js.py
```diff
@@ -0,0 +1,407 @@
+"""Java and JavaScript call parsing and value conversion for BFCL.
```
gregorybchris, line 1: What should out philosophy be for adding utilities like this that are really not eval-specific, but so far only one eval uses them? This is under the bfcl folder. I'm fine having this specialized for this eval for now if you prefer and we can generalize it later if needed.

gregorybchris (approved):

README.md
pdasigi: Agreed. This should be in a README specific to bfcl. I'll move it.

src/olmo_eval/common/scorers/bfcl/java_js.py
pdasigi: I think it makes sense to keep this eval specific for now, and make it more general and move it to a more general location when other evals need it.

PR #388, opened 2026-09-18
Freeze OMEGA hill-climb pair and pin LiveCodeBench grading data

abhishekraok:
Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Add `omega_500:hillclimb` for the pinned AllenAI snapshot and strict primary metric. Preserve the existing `omega_500` dataset, flex primary, scorer, and context-adaptive generation default: the two snapshots differ in 11 problems and must not share a result identity. Both new OMEGA tasks default to a 32K generation cap; the SFT hill-climb launcher also sets it explicitly for its 64K serving context.

Add `omega_500_out`: 500 frozen test_out IDs matching the anchor's 35-family proportions. Normalize all eight family aliases to the anchor's names and report missing/extra ID examples plus duplicate counts on drift. The generator is importable and downloads only the required parquet splits from explicit immutable Hub revisions; generated provenance comes from those same arguments. SWE-bench and locked final sets are excluded.

Pin LiveCodeBench prompts and separately loaded execution test cases to the same revision. Resolve data sources through `get_data_source()` so URI-string overrides work. The hill-climb uses `livecodebench:lite` (release_v3, K=1), with its own 32K override. Existing SFT dev/paper batteries are retained.

Claude review addressed in bb5299d: isolated variant; preserved historical context-adaptive default; URI handling; canonical family labels; actionable drift errors; verified generator revisions and importable entry point; unittest guard at file end; cache cleanup in finally. The corresponding SFT launcher selects `omega_500:hillclimb` at open-instruct@14ccee75f.

Validation:

- 1,041 task/extraction/generator tests passed, 6 skipped. The final explicit-revision CLI test adjustment also passed all 5 generator tests.
- Direct OMEGA unittest execution runs all 13 tests.
- Fresh pinned-snapshot manifest generation is byte-identical after formatting.
- Source/test/script Ruff and formatting checks pass. Type checking has only the pre-existing unused-ignore warning in `harness/sandbox/executor.py`.
- Hill-climb launcher dry-run confirms the new variant and explicit 32K overrides for both OMEGA tasks.
- Previously completed real Docker LCB integration: 5 passed, covering stdin/call-based solutions, wrong answers, private cases, and the 612-problem v3 window. These review changes do not change the sandbox grader.

GPU serving qualification remains separate: the earlier B300 grouped-MoE gate failed before predictions, and a bounded native-MoE diagnostic was submitted. No model-quality result is claimed.

User-requested protocol update (e993bf3): default both new OMEGA tasks to 32768 tokens while leaving historical omega_500 unchanged. The 21 targeted OMEGA tests pass, as do repository Ruff lint and format checks.

Screen-reporting fix: retain finish reasons and completion-token counts from inline and server vLLM in prediction artifacts, including without logprobs. Chat logprob processing now preserves usage metadata. This makes cap-hit diagnostics available for future screen runs; already-running serving gates retain their original revision. Validation: 47 provider/prediction tests passed; repository Ruff checks passed.

OMEGA extraction performance: skip the leading-wildcard boxed-answer search when the normalized text contains no boxed opener. An impossible regex keeps the cascade position and format score unchanged. Differential tests compare the old cascade on unboxed/prefixed/boxed/incomplete-box answers; 22 OMEGA tests pass. A local synthetic unboxed input took 45 ms at 4K characters and 181 ms at 8K before this change; 128K characters takes about 4 ms afterward. Existing serving gates keep their pinned code; this improves future screens without changing answer selection.

The K=1 screen-diagnostics helper reports cap hits, empty outputs, grading errors, missing records and unknown telemetry, without recomputing aggregate task scores. Four tests cover missing telemetry, duplicate records, unsupported multi-sample inputs and candidate failures versus grader failures.


abhishekraok (commented):
Review by Claude Fable 5.1 (Claude Code).

Reviewed at cc112708; 6320156d only adds `tests/scripts/test_freeze_omega500_out.py`.

Verified directly: all three pinned revisions exist (LCB revision has test.jsonl–test6.jsonl); all 35 manifest parquet paths exist at the omega-explorative revision; regenerating the manifest with the script against fresh snapshots at those revisions is byte-identical after ruff format; the real `omega_500_out` load returns 500 unique instances; touched unit tests, ruff, and ty pass.

**Blocking question:** `omega_500` is rewritten in place rather than getting a variant. I loaded both datasets: `allenai/omega-500@113a7eb` differs from `saumyamalik/omega-500` in 11 of 500 problems (all `logic_puzzles_blocked_grid`, different grids and answers). With the strict primary and 32K cap, results stored under `omega_500` before and after this PR are silently incomparable, and the deleted comment warned about exactly this. Suggest `omega_500:hillclimb` (or a new name) and leaving the certified regime untouched.

Other items inline: unclamped `max_tokens=32768` (400s on ≤32K-context models), `config.data_source.revision` breaks on URI-string data sources, family keys differ across the in/out pair for the 8 aliased families, generator stamps revisions it never verifies (the subprocess tests added in 6320156d cover selection, thanks), new test class sits after the `__main__` guard.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -121,23 +120,20 @@ def score(self, instance: Instance, output: LMOutput) -> float:
 
 @register("omega_500")
 class Omega500(Task):
-    data_source = DataSource(path="saumyamalik/omega-500")
+    data_source = DataSource(
```
abhishekraok: Same 500 ids as `saumyamalik/omega-500`, but 11 items differ in question and ground truth (e.g. `omega_500_017`: 20x22 grid / GT 752952 → 47x42 grid / GT 320000). Changing dataset, primary metric, and budget under the existing task name makes prior `omega_500` results incomparable. Register a variant instead.

src/olmo_eval/evals/tasks/omega_500.py
```diff
@@ -121,23 +120,20 @@ def score(self, instance: Instance, output: LMOutput) -> float:
...
-    primary_metric = AccuracyMetric(name="exact_match_flex", scorer=_FLEX)
+    primary_metric = AccuracyMetric(name="exact_match", scorer=_STRICT)
     strip_thinking = True
-    # Defaults mirror oe-eval's ``omega_500:0-shot-chat_deepseek`` — the
-    # OLMO_3 suite entry and the config the parity certification ran on.
-    # (Plain ``0-shot-chat`` upstream uses temperature 0.7 with no top_p;
-    # register a variant if that regime is ever needed.) max_tokens=None
-    # generates to the model's context limit, matching the reference
-    # regime's effective behavior on any context size.
     sampling_params = SamplingParams(
```
abhishekraok, line 142: No provider clamps an explicit `max_tokens` against `max_model_len`. On a model served at ≤32K context, vLLM rejects every request (prompt + 32768 > max_model_len) where `None` previously adapted. Either clamp in the provider or keep `None` here and apply the cap in the hill-climb suite override.

src/olmo_eval/evals/tasks/livecodebench.py
```diff
@@ -383,6 +388,7 @@ def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance:
                 # cases without them travelling on the instance.
                 "row": index,
                 "test_repo": LIVECODEBENCH_REPO,
+                "test_revision": self.config.data_source.revision,
```
abhishekraok: `TaskConfig.data_source` may also be a URI string (`get_data_source` branches on `isinstance(..., str)`), which makes this an AttributeError. `self.config.get_data_source().revision` handles both.

src/olmo_eval/evals/tasks/omega_500_out.py
```diff
@@ -0,0 +1,38 @@
...
+
+    @property
+    def instances(self) -> Iterator[Instance]:
+        instances = list(self._load_instances_cached())
+        observed = [instance.metadata["id"] for instance in instances]
+        if len(observed) != len(SELECTED_IDS) or set(observed) != SELECTED_IDS:
+            raise ValueError("OMEGA test_out data does not match the frozen 500-item manifest")
+        yield from instances
+
+    def process_doc(self, doc: dict[str, Any], index: int = 0) -> Instance | None:
```
abhishekraok, line 49: Inherited `process_doc` emits `setting_key` here (`geometry_polygon_rotation`) while `omega_500` emits `geometry_rotation`, so 8 families have different keys across the pair. Normalize via the inverse of `FAMILY_ALIASES` so per-family in/out breakdowns line up, and test that the family sets are equal.

src/olmo_eval/evals/tasks/omega_500_out.py
```diff
@@ -0,0 +1,38 @@
...
+        revision="b04ae8d4757a2229e8ed65ac6a923e502d0bbb95",
+        data_files=tuple(f"{config}/test_out-00000-of-00001.parquet" for config in IDS_BY_CONFIG),
+    )
+
+    @property
+    def instances(self) -> Iterator[Instance]:
+        instances = list(self._load_instances_cached())
+        observed = [instance.metadata["id"] for instance in instances]
+        if len(observed) != len(SELECTED_IDS) or set(observed) != SELECTED_IDS:
+            raise ValueError("OMEGA test_out data does not match the frozen 500-item manifest")
```
abhishekraok: Nit: include a few missing/extra ids and the duplicate count in the message; otherwise drift means diffing 500 ids by hand.

scripts/hillclimb/freeze_omega500_out.py
```diff
@@ -0,0 +1,63 @@
...
+    "logic_puzzles_grid_rook": "logic_gridworld_rookmove",
+    "logic_puzzles_zebralogic": "logic_zebralogic",
+    "number_theory_digit_sum": "numbertheory_qr_sum",
+    "number_theory_prime_mod": "numbertheory_lte_qr",
+    "number_theory_triple_count": "numbertheory_ordered_lte",
+}
+counts = Counter(
+    r["family"] for r in pq.read_table(root / "omega-500/train-00000-of-00001.parquet").to_pylist()
+)
+text = '''"""Frozen OMEGA test_out dev IDs, stratified to match omega-500 families.
```
abhishekraok: Revisions are stamped into the docstring but never checked against the snapshot dir, so a rerun on a newer snapshot produces a manifest with false provenance. Take revisions as args and `snapshot_download(..., revision=..., allow_patterns=[...])` — that is how I reproduced the manifest byte-for-byte.

scripts/hillclimb/freeze_omega500_out.py
```diff
@@ -0,0 +1,63 @@
+"""Reproduce the frozen OMEGA out manifest from pinned local parquet snapshots."""
+
+import argparse
+import hashlib
+from collections import Counter
+from pathlib import Path
+
+import pyarrow.parquet as pq
+
+parser = argparse.ArgumentParser(description=__doc__)
```
abhishekraok: Addressed in 6320156d by `tests/scripts/test_freeze_omega500_out.py`, which exercises selection, alias mapping, and the invalid-pool assertions via subprocess. Remaining nit only: module-level argparse means the logic can be tested end-to-end but not imported.

tests/evals/tasks/test_omega_500.py
```diff
@@ -121,3 +126,26 @@ def test_flex_scorer_name_derived(self) -> None:
 
 if __name__ == "__main__":
     unittest.main()
+
+
```
abhishekraok: Defined after `if __name__ == "__main__": unittest.main()`, so running the file directly skips these. Move the guard to the end.

tests/evals/tasks/test_livecodebench.py
```diff
@@ -645,3 +648,26 @@ def test_undecodable_test_cases_yield_no_tests_and_no_pass(self) -> None:
...
+
+
+def test_prompt_and_grader_use_same_dataset_revision():
+    for name in ("livecodebench", "livecodebench_hidden"):
+        task = get_task(name)
+        assert task.config.data_source.revision == LIVECODEBENCH_REVISION
+        instance = task.process_doc(STDIN_DOC)
+        assert instance.metadata["test_revision"] == LIVECODEBENCH_REVISION
+
+
```
abhishekraok, line 660: Nit: `cache_clear()` only runs on the success path; use try/finally or a fixture.

scripts/hillclimb/freeze_omega500_out.py
```diff
@@ -0,0 +1,63 @@
...
+    "logic_puzzles_zebralogic": "logic_zebralogic",
+    "number_theory_digit_sum": "numbertheory_qr_sum",
+    "number_theory_prime_mod": "numbertheory_lte_qr",
+    "number_theory_triple_count": "numbertheory_ordered_lte",
+}
+counts = Counter(
+    r["family"] for r in pq.read_table(root / "omega-500/train-00000-of-00001.parquet").to_pylist()
+)
+text = '''"""Frozen OMEGA test_out dev IDs, stratified to match omega-500 families.
+
```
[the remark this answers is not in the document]
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Added generator fixture tests covering deterministic selection, aliases, invalid candidate pools, and pinned revision retrieval. All 5 generator tests pass.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Added omega_500:hillclimb and restored the existing omega_500 dataset, flex primary metric, scorer, and generation defaults.

src/olmo_eval/evals/tasks/omega_500.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Restored max_tokens=None for context-adaptive defaults. The SFT launcher explicitly applies 32768 to both hill-climb tasks at its 65536 serving context.

src/olmo_eval/evals/tasks/livecodebench.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Now resolves revision through self.config.get_data_source(), with a regression test using a URI-string override.

src/olmo_eval/evals/tasks/omega_500_out.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Normalizes the eight aliases through the inverse map. The regression test verifies that companion and anchor family sets match.

src/olmo_eval/evals/tasks/omega_500_out.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). The error now includes missing/extra ID examples and the duplicate count, covered by a drift regression test.

scripts/hillclimb/freeze_omega500_out.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). The generator accepts immutable revision arguments and passes them to snapshot_download with split-specific allow_patterns. Provenance uses those same arguments; regeneration from real pinned snapshots is byte-identical.

scripts/hillclimb/freeze_omega500_out.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Moved argument parsing into main(argv) behind the module guard and exposed importable freeze_manifest. Tests import the module without executing the CLI.

tests/evals/tasks/test_omega_500.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Moved the unittest main guard to the end. Direct execution now runs all 13 tests.

tests/evals/tasks/test_livecodebench.py
abhishekraok: Codex · codex://threads/01a0b658-a648-7e20-bc79-f871e3fb4e18

Addressed in [bb5299d](https://github.com/allenai/olmo-eval/commit/bb5299d6d0ba66dd2ca5d13201ab7cc68c350091). Wrapped cache-using assertions in try/finally so cache_clear runs on failure as well as success.

gregorybchris:

Looks good overall! The frozen manifest, the drift error, and the revision guards are careful work. Two things should be fixed before merging.

I found two blocking issues. Both affect persisted per-instance data. The first changes the stored instance IDs of the historical `omega_500` task without changing its task hash. The second persists a wrong per-output token count when `num_samples > 1`. The non-blocking items cover the variant's shape, LiveCodeBench task identity, and structure.

I ran the project checks on the branch head. Ruff and formatting pass. `ty` shows only the existing unused-ignore warning in `harness/sandbox/executor.py`. The unit suite passes (2450 passed, 28 skipped). I also checked the Hub: all three pinned revisions (`livecodebench/code_generation_lite@0fe84c39`, `allenai/omega-500@113a7eb8`, `allenai/omega-explorative@b04ae8d4`) are the current HEAD commits.

### Blocking

1. `src/olmo_eval/evals/tasks/omega_500.py:158`
   `process_doc` now prefers `doc["id"]`. `saumyamalik/omega-500` has an `id` column (`omega_500_000`, ...), so the historical `omega_500` task now stores `native_id="omega_500_000"` where it used to store `"0"`. The task hash does not change. I computed `TaskConfig.to_dict()` on `main` and on this branch and got the same hash. On the same dataset row, `process_doc` returns `0` on `main` and `'omega_500_000'` here. `analysis/pairwise.py` matches instances by `(task_hash, native_id)`. A run of `omega_500` from before this PR and one from after it therefore share a hash but have no instances in common. The description says the historical task is preserved, and this is the one place it is not.
   Could we keep `doc.get("index", index)` for `omega_500` and use the dataset ID only in the new tasks? Registering the AllenAI snapshot as its own `Omega500` subclass (see non-blocking 1) makes that easy. `Omega500Out` already overrides `process_doc`. If you and the maintainers would rather take the better IDs everywhere, please say so in the PR description, since older `omega_500` runs will stop pairing with newer ones. Either way, please add a test that pins the `native_id` the historical task produces for a row that has an `id` column.

2. `src/olmo_eval/runners/io/builders.py:74`, with `src/olmo_eval/inference/providers/vllm_server.py:958` and `:787`
   The server path fills `completion_tokens` from the response-level `usage`. The server sums that over all `n` choices. I built a chat response with two choices of 10 and 90 tokens and `usage.completion_tokens=100`, ran it through `_generate_chat` and `build_predictions` with `num_samples=2`, and both persisted outputs report `completion_tokens: 100`. The inline `vllm` provider stores `len(completion.token_ids)` per output. The same prediction field therefore means different things depending on the provider. Before this PR the value stayed in in-memory metadata. Now it is written to every prediction row. `screen_diagnostics.py` refuses K>1, so your own use is safe. Any other reader, or a pass@k task on `vllm_server`, gets inflated per-sample counts, and a cap check against them would report false cap hits.
   Please persist only per-output counts, and add a test with two choices. `usage` is only consumed from metadata by `build_predictions` (the throughput collector reads the response directly), so one option is to drop the field when the response has more than one choice:
   ```python
   if usage:
       metadata["prompt_tokens"] = usage.prompt_tokens
       if len(response.choices) == 1:
           metadata["completion_tokens"] = usage.completion_tokens
   ```
   The completions path (`_build_completion_output` via `_completion_usage_metadata`) needs the same treatment.

### Non-blocking

1. `src/olmo_eval/evals/tasks/omega_500.py:174`
   `omega_500:hillclimb` names a research protocol and bundles three separate changes: a different dataset snapshot, a different primary metric, and a 32K cap. Per the decisions in the guide, regimes compose from variants, and variant names describe a format or metric modifier. The AllenAI snapshot corrects some questions and answers, which fits the versioned-fix pattern. Could it be `omega_500:v2` (or a separate registered task) with the strict primary, and the 32K cap set by the launcher through a sampling override? The launcher already sets it explicitly. The snapshot revision also appears in three places (this variant, `OMEGA_REVISION` in the freeze script, and the tests). Could we define it once in `omega_500.py` and import it?

2. `src/olmo_eval/evals/tasks/livecodebench.py:352` and `:421`
   `revision` is serialized in `DataSource.to_dict()`, so pinning it changes the task hash of every existing `livecodebench` and `livecodebench_hidden` result. The pin is the current HEAD, so the data is unchanged and the change only affects identity. That is fine, but please state it in the PR description so nobody is surprised when new LCB runs do not pair with earlier ones.

3. `src/olmo_eval/evals/tasks/livecodebench.py:391`
   `test_revision` now follows the configured data source, but `test_repo` is still the hard-coded `LIVECODEBENCH_REPO`. If `data_source` is overridden to point at a mirror, the grader fetches the canonical repo at the mirror's revision. The `question_id` check in `_stage_problem` makes that fail loudly, so it is not a silent-zero risk. Could `test_repo` also come from `self.config.get_data_source().path`?

4. `scripts/hillclimb/`
   Both scripts serve one internal development protocol, and `scripts/internal/` already exists for that kind of tooling. Could they move there? Alternatively, the diagnostics script could become a general predictions audit that is not tied to the hill-climb name.

5. This PR bundles six independent changes: the OMEGA variant, the new `omega_500_out` task, the LCB pin, `finish_reason` persistence in the vLLM providers, the extraction speedup, and the diagnostics script. No need to split it now, but separate PRs would make changes like these easier to review and revert in the future.

### Nits

- Nit: `src/olmo_eval/evals/tasks/_omega_500_out_ids.py` could live in `src/olmo_eval/evals/tasks/constants/`, alongside the other static task tables.
- Nit: `scripts/hillclimb/freeze_omega500_out.py:71` validates the candidate pool with `assert`, which `python -O` skips. A `ValueError` would always run.
- Nit: `tests/evals/tasks/test_omega_500.py:132` `TestOmegaAnswerPrefixes` holds a stable-ID test. The class name doesn't match its contents.
- Nit: every commit is titled `Codex · codex://threads/...`. Please give the squash commit a descriptive message.

### Before merge

Please fix the two blocking items and add regression tests for each: a pinned `native_id` for historical `omega_500`, and per-output `completion_tokens` with two choices. After that, you may merge once CI is green.

abhishekraok:
Claude Opus 5.5 (Claude Code) · claude://claude.ai/epitaxy/local_2cdf0c63-88a3-45aa-b486-78ff5a91ca2a

@gregorybchris thanks, this was a careful review. Addressed in ed6f2912 and 91569af7.

### Blocking

1. **Historical `omega_500` IDs.** Fixed. `omega_500` uses `doc.get("index", index)` again. On a row with an `id` column it now returns `0`, and its task hash is `4a504548233d37ea` on both `main` and this branch. `test_historical_task_keeps_positional_id` pins this. Dataset IDs are used only by the new subclass and `omega_500_out`.
2. **Multi-choice `completion_tokens`.** Fixed as you suggested, on both paths. `_completion_usage_metadata` now drops `completion_tokens` when the response has more than one choice, and the chat path uses the same helper. Tests cover two choices through chat into `build_predictions`, SDK completions, and raw-HTTP completions. They fail on the previous code. This also fixes the throughput collector, which summed the response total once per output.

### Non-blocking

1. **Variant shape.** It is now a separate registered task, `Omega500HillClimb`, following the `mbpp:3shot:v2` pattern, because a variant can't change `process_doc`. I kept the spec `omega_500:hillclimb` rather than `:v2`. The open-instruct launchers and the experiment ledger select it by that name, and making it a separate task changes its hash anyway, so a rename would only add churn. The 32K default stays in the task because the author chose it as the protocol default. The launcher also passes `-o max_tokens=32768`, so resolved hashes are the same either way. The revisions are now defined once, as `OMEGA_500_REVISION` in `omega_500.py` and `OMEGA_EXPLORATIVE_REVISION` in `omega_500_out.py`. The freeze script and the tests import them, and a test checks that the manifest's recorded provenance matches both pins.
2. **LCB hash change.** Now stated in the PR description, with a table of which task hashes move.
3. **`test_repo`.** Done. It comes from `get_data_source().path` with `hf://` stripped, and is tested with a mirror `DataSource`, a URI string, and the default.
4. **Scripts.** Moved to `scripts/internal/`.
5. **Bundling.** Understood. I'll leave it as is for this PR.

### Nits

- The manifest moved to `tasks/constants/omega_500_out.py`.
- The generator raises `ValueError` for duplicate or insufficient candidate pools.
- The class is renamed `TestOmegaDocIds`, and the answer-prefix tests stay separate.
- For the squash commit, the PR title makes a descriptive message.

Validation: 2458 passed, 28 skipped. Ruff and format pass. `ty` shows only the existing executor warning.

gregorybchris (approved):

Thanks for the follow-up. Both blocking items are fixed in ed6f2912 and 91569af7, and the non-blocking items and nits are addressed. The unit suite, Ruff, and `ty` all pass at 91569af7, and the task hashes for `omega_500` and `omega_500_out` match the earlier runs. Approving.

